PR #8, opened 2023-03-07
load balance for sudo account.

SaMotlagh:


node/src/chain_spec.rs
```diff
@@ -153,7 +153,7 @@ struct ColdkeyHotkeys {
 }
 
 pub fn finney_config() -> Result<ChainSpec, String> {
-	let path: PathBuf = std::path::PathBuf::from("/Users/ec2-user/repos/subtensorv3/nakamoto_gen.json");
+	let path: PathBuf = std::path::PathBuf::from("/Users/sam/Documents/GIT/subtensorv3/nakamoto_gen.json");
```
shibshib: Careful here, this is your local path right? should keep it as ec2-user

shibshib (approved):

PR #10, opened 2023-03-07
Refactor local-net Bash scripts

S0AndS0:
Mostly reduces sub-shell and piping costs, plus by utilizing Bash lists it be possible to exert more control over field separators

Only bit I currently question is at `scripts/localnet.sh:3`
```bash
: "${CHAIN:=local}"
```
... and probably could move defaulting, or checking, of variables' value to the first usage


Rubberbandits:
All the changes look good, thanks again!

Rubberbandits (approved):

PR #19, opened 2023-03-09
Set weights to 0 and added development chain spec

shibshib:
This PR:
- Adds some logic to calculate weights (although it is unused for now as we wait on the throttling logic PR).
- Sets all dispatch call weights to 0 for now, until we set them using a benchmark.

### DIFF
--- a/node/src/chain_spec.rs
+++ b/node/src/chain_spec.rs
@@ -198,9 +198,9 @@ pub fn finney_config() -> Result<ChainSpec, String> {
 
 	Ok(ChainSpec::from_genesis(
 		// Name
-		"Finney",
+		"Bittensor",
 		// ID
-		"Finney",
+		"bittensor",
 		ChainType::Development,
 		move || {
 			finney_genesis(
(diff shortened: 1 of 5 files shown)


SaMotlagh (approved):

node/src/chain_spec.rs
```diff
@@ -198,9 +198,9 @@ pub fn finney_config() -> Result<ChainSpec, String> {
 
 	Ok(ChainSpec::from_genesis(
 		// Name
-		"Finney",
+		"Bittensor",
 		// ID
-		"Finney",
+		"bittensor",
```
eduardogr, line 203: Just got my attention, Is the lowercase on purpose? or we should have the ID with the Capital char as we have the Name?

shibshib: yes lowercase is on purpose, since it's an ID and not the name :) the name can be capitalized.

node/src/chain_spec.rs
```diff
@@ -246,7 +246,7 @@ pub fn finney_config() -> Result<ChainSpec, String> {
 		// Telemetry
 		None,
 		// Protocol ID
-		None,
+		Some("bittensor"),
```
eduardogr, line 249: Could the ID be extracted to a constant? sounds a good practice to keep the value aside and not duplicate it.
What do you think about that?

PR #20, opened 2023-03-09
Generic rate limiting for TXs

Rubberbandits:


pallets/subtensor/src/lib.rs
```diff
@@ -1244,7 +1259,15 @@ pub mod pallet {
 			Self::do_sudo_set_serving_rate_limit( origin, serving_rate_limit )
 		}
 
+		// Sudo call for setting tx rate limit
 		#[pallet::weight((0, DispatchClass::Operational, Pays::No))]
+		pub fn sudo_set_tx_rate_limit( origin:OriginFor<T>, tx_rate_limit: u64 ) -> DispatchResult {  
+			Self::do_sudo_set_tx_rate_limit( origin, tx_rate_limit )
+		}
+
+		#[pallet::weight((Weight::from_ref_time(39_699_000 as u64)
```
SaMotlagh, line 1268: did you add weights for sudo_set_max_burn?

pallets/subtensor/src/lib.rs
```diff
@@ -347,6 +349,17 @@ pub mod pallet {
...
 
+	// Rate limiting
+	#[pallet::type_value]
+	pub fn DefaultTxRateLimit<T: Config>() -> u64 { T::InitialTxRateLimit::get() }
+	#[pallet::type_value]
+	pub fn DefaultLastTxBlock<T: Config>() -> u64 { 0 }
+
+	#[pallet::storage] // --- ITEM ( tx_rate_limit )
+	pub(super) type TxRateLimit<T> = StorageValue<_, u64, ValueQuery, DefaultTxRateLimit<T>>;
+	#[pallet::storage] // --- MAP ( key ) --> last_block
```
SaMotlagh, line 343: hotkey?

pallets/subtensor/src/lib.rs
SaMotlagh: did you add weights for sudo_set_max_burn?

SaMotlagh (changes requested):

pallets/subtensor/src/lib.rs
shibshib: let's remove the weights from this one, we don't need them for now.

pallets/subtensor/tests/mock.rs
```diff
@@ -114,6 +114,7 @@ parameter_types! {
 	pub const InitialDefaultTake: u16 = 11_796; // 18% honest number.
 	pub const InitialWeightsVersionKey: u16 = 0; 
 	pub const InitialServingRateLimit: u64 = 0; // No limit.
+	pub const InitialTxRateLimit: u64 = 0; // No limit.
```
shibshib: should we not set a limit right off the bat?

pallets/subtensor/src/lib.rs
Rubberbandits: This is the storage value to allow configuration of rate limit on the network, not where we store the last tx block per key

pallets/subtensor/src/lib.rs
Rubberbandits: I don’t remember setting a weight here but it might’ve gotten added during a merge conflict resolution on accident.

pallets/subtensor/tests/mock.rs
Rubberbandits: What limit should be set? 2 blocks per stake/delegate action?

shibshib (approved):

PR #26, opened 2023-03-09
Add tests for neuron info pallet

S0AndS0:
This adds missing development dependency as well as tests for `Some`/`None` and Vector return values
> Note; suggestions on how to address new `@TODO` are most welcomed


shibshib:
Looks good @S0AndS0 but can you just remove the Cargo.lock file from the commits? it's auto-generated anyway

shibshib (approved):

camfairchild (approved):
LGTM

S0AndS0:
> @shibshib Looks good @S0AndS0 but can you just remove the Cargo.lock file from the commits? it's auto-generated anyway

I can though am a bit apprehensive about doing so because the official documentation encourages tracking `Cargo.lock` files for binaries;

> ### [Why do binaries have `Cargo.lock` in version control, but not libraries?](https://doc.rust-lang.org/cargo/faq.html#why-do-binaries-have-cargolock-in-version-control-but-not-libraries)
>
> The purpose of a `Cargo.lock` lockfile is to describe the state of the world at the time of a successful build. Cargo uses the lockfile to provide deterministic builds on different times and different systems, by ensuring that the exact same dependencies and versions are used as when the `Cargo.lock` file was originally generated.
>
> This property is most desirable from applications and packages which are at the very end of the dependency chain (binaries). As a result, it is recommended that all binaries check in their `Cargo.lock`.

... For the meantime I'll rebase from `origin/main` so as to remove blockages caused by merge conflict

PR #31, opened 2023-03-13
Custom rpc return vec u8

camfairchild:
This PR changes the response format of the custom RPCs to a `Vec<u8>`.  
This should allow us to upgrade the format of the responses more flexibly, inside of the runtime, provided we update the client-decoding to match.  
### Node (server) -side
We essentially do 
```rust
serde_json::to_string(&result).unwrap().as_bytes().to_vec()
```
to get a `Vec<u8>` (utf-8) of a JSON representation for the result.  
### Python Client-side 
Then, we do 
```python
result: List[int] = json_body['result']
as_bytes = bytes(result)
as_json_str = as_bytes.decode('utf-8')
as_json: Dict[str, Any] = json.loads(as_json_str)
return as_json
```

### DIFF
--- a/pallets/subtensor/rpc/src/lib.rs
+++ b/pallets/subtensor/rpc/src/lib.rs
@@ -8,7 +8,7 @@ use jsonrpsee::{
 use sp_blockchain::HeaderBackend;
 use sp_runtime::{
 	generic::BlockId,
-	traits::{Block as BlockT},
+	traits::{Block as BlockT}
 };
 use std::sync::Arc;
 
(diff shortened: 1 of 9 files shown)


camfairchild:
Note: This update requires a restart of the chain to reflect the change of RPC response format

shibshib (approved):

S0AndS0:
The [`unwrap`][1] in `runtime/src/lib.rs` has be a bit concerned;
> `body: serde_json::to_string(&result).unwrap().as_bytes().to_vec()`

... is there a chance the `InfoResponse` it returns is compatible with `?` syntax, so we avoid a panic state?
> `body: serde_json::to_string(&result)?.as_bytes().to_vec()`

[1]: https://github.com/opentensor/subtensorv3/pull/31/files#diff-0ec06ea58bd455f09ce6b3bb4c2c1c0d37bda51c1e1be2151c560c9c973959ecR676

pallets/subtensor/rpc/src/lib.rs
S0AndS0, line 1: I really like the simplifications!  Specifically removal of wrapping `Option` within a `RpcResult`

camfairchild:
> The [`unwrap`](https://github.com/opentensor/subtensorv3/pull/31/files#diff-0ec06ea58bd455f09ce6b3bb4c2c1c0d37bda51c1e1be2151c560c9c973959ecR676) in `runtime/src/lib.rs` has be a bit concerned;
> 
> > `body: serde_json::to_string(&result).unwrap().as_bytes().to_vec()`
> 
> ... is there a chance the `InfoResponse` it returns is compatible with `?` syntax, so we avoid a panic state?
> 
> > `body: serde_json::to_string(&result)?.as_bytes().to_vec()`

I realized it doesn't effect the rest of the node if it panics as it's caught in the custom rpc handler and logged.

S0AndS0 (approved):

PR #34, opened 2023-03-14
🐳 Started dockerfile setup

shibshib:
This PR sets up docker containers for users. It significantly reduces the container size to from about 2GB to 100MB, and allows for a very lightweight docker image that runs the subtensor layer for miners. 

Additionally, this Implements warp sync capability ✅ 

This is a high surface PR as it involves upgrading a lot of the libraries that we use, and requires more testing to ensure it works properly.


Dockerfile
```diff
@@ -1,37 +1,43 @@
...
+        ai.opentensor.image.description="Opentensor Subtensor Blockchain" \
+        ai.opentensor.image.revision="${VCS_REF}" \
+        ai.opentensor.image.created="${BUILD_DATE}" \
+        ai.opentensor.image.documentation="https://docs.bittensor.com"
+
+# show backtraces
+ENV RUST_BACKTRACE 1
+
+# Necessary libraries for Rust execution
+RUN apt-get update && apt-get install -y curl build-essential protobuf-compiler clang git
```
eduardogr, line 22: do we need git inside of the image?

Dockerfile
```diff
@@ -1,37 +1,43 @@
...
-ENTRYPOINT ["/usr/local/bin/node-subtensor"]
+
+# Update to nightly toolchain
+RUN ./scripts/init.sh
+
+# Cargo build
+RUN cargo build --release --features runtime-benchmarks --locked
+EXPOSE 30333 9933 9944
+
+FROM $BASE_IMAGE
```
eduardogr, line 39: i have a doubt here, is not the name of the stage what we have to reference here?? 
In our case will be `builder`, is this right?

node/src/chain_spec.rs
```diff
@@ -243,7 +244,9 @@ pub fn finney_config() -> Result<ChainSpec, String> {
 			)
 		},
 		// Bootnodes
-		vec![],
+		vec![
+			"/ip4/13.58.175.193/tcp/30333/p2p/12D3KooWEXY393DjTvBip3DnZeJkSAeZrg6edxqu45ywHzzQS2BY".parse().unwrap()
```
eduardogr: can we place this IP, and the node ID, somewhere that makes it clear for us that is something that could change? Some config file or constants files?

SaMotlagh (approved):

docker-compose.yml
```diff
@@ -1,17 +1,21 @@
...
+      - "9933"
     environment:
       - CARGO_HOME=/var/www/node-subtensor/.cargo
-    volumes:
-      - .:/var/www/node-subtensor
-      - type: bind
-        source: ./.local
-        target: /root/.local
-    command: bash -c "cargo build --release && ./target/release/node-subtensor --dev --ws-external"
+    command: bash -c "./node-subtensor --base-path /tmp/blockchain --chain /raw_spec.json --rpc-external --ws-external --rpc-cors all --no-mdns --ws-max-connections 10000 --in-peers 500 --out-peers 500 --bootnodes /ip4/13.58.175.193/tcp/30333/p2p/12D3KooWEXY393DjTvBip3DnZeJkSAeZrg6edxqu45ywHzzQS2BY"
```
eduardogr: There is a way to write the command in multiple lines for a docker compose file. I was looking for it, but I don't find it. 
I think will be nice to have it the command in multiple lines, it can make it easier to read

eduardogr (approved):
With the absence of resolving the comments I've made, LGMT

Nice!

S0AndS0 (approved):

Dockerfile
S0AndS0, line 1: Few thoughts for `Dockerfile` that I noted;

- [`ADD`][1] may be able to replace `mkdir` command and `COPY` instruction

```diff
diff --git a/Dockerfile b/Dockerfile
index a1df576..2e74e9f 100644
--- a/Dockerfile
+++ b/Dockerfile
@@ -25,9 +25,8 @@ RUN apt-get update && apt-get install -y curl build-essential protobuf-compiler
 RUN curl https://sh.rustup.rs -sSf | sh -s -- -y
 ENV PATH="/root/.cargo/bin:${PATH}"

-RUN mkdir -p /subtensor
 WORKDIR /subtensor
-COPY . .
+ADD . .

 # Update to nightly toolchain
 RUN ./scripts/init.sh
```

- [Official][2] Rust Docker image could reduce installation setup time

```diff
diff --git a/Dockerfile b/Dockerfile
index a1df576..888759b 100644
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,5 +1,5 @@

-ARG BASE_IMAGE=ubuntu:20.04
+ARG BASE_IMAGE=rust:1.67

 FROM $BASE_IMAGE as builder
 SHELL ["/bin/bash", "-c"]
@@ -21,10 +21,6 @@ ENV RUST_BACKTRACE 1
 # Necessary libraries for Rust execution
 RUN apt-get update && apt-get install -y curl build-essential protobuf-compiler clang git

-# Install cargo and Rust
-RUN curl https://sh.rustup.rs -sSf | sh -s -- -y
-ENV PATH="/root/.cargo/bin:${PATH}"
-
 RUN mkdir -p /subtensor
 WORKDIR /subtensor
 COPY . .
```

[1]: https://docs.docker.com/engine/reference/builder/#add
[2]: https://www.docker.com/blog/simplify-your-deployments-using-the-rust-official-image/

docker-compose.yml
S0AndS0: @eduardogr Yep it is possible to abuse the `entrypoint` key, though may still not be as pretty as either of us would hope, eg

```diff
diff --git a/docker-compose.yml b/docker-compose.yml
index 060457e..6f3fcc9 100644
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -17,5 +17,25 @@ services:
       - "9933"
     environment:
       - CARGO_HOME=/var/www/node-subtensor/.cargo
-    command: bash -c "./node-subtensor --base-path /tmp/blockchain --chain /raw_spec.json --rpc-external --ws-external --rpc-cors all --no-mdns --ws-max-connections 10000 --in-peers 500 --out-peers 500 --bootnode
s /ip4/13.58.175.193/tcp/30333/p2p/12D3KooWEXY393DjTvBip3DnZeJkSAeZrg6edxqu45ywHzzQS2BY"
+    entrypoint:
+      - bash
+      - -c
+      - ./node-subtensor
+      - --base-path
+      - /tmp/blockchain
+      - --chain
+      - /raw_spec.json
+      - --rpc-external
+      - --ws-external
+      - --rpc-cors
+      - all
+      - --no-mdns
+      - --ws-max-connections
+      - 10000
+      - --in-peers
+      - 500
+      - --out-peers
+      - 500
+      - --bootnodes
+      - /ip4/13.58.175.193/tcp/30333/p2p/12D3KooWEXY393DjTvBip3DnZeJkSAeZrg6edxqu45ywHzzQS2BY
```

... each argument/parameter would end up on separate lines, due to `IFS` defaults, but if `=` is a valid option/value separator for Rust's argument parsing then it may be possible to pare-up things

S0AndS0 (approved):

Dockerfile
shibshib: Noted! Let me make these in a subsequent PR

docker-compose.yml
shibshib: Thanks that's a good point! Making a subsequent PR for this

PR #42, opened 2023-03-16
Custom rpc scale codec

camfairchild:
This PR modifies the custom RPC encodings to use compact ints and better interates over the storage maps.  
Results in ~1-2s pull of the metagraph with 4096 neurons.


runtime/src/lib.rs
```diff
@@ -705,19 +704,18 @@ impl_runtime_apis! {
...
 			} else {
 				vec![]
 			}
 		}
 
 		fn get_subnets_info() -> Vec<u8> {
 			let result = SubtensorModule::get_subnets_info();
-			serde_json::to_string(&result).expect("Could not convert SubnetInfo list to JSON")
-				.as_bytes().to_vec()
+			result.encode()
```
SaMotlagh, line 718: does this function always return value?

SaMotlagh (approved):

unconst (approved):

PR #54, opened 2023-03-19
check immunity period while identifying pruning neuorn.

SaMotlagh:
this PR includes: 
- check immunity period while identifying pruning uid. The uid with lowest pruning score out of immunity period is pruned. if all uids are in their immunity period, the uid with lowest prunning score is pruned.
- add test cases
- fixed some failing test cases

### DIFF
--- a/pallets/subtensor/src/registration.rs
+++ b/pallets/subtensor/src/registration.rs
@@ -352,36 +352,57 @@ impl<T: Config> Pallet<T> {
         return real_hash
     }
 
-    // Determine which peer to prune from the network by finding the element with the lowest pruning score.
+    // Determine which peer to prune from the network by finding the element with the lowest pruning score out of
+    // immunity period. If all neurons are in immunity period, return node with lowest prunning score.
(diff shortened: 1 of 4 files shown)


Rubberbandits (approved):

camfairchild (approved):
LGTM

pallets/subtensor/src/registration.rs
S0AndS0, line 1: I may be wrong, likely am, because at first glance the checks on lines `370-371` and `383-384`look identical;

- `if current_block - block_at_registration <  immunity_period`
- `if min_score_in_immunity_period > pruning_score`

... may be good candidates for abstraction to function?

S0AndS0 (approved):

pallets/subtensor/src/registration.rs
SaMotlagh: you are right; the checks are identical, the body is different. however, I will definitely rewrite this fn with more abstraction after launch. thanks for suggestion :)

PR #55, opened 2023-03-20
Track hyperparameter changes

opentaco:
### Track hyperparameter changes

- Any hyperparameter changes made through sudo call on-chain should also be manually recorded in this file.
- Normally a PR for the change requested in this file should be made and approved before the actual sudo call.
- Each netuid should track its own set of hyperparameters in a separate block in this file under its own heading (only netuid 3 has been added for now).

### DIFF
--- a/hyperparameters.md
+++ b/hyperparameters.md
@@ -1,0 +1,38 @@
+### netuid 3
+```rust
+Rho: u16 = 10;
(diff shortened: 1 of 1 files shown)


SaMotlagh (approved):

camfairchild (approved):
👍🏼

PR #68, opened 2023-03-21
Increased block weights to 4 seconds of compute

shibshib:
This increases the block weight limit to 4 seconds of compute for a 12 second average block time.


unconst (approved):
LGTM

SaMotlagh (approved):

PR #70, opened 2023-03-22
[Epoch Improvement] Use only validator stake in epoch

opentaco:
### [Epoch Improvement] Use only validator stake in epoch

Now explicitly masks out non-validators from active_stake, and does not just rely on masking in the weights:
- Main benefit is that stake-weighted median should now be sparse and much faster, due to non-validator stake giving zero weights being factored out.
- Ensures that the majority (kappa) value is now exact, no longer offset by zero-weighting non-validator stake.

### DIFF
--- a/pallets/subtensor/src/epoch.rs
+++ b/pallets/subtensor/src/epoch.rs
@@ -49,6 +49,7 @@ impl<T: Config> Pallet<T> {
         // ===========
         // == Stake ==
         // ===========
+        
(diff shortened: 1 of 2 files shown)


shibshib (approved):

isabella618033 (approved):
Amazing!

shibshib:
@opentaco LGTM but can you also increase spec_version to 109?

Eugene-hu (approved):

SaMotlagh (approved):

PR #71, opened 2023-03-22
increased priority for set_weight call.

SaMotlagh:


Eugene-hu (approved):

adriansmares:
Wouldn't it be better for the priority to take into account what kind of weights are set here ? Keep-alives shouldn't weight as much as validator weights.

shibshib (approved):

shibshib:
@SaMotlagh please resolve conflicts and let me know when ready, I'll push this on a chain upgrade.

PR #75, opened 2023-03-22
enable registration.

SaMotlagh:
enable registration.


pallets/subtensor/src/lib.rs
```diff
@@ -609,7 +609,6 @@ pub mod pallet {
 		MaxAllowedUidsExceeded, // --- Thrown when number of accounts going to be registered exceed MaxAllowedUids for the network.
 		TooManyUids, // ---- Thrown when the caller attempts to set weights with more uids than allowed.
 		TxRateLimitExceeded, // --- Thrown when a transactor exceeds the rate limit for transactions.
-		RegistrationDisabled // --- Thrown when registration is disabled
```
shibshib: I think @Rubberbandits is working on a sudo registration disabling mechanism, let's keep this enum for now.

shibshib (approved):

pallets/subtensor/src/lib.rs
SaMotlagh: you are right. I did fix it.

PR #78, opened 2023-03-23
Per-subnet registration allowance

Rubberbandits:
This PR allows subnets to allow/disallow registration to it through a sudo call.


pallets/subtensor/src/lib.rs
```diff
@@ -1296,6 +1301,12 @@ pub mod pallet {
 			Self::do_sudo_set_tx_rate_limit( origin, tx_rate_limit )
 		}
 
+		// Sudo call for setting registration allowed
+		#[pallet::weight((0, DispatchClass::Operational, Pays::No))]
+		pub fn sudo_set_registration_allowed( origin:OriginFor<T>, netuid: u16, registration_allowed: bool ) -> DispatchResult {  
+			Self::do_sudo_set_network_registration_allowed( origin, netuid, registration_allowed )
+		}
+
```
SaMotlagh, line 1321: - we need to add index for each call. that would be nice if you add this call at the end and increment index. 
- Also, you could run benchmark fn for this call and see how many r/w it needs. it must be similar to other calls and need one r/w. If so, I recommend to add similar weights for it.

SaMotlagh (changes requested):

camfairchild (changes requested):
Needs a test for new sudo call, should check if registration fails with it set to `False`

Rubberbandits (commented):
LGTM

shibshib (approved):

PR #79, opened 2023-03-23
Optimize (re-)build time via dependency caching

S0AndS0:
Highlights of proposed changes within this Pull Request;

- Adds `matrix` for testing and building for multiple environments
> We can now target MacOS (+EULA) and/or nightly Rust branch!

- Instead of defining `SKIP_WASM_BUILD` before every command we can
  leverage the shared `env` definitions

- Instead of manually installing Rust toolchain we can utilize;

```yaml
      - name: Install Rust ${{ matrix.rust-branch }}
        uses: actions-rs/toolchain@v1.0.6
        with:
          toolchain: ${{ matrix.rust-branch }}
          profile: minimal
```

- Persistent caching of dependencies between similar builds _should_
  speed-up subsequent builds

```yaml
      - name: Utilize Rust shared cached
        uses: Swatinem/rust-cache@v2.2.1
        with:
          key: ${{ matrix.os }}-${{ env.RUST_BIN_DIR }}
```

> Note; as of last tests the above shaves a whole seven minutes when
> rebuild conditions are triggered... what a deal!

- In the future it may also be a good idea to utilize Action `inputs`
  and `outputs`, in combination with `sccache`, and further separate
  certain `name`ed `steps` into dedicated files


.github/workflows/check-rust.yml
```diff
@@ -0,0 +1,136 @@
...
+        rust-branch:
+          - stable
+
+        rust-target:
+          - x86_64-unknown-linux-gnu
+          # - x86_64-apple-darwin
+
+        os:
+          - ubuntu-latest
+          # - macos-latest
```
eduardogr, line 62: can we enable this also for macos?? That would be great

.github/workflows/check-rust.yml
```diff
@@ -0,0 +1,136 @@
...
+      SKIP_WASM_BUILD: 1
+      TARGET: ${{ matrix.rust-target }}
+
+    ##
+    steps:
+      - name: Check-out repository under $GITHUB_WORKSPACE
+        uses: actions/checkout@v2
+
+      - name: Set RELEASE mode
+        if: github.event_name == 'push' && github.ref == 'refs/heads/staging'
```
eduardogr: do we have a staging branch here?? I am not able to find it
We don't use a staging branch in this repository, is that correct?

eduardogr:
LGTM! I'll wait till we resolve the comments to approve but it looks nice!

.github/workflows/check-rust.yml
S0AndS0: Yup, :smiling_imp: that is most certainly the plan!
... But for now I wanna make sure this doesn't break for OSs we know to be good

.github/workflows/check-rust.yml
S0AndS0: :facepalm: oof, good catch!

And checking a [blog][1] entry from a few years ago, it seems we can now utilize `$default-branch` to avoid future churn.  I'll get this fixed in a few minutes for sure

[1]: https://github.blog/changelog/2020-07-22-github-actions-better-support-for-alternative-default-branch-names/

shibshib:
@S0AndS0 you will want to use nightly toolchain `nightly-2023-01-18` as per #115

S0AndS0:
> @S0AndS0 you will want to use nightly toolchain `nightly-2023-01-18` as per #115

Thanks @shibshib. I'll get that sorted _soon(:tm:)_!  Other than that anything else objectionable blocking this merge?

shibshib (approved):

PR #81, opened 2023-03-28
Increase burn and difficulty hyperparams

opentaco:
Increase burn and difficulty hyperparams


quac88 (commented):
Hey Taco, will you please give your reasoning for increasing the burn and POW diff?

Eugene-hu (approved):

Eugene-hu:
The burn and difficulty were increased to match the consistent level of Nakamoto prior to the switch to Finney. The POW difficulty matched the exact difficulty of Nakamoto at block 3606775. For the burn, we purposely increased it to an amount higher than expected. This will allow system to reach equilibrium over time without a sudden influx of burn registrations. This change was applied prior to the registration finney launch on March 23rd, 2023.

shibshib (approved):

PR #93, opened 2023-03-29
changed min burn to 1 tao in hyperparameter file.

SaMotlagh:


shibshib (approved):

shibshib (approved):

Eugene-hu (approved):

S0AndS0 (approved):

hyperparameters.md
```diff
@@ -32,7 +32,7 @@ MinDifficulty: u64 = 10_000_000;
 MaxDifficulty: u64 = u64::MAX / 4;
 ServingRateLimit: u64 = 250; // [250 @ 7,166]
 Burn: u64 = 100_000_000_000; // 100 tao [100 tao @ 26,310]
-MinBurn: u64 = 10_000_000_000; // 10 tao [10 tao @ 26,310]
+MinBurn: u64 = 1_000_000_000; // 1 tao [1 tao @ 26,310]
```
opentaco, line 35: The idea is to keep the change history in the comment by appending the new change:
```rust
MinBurn: u64 = 1_000_000_000; // 1 tao [10 tao @ 26,310] [1 tao @ set_block]
```

SaMotlagh: you are right. min burn is already set to 1 tao on the chain. It has never been set 10 tao. so I believe the block number that we set burn, we actually set it to 1 tao instead of 10 tao. Therefore, this change address that.

PR #95, opened 2023-03-31
[BIT-638] Max-upscale pruning_scores

opentaco:
Currently `pruning_scores` are a u16 quantization of normalized emission, which is problematic given ~4096 values and a Pareto distribution of emissions. Currently finney has a max emission of ~11% and a min non-zero emission of ~0.0002% (0.000002), but the lowest u16 quantity is 1/2^16 = 0.000015 (0.0015%).

The purpose of `pruning_scores` is to make relative comparisons during registration to determine the neuron to replace, which has the lowest `pruning_score`. Therefore, it doesn't have to be normalized, so this PR proposes max-upscaling where `max(pruning_scores) = u16::MAX`. So scale scores relative to the max score, so 0.0002/11 = 0.000018 (> 0.000015 u16 min quantity).

Assumes that `pruning_scores` is used only for relative comparison, and doesn't have to be normalized.

### DIFF
--- a/pallets/subtensor/src/math.rs
+++ b/pallets/subtensor/src/math.rs
@@ -37,6 +37,23 @@ pub fn vec_u16_proportions_to_fixed( vec: Vec<u16> ) -> Vec<I32F32> { vec.into_i
 pub fn vec_fixed_proportions_to_u16( vec: Vec<I32F32> ) -> Vec<u16> { vec.into_iter().map(|e| fixed_proportion_to_u16(e) ).collect() }
 
 #[allow(dead_code)]
+pub fn vec_max_upscale_to_u16( vec: &Vec<I32F32> ) -> Vec<u16> {
+    let u16_max: I32F32 = I32F32::from_num( u16::MAX );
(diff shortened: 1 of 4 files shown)


pallets/subtensor/src/math.rs
```diff
@@ -807,6 +832,42 @@ mod tests {
         vector.iter().map( | x | I32F32::from_num( *x ) ).collect()
     }
 
+    #[test]
```
S0AndS0, line 840: I like the quantity of tests!... though am somewhat concerned what'll happen if an max values for `I32F32` are passed, because at first glance it seems like that may be a panic condition

S0AndS0 (approved):

pallets/subtensor/src/math.rs
opentaco: Added handling in case of unnormalized values, and added a function comment that specifies assumption of non-negative normalized inputs.

camfairchild (approved):
LGTM

Eugene-hu (approved):

shibshib:
Main chain upgraded @ block 114,315

PR #96, opened 2023-04-03
Add more `weights.rs` unit tests

S0AndS0:


camfairchild (approved):
LGTM.

opentaco (approved):
Very comprehensive, LGTM.

Note that there's a PR in the works for max-upscaling weights, which means that weights won't be normalized for storage anymore, but instead will be max-upscaled.

PR #98, opened 2023-04-04
Removed unnecessary test functions from critical path

shibshib:
This removes unnecessary create network test functions as we already have add network, and it is since been deprecated.


Rubberbandits (approved):

shibshib:
As it's a simple function but it's an emergency, chain upgraded @block 107,345

S0AndS0 (approved):

eduardogr (approved):

PR #101, opened 2023-04-11
Unstake on dereg

camfairchild:
This PR:
- adds a few tests for the replace on neuron function
- adds functionality to unstake all coldkeys delegated to a neuron, when **fully** deregistered (not a member on any subnets)
     - adds tests for these
- adds new `staking::unstake_all_coldkeys_from_hotkey_account( hotkey: &T::AccountId )` function  
   - adds tests for this


pallets/subtensor/src/staking.rs
```diff
@@ -373,4 +374,23 @@ impl<T: Config> Pallet<T> {
...
+        for ( delegate_coldkey_i, stake_i ) in < Stake<T> as IterableStorageDoubleMap<T::AccountId, T::AccountId, u64 >>::iter_prefix( hotkey ) {
+            
+            // Convert to balance and add to the coldkey account.
+            let stake_i_as_balance = Self::u64_to_balance( stake_i );
+            if stake_i_as_balance.is_none() {
+                continue; // Don't unstake if we can't convert to balance.
+            } else {
+                // Stake is successfully converted to balance.
+                // Add the balance to the coldkey account.
+                Self::add_balance_to_coldkey_account( &delegate_coldkey_i, stake_i_as_balance.unwrap() );
```
S0AndS0: The use of `stake_i_as_balance` here vs `stake_i` further on (at line `391` has be a bit concerned.

> Also the order of operations may be better if reversed, though this could be my past JavaScript trauma talking here

pallets/subtensor/tests/staking.rs
```diff
@@ -993,3 +993,114 @@ fn test_full_with_delegating() {
 }
 
 
+/************************************************************
```
S0AndS0, line 1001: :+1: I really appreciate the comments that break-up sections of tests!

S0AndS0 (approved):

pallets/subtensor/src/staking.rs
camfairchild: > > Also the order of operations may be better if reversed, though this could be my past JavaScript trauma talking here  

Swapped the statements

> The use of `stake_i_as_balance` here vs `stake_i` further on (at line `391` has be a bit concerned.    

I'm not sure there's a better way to do that. afaik this is how we do it elsewhere, because the stake map is a `u64` and the balances are a `Balance`

SaMotlagh (approved):

pallets/subtensor/src/staking.rs
S0AndS0: > Swapped the statements

Thank you!

> I'm not sure there's a better way to do that. afaik this is how we do it elsewhere, because the stake map is a `u64` and the balances are a `Balance`

:thinking: oof, that is a bit troubling from my perspective...  Initially I thought, "why not cast `Balance` to `u64`", but diving into the underlying source for the `Balance` shows we may not be guaranteed loss-less conversion;

- `<substrate-cache>/primitives/arithmetic/src/traits.rs:151`

```rust
/// A meta trait for arithmetic.
///
/// Arithmetic types do all the usual stuff you'd expect numbers to do. They are guaranteed to
/// be able to represent at least `u32` values without loss, hence the trait implies `From<u32>`
/// and smaller integers. All other conversions are fallible.
pub trait AtLeast32Bit: BaseArithmetic + From<u16> + From<u32> {}
```

- `<substrate-cache>/primitives/arithmetic/src/traits.rs:161`

```rust
/// A meta trait for arithmetic.  Same as [`AtLeast32Bit `], but also bounded to be unsigned.
pub trait AtLeast32BitUnsigned: AtLeast32Bit + Unsigned {}
```

... Regardless this is smelling like a _yak shaving_ sort of thing, so I'll totally trust your implementation is solid.

S0AndS0 (approved):

shibshib (approved):

PR #104, opened 2023-04-13
Publish Git Tag (+ Rust target) as GitHub Release

S0AndS0:
:warning:s

- `artifacts` paths will require updating after Finney
- `bodyFile` path will need populated, and edited for every Git Tag

### DIFF
--- a/.github/workflows/publish-tag.yml
+++ b/.github/workflows/publish-tag.yml
@@ -1,0 +1,102 @@
+name: Rust build, benchmarks, and tests
+
+concurrency:
+  group: ci-${{ github.ref }}
+  cancel-in-progress: true
+
+##
+on:
+  ##
+  # Run when a semantic version is tagged
+  push:
+    tags:
+      - 'v[0-9]+.[0-9]+.[0-9]+'
+
+##
+# Environment variables shared for all targets
+env:
+  CARGO_TERM_COLOR: always
+  RELEASE: true
+  RELEASE_NAME: release
+  RUSTFLAGS: -A warnings
+  RUST_BACKTRACE: full
+  SKIP_WASM_BUILD: 1
+  VERBOSE: ${{ github.events.input.verbose }}
+
+##
+# Test and build and publish
+jobs:
+  check:
+    name: Tests and publishes targeting ${{ matrix.rust-target }} for OS ${{ matrix.os }}
+    runs-on: ${{ matrix.os }}
+
+    ##
+    # Define multiple targets for builds and tests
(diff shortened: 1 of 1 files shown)


.github/workflows/publish-tag.yml
Rubberbandits, line 1: I believe we should also include the wasm target for checks, and usually substrate has a few more dependencies that need to be downloaded aside from what is shipped via the cargo.toml. An installation script can be curl'd from `https://getsubstrate.io`

S0AndS0: Interesting, I'd think any third-party dependence _should_ be installed by clients separately and not bundled with our packaging automation.

> Side note; once #79 is merged it'll be possible to remove checks/tests in this Action, and instead make it dependent on tests/checks passing elsewhere... oh such stretchy goals :shrug:

.github/workflows/publish-tag.yml
```diff
@@ -0,0 +1,102 @@
...
+      ## TODO: double-check `artifacts` path(s) be correct
+      # :warning: values for the following must always match;
+      #   - RUST_BIN_DIR
+      #   - target/${{ matrix.rust-target }}/release/node-subtensor
+      - name: Create Release
+        uses: ncipollo/release-action@v1.12.0
+        with:
+          allowUpdates: false
+          artifactErrorsFailBuild: true
+          bodyFile: CHANGELOG.md
```
eduardogr, line 97: what is this for?

It is going to read from that file or to write the github release there aswell??

eduardogr (approved):
I have loved this usecase!
I think it makes a lot of sense to use a github action for this! 

Thanks for this!

.github/workflows/publish-tag.yml
S0AndS0: Yup!... According to the [`ncipollo/release-action`][1] Action;

> "An optional body file for the release. This should be the path to the file."

... Effectively the file will need to be updated as part of a Git Tag check-list, because that'll be published as the content for the GitHub Release, eg.

- [ ] Bump the correct [semantic version][2] component
- [ ] Update the `CHANGELOG.md` file with any relevant notes and/or references
- [ ] Tag via CLI `git tag --annotate vM.I.P --message 'Spiffy code-name'`
- [ ] Push via CLI `git push origin vM.I.P`
- [X] :robot: CI/CD takes over here to accept or reject release

[1]: https://github.com/ncipollo/release-action
[2]: https://semver.org/

Rubberbandits (approved):

PR #105, opened 2023-04-17
add collective pallet.

SaMotlagh:


Rubberbandits:
Latest merge conflict resolution includes upping spec_version to 121

pallets/subtensor/Cargo.toml
```diff
@@ -43,6 +46,8 @@ sp-tracing = { git = "https://github.com/paritytech/substrate", default-features
 parity-util-mem = { version = "0.11.0", features = ['primitive-types'] }
 rand = "0.8"
 sp-core = { git = "https://github.com/paritytech/substrate", default-features = false, branch = "polkadot-v0.9.39" }
+# Used for sudo decentralization
+pallet-collective = { version = "4.0.0-dev", default-features = false, git = "https://github.com/paritytech/substrate.git", branch = "polkadot-v0.9.39" }
```
shibshib: Not sure if you need it in both test and production dependencies. You likely only need it in the main dependencies. If you need something for testing and not production then you can keep it only in dev dependencies, afaik.

runtime/src/lib.rs
```diff
@@ -320,6 +321,42 @@ impl pallet_sudo::Config for Runtime {
 	type RuntimeCall = RuntimeCall;
 }
 
+// Configure collective pallet for council
+parameter_types! {
+	pub const CouncilMotionDuration: BlockNumber = 3 * MINUTES;
+	pub const CouncilMaxProposals: u32 = 100;
+	pub const CouncilMaxMembers: u32 = 100;
```
shibshib: I don't think we need that many council members initially, or for testing, can we limit it to 3 for now?

shibshib (approved):

S0AndS0 (approved):

pallets/collective/Cargo.toml
```diff
@@ -0,0 +1,47 @@
+[package]
+name = "pallet-collective"
+version = "4.0.0-dev"
+authors = ["Parity Technologies <admin@parity.io>"]
```
camfairchild: Should be opentensor? Or at least adding opentensor

pallets/collective/src/migrations/v4.rs
```diff
@@ -0,0 +1,148 @@
+// This file is part of Substrate.
+
+// Copyright (C) Parity Technologies (UK) Ltd.
+// SPDX-License-Identifier: Apache-2.0
+
+// Licensed under the Apache License, Version 2.0 (the "License");
```
camfairchild: Why do we need the migration? Does collective already exist in the chain storage?

camfairchild (commented):
Review in progress, just had some questions 
@Rubberbandits

pallets/collective/Cargo.toml
Rubberbandits: Probably should add it, it’s partially modified

pallets/collective/src/migrations/v4.rs
Rubberbandits: It doesn’t, leftover from forking the pallet. I believe migrations are only triggered when specified essentially.

camfairchild: > It doesn’t, leftover from forking the pallet. I believe migrations are only triggered when specified essentially.

Definitely should remove then

camfairchild (approved):

PR #108, opened 2023-04-18
[BIT-224] Quadratic voting

opentaco:
### [BIT-224] Quadratic voting

**Quadratic voting**: maximize consensus statistical power

**Problem**: Miners at times complain about relatively inaccurate and outdated consensus dominated by top stake validators.
**Cause**: Validator UIDs roughly sample the network at the same rate, yet the pareto high validators apply much more stake per miner-performance-info-sample and dominate consensus toward their limited sampling.
**Objective**: Maximize the info and samples (UIDs) per effective stake-unit while minimizing the loss of effective stake.
**Solution**: Quadratic voting or more generally power voting, where `(effective_stake)^(power) = input_stake`.
**Example**: 10x => 8x divs multiplier `800k vali => (800k)^(1/1.1) = 233k eff. stake` vs `80k vali => (80k)^(1/1.1) = 29k eff. stake`
**Optimum**: Uniformly distributed stake over validator UIDs, maximizes the overall validation samples per stake-unit.
**Result**: Largest stake validators will acquire and run multiple validator UIDs and split stake across them toward validator avg stake, to retain best divs/stake; greater competition for validator slots.

--
After this is merged, then concurrently with chain upgrade do merge https://github.com/opentensor/bittensor/pull/1273

### DIFF
--- a/pallets/subtensor/src/lib.rs
+++ b/pallets/subtensor/src/lib.rs
@@ -149,6 +149,8 @@ pub mod pallet {
 		type InitialScalingLawPower: Get<u16>;
 		#[pallet::constant] // Initial synergy scaling law power.
 		type InitialSynergyScalingLawPower: Get<u16>;
+		#[pallet::constant] // Initial quadratic voting power.
+		type InitialQuadraticVotingPower: Get<u16>;
 		#[pallet::constant] // Immunity Period Constant.
 		type InitialImmunityPeriod: Get<u16>;
 		#[pallet::constant] // Activity constant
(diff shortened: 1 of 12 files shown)


pallets/subtensor/src/lib.rs
```diff
@@ -1527,12 +1534,20 @@ pub mod pallet {
...
+			Self::do_sudo_set_quadratic_voting_power( origin, netuid, quadratic_voting_power )
+		}
+
+		#[pallet::call_index(47)]
 		#[pallet::weight((0, DispatchClass::Operational, Pays::No))]
 		pub fn sudo_set_total_issuance(origin: OriginFor<T>, total_issuance: u64 ) -> DispatchResult {
 			Self::do_set_total_issuance(origin, total_issuance)
 		}
 		
-		#[pallet::call_index(47)]
```
opentaco: @camfairchild may these call_index values be modified like this?

camfairchild: You should assign the next call index to your new sudo call. We should avoid reassigning call indices, as this is effectively the API call name

opentaco: Noted, fixed, thx.

camfairchild (approved):
Approved. 
Note: This change will break peers that don't update the python-side

PR #113, opened 2023-04-20
now accountId type is U256 in test env and all test cases support it.

SaMotlagh:
- changed AccountId type in test env from u64 -> U256.
- changed all test cases to accept AccountId as U256.

### DIFF
--- a/pallets/subtensor/tests/benchmarking.rs
+++ b/pallets/subtensor/tests/benchmarking.rs
@@ -1,5 +1,3 @@
-#![recursion_limit = "512"]
-
 use pallet_subtensor::Error;
 use frame_support::assert_ok;
 use frame_system::Config;
--- a/pallets/subtensor/tests/benchmarking.rs
+++ b/pallets/subtensor/tests/benchmarking.rs
@@ -1,3 +1,5 @@
+#![recursion_limit = "512"]
+
 use pallet_subtensor::Error;
 use frame_support::assert_ok;
 use frame_system::Config;
--- a/pallets/subtensor/tests/networks.rs
+++ b/pallets/subtensor/tests/networks.rs
@@ -5,6 +5,7 @@ use frame_support::weights::{GetDispatchInfo, DispatchInfo, DispatchClass, Pays}
 use frame_system::Config;
 use frame_support::{sp_std::vec};
 use frame_support::{assert_ok};
+use sp_core::U256;
 
 /*TO DO SAM: write test for LatuUpdate after it is set */
 
@@ -52,7 +53,7 @@ fn test_clear_min_allowed_weight_for_network() {
         let min_allowed_weight = 2;
         let tempo: u16 = 13;
         add_network(netuid, tempo, 0);
-	register_ok_neuron( 1, 55, 66, 0);
+	register_ok_neuron( 1, U256::from(55), U256::from(66), 0);
         SubtensorModule::set_min_allowed_weights(netuid, min_allowed_weight);
         assert_eq!(SubtensorModule::get_min_allowed_weights(netuid), 2);
         assert_ok!(SubtensorModule::do_remove_network(<<Test as Config>::RuntimeOrigin>::root(), netuid));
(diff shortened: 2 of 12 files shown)


S0AndS0 (approved):
So far looks as good as it can be given the circumstances and time constraints!

pallets/subtensor/tests/benchmarking.rs
```diff
@@ -1,3 +1,4 @@
+
```
eduardogr: needed extra line?

pallets/subtensor/tests/networks.rs
```diff
@@ -66,19 +67,19 @@ fn test_remove_uid_for_network() {
...
+        assert!(SubtensorModule::get_uid_for_net_and_hotkey(netuid, &U256::from(55)).is_ok());
         assert_eq!(neuron_id, 0);
-        register_ok_neuron( 1, 56, 67, 300000);
-        let neuron_uid = SubtensorModule::get_uid_for_net_and_hotkey(netuid, &56).unwrap();
+        register_ok_neuron( 1, U256::from(56), U256::from(67), 300000);
+        let neuron_uid = SubtensorModule::get_uid_for_net_and_hotkey(netuid, &U256::from(56)).unwrap();
         assert_eq!(neuron_uid, 1);
         assert_ok!(SubtensorModule::do_remove_network(<<Test as Config>::RuntimeOrigin>::root(), netuid));
-        assert!(SubtensorModule::get_uid_for_net_and_hotkey(netuid, &55).is_err());
+        assert!(SubtensorModule::get_uid_for_net_and_hotkey(netuid, &U256::from(55)).is_err());
```
eduardogr, line 82: just a general idea for _potential_ modifications of the same value.

If we encapsule this in a function the call this function `U256::from`, e.g:

```rust
// take this as pseudocode please 
fn build_value(value:...) {
  ...
  return &U256::from(value)
}
```

next changes will impact few lines and this behaviour is more localized in a single function instead of being all over the place.

As I said, i do not know if this makes sense in the context of the project or Rust, but I share with you this idea in case you think we can have another change like this eventually @SaMotlagh

pallets/subtensor/tests/sudo.rs
```diff
@@ -51,37 +51,37 @@ fn test_sudo_registration() {
...
+        assert_eq!( SubtensorModule::get_coldkey_balance( &U256::from(8) ), 11 );
+        assert_eq!( SubtensorModule::get_hotkey_for_net_and_uid( 0, 0).unwrap(), U256::from(0) );
+		assert_eq!( SubtensorModule::get_hotkey_for_net_and_uid( 0, 1).unwrap(), U256::from(1) );
+		assert_eq!( SubtensorModule::get_hotkey_for_net_and_uid( 0, 2).unwrap(), U256::from(2) );
+		assert_eq!( SubtensorModule::get_hotkey_for_net_and_uid( 0, 3).unwrap(), U256::from(3) );
+		assert_eq!( SubtensorModule::get_hotkey_for_net_and_uid( 0, 4).unwrap(), U256::from(4) );
+		assert_eq!( SubtensorModule::get_hotkey_for_net_and_uid( 0, 5).unwrap(), U256::from(5) );
+		assert_eq!( SubtensorModule::get_hotkey_for_net_and_uid( 0, 6).unwrap(), U256::from(6) );
+		assert_eq!( SubtensorModule::get_hotkey_for_net_and_uid( 0, 7).unwrap(), U256::from(7) );
+		assert_eq!( SubtensorModule::get_hotkey_for_net_and_uid( 0, 8).unwrap(), U256::from(8) );
```
eduardogr, line 80: out of curiosity... why is this tabbed?

pallets/subtensor/tests/benchmarking.rs
SaMotlagh: no :D I cleaned up but this line was left out.

eduardogr (approved):
LGTM! Very nice!

Take a look at the comments in case they make sense

pallets/subtensor/tests/networks.rs
camfairchild: Maybe something like a `AccountId::from_num()`

SaMotlagh: thanks, that's right! we could have sth like from_num(). This way is probably simpler to read as conversion is explicit though. If we need to change type in future again, I will wrap it inside a function.

pallets/subtensor/tests/sudo.rs
SaMotlagh: huum. in my vscode, I do not see any extra tabs and all lined are aligned. not sure why we see it different here.

eduardogr: this is the interpretation though.... sounds like it depends on the editor,  there is  probably some difference there

PR #117, opened 2023-04-27
Fix burn not reducing for Reg per interval =1

RD4Fun:
burn does not reduce when target registrations this interval == 1   
Special case of no registrations in interval and target =0 reduce both difficulty and burn if both == 0

### DIFF
--- a/pallets/subtensor/src/block_step.rs
+++ b/pallets/subtensor/src/block_step.rs
@@ -227,7 +227,7 @@ impl<T: Config> Pallet<T> {
                 let burn_registrations_this_interval: u16 = Self::get_burn_registrations_this_interval( netuid );
                 let target_registrations_this_interval: u16 = Self::get_target_registrations_per_interval( netuid );
                 // --- 5. Adjust burn + pow
-                // There are four cases to consider. A, B, C, D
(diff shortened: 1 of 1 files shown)


RD4Fun:
There is only one possibility that allows the burn to reduce
chance of only one (1) PoW with a long interval registration (it is at 225 blocks) is not probable

This code is only for the case when there is 1 registration per interval
I actually recommend that the minimum number of registrations per interval is set to 2 and increase interval length as required so the issue does not occur

Truth table for "registration target" =1

Reg Interval eq 1

S0AndS0:
Oh excellent catch, and I appreciate the addition of a truth-table!

As encoded I think the plan is sound, and the only thing I think worthwhile to add to this targeted Pull Request are, some unit-tests to mitigate future regression bugs from being introduced.

camfairchild (approved):
LGTM. thanks for contributing 🔥👍

shibshib (approved):

shibshib:
Looks good to me @RD4Fun but if you wish for it to be included in a chain upgrade then we must increase the spec_version inside runtime/src/lib.rs, else we can include it in the next chain upgrade.

PR #118, opened 2023-04-27
increased priority for set_weight call.

SaMotlagh:


S0AndS0 (approved):

pallets/subtensor/src/lib.rs
```diff
@@ -1581,7 +1581,8 @@ pub mod pallet {
 			if Uids::<T>::contains_key( netuid, &hotkey ) {
 				let uid = Self::get_uid_for_net_and_hotkey(netuid, &hotkey.clone()).unwrap();
 				let current_block_number: u64 = Self::get_current_block_as_u64();
-				return current_block_number - Self::get_last_update_for_uid(netuid, uid as u16);
+				let default_priority: u64 = current_block_number - Self::get_last_update_for_uid(netuid, uid as u16);
+				return default_priority + u32::max_value() as u64;
```
S0AndS0, line 1585: Only bit that makes me pause is; what'll happen if `default_priority` somehow ends up greater than `u32::max_value()`, possible overflow panic?

... Though looking at the math the `current_block_number` would need to be _kinda_ high

S0AndS0 (approved):

pallets/subtensor/src/lib.rs
shibshib: I think this is unlikely, because `current_block_number` has to be  9,223,372,036,854,775,807 which I highly doubt we will hit within our lifetimes :D

shibshib (approved):

shibshib:
Upgraded @block 273,326

PR #127, opened 2023-05-01
Eliminate compiler warnings for deadcode & unused imports/variables

Rubberbandits:


SaMotlagh (approved):

node/src/chain_spec.rs
```diff
@@ -15,15 +15,19 @@ use sp_core::crypto::Ss58Codec;
 // Specialized `ChainSpec`. This is a specialization of the general Substrate ChainSpec type.
 pub type ChainSpec = sc_service::GenericChainSpec<GenesisConfig>;
 
+// These functions are unused in production compiles, util functions for unit testing
+#[allow(dead_code)]
```
S0AndS0, line 19: Would it be possible to instead, or in addition, utilize [conditional compiler][1] trickery?

```rust
#[cfg(not(release))]
```

[1]: https://doc.rust-lang.org/reference/conditional-compilation.html

S0AndS0 (approved):

node/src/chain_spec.rs
Rubberbandits: It'll still throw a compiler warning even when compiling in debug

PR #131, opened 2023-05-03
[BIT-104] Max-upscale bonds

opentaco:
### [BIT-104] Max-upscale bonds

Now stores column-wise max-upscaled u16 bonds, instead of normalized u16 bonds, to maximize u16 utilization. With max-upscaled u16 bonds, we have `max_bond=u16::MAX` over each column (server). Bonds normalization is eventually done in epoch, but bonds storage can be unnormalized.


camfairchild (approved):
LGTM

pallets/subtensor/tests/epoch.rs
```diff
@@ -583,6 +583,298 @@ fn test_16384_graph_sparse() {
 	});
 }
 
+// Test bonds exponential moving average over a sequence of epochs.
+#[test]
+fn test_bonds() {
+	new_test_ext().execute_with(|| {
+		let sparse: bool = true;
+		let n: u16 = 8;
```
S0AndS0, line 591: What is `n`?
> I see it being used as a limit for the `for` loop, and as a value for `max_allowed_validators` related stuff too

pallets/subtensor/tests/epoch.rs
```diff
@@ -583,6 +583,298 @@ fn test_16384_graph_sparse() {
...
+		SubtensorModule::epoch( netuid, 1_000_000_000 ); // run first epoch to set allowed validators
+		run_to_block( 1 ); // run to next block to ensure weights are set on nodes after their registration block
+
+		// === Set weights [val->srv1: 0.1, val->srv2: 0.2, val->srv3: 0.3, val->srv4: 0.4]
+		for uid in 0..(n/2) as u64 {
+			assert_ok!(SubtensorModule::set_weights(RuntimeOrigin::signed(U256::from(uid)), netuid, ((n/2)..n).collect(), vec![ u16::MAX/4, u16::MAX/2, (u16::MAX/4)*3, u16::MAX], 0));
+		}
+		if sparse { SubtensorModule::epoch( netuid, 1_000_000_000 ); }
+		else { SubtensorModule::epoch_dense( netuid, 1_000_000_000 ); }
+		/*  n: 8
```
S0AndS0, line 625: :question: Any way to have these new commented bits be asserted as true by tests?

S0AndS0 (approved):

pallets/subtensor/tests/epoch.rs
opentaco: In these tests it typically refers to the size of the network.

pallets/subtensor/tests/epoch.rs
opentaco: It is possible, but the focus with `test_bonds` is just to check the bond carry-over across epochs, so the limited asserts suffice.

PR #133, opened 2023-05-05
Benchmark extrinsics

SaMotlagh:
- removed benchmark extrinsics with no origin
- removed benchmarking test file including tests for those benchmarking calls.

this PR addresses BIT 307.


S0AndS0 (approved):

pallets/subtensor/src/lib.rs
```diff
@@ -1536,42 +1536,11 @@ pub mod pallet {
 			Self::do_set_total_issuance(origin, total_issuance)
 		}
 
-		#[pallet::call_index(50)]
+		#[pallet::call_index(47)]
```
camfairchild:
```suggestion
		#[pallet::call_index(50)]
```

We shouldn't change the call indices for existing calls

camfairchild (changes requested):
One change

pallets/subtensor/src/lib.rs
SaMotlagh: huum, right. fixed.

camfairchild (approved):
LGTM 🔥🔥🔥

PR #135, opened 2023-05-08
[BIT-325] Disallow <= 0 in port argument of serve_* functions

Rubberbandits:
This addresses issue BIT-326 by creating a new function that delegates input validation for serve_prometheus and serve_axon extrinsics. Execution time shouldn't be impacted by much but still recommend benchmarking once this is merged into main.

### DIFF
--- a/pallets/subtensor/src/serving.rs
+++ b/pallets/subtensor/src/serving.rs
@@ -79,7 +79,7 @@ impl<T: Config> Pallet<T> {
         // --- 4. Get the previous axon information.
         let mut prev_axon = Self::get_axon_info( netuid, &hotkey_id );
         let current_block:u64 = Self::get_current_block_as_u64();
-        ensure!( Self::axon_passes_rate_limit( netuid, &prev_axon, current_block ), Error::<T>::ServingRateLimitExceeded );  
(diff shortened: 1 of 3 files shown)


pallets/subtensor/src/serving.rs
```diff
@@ -263,5 +274,20 @@ impl<T: Config> Pallet<T> {
         return true;
     }
 
+	pub fn validate_axon_data(axon_info: &AxonInfoOf) -> Result<bool, pallet::Error<T>> {
+		if axon_info.port <= 0 {
```
S0AndS0: :warning: If the [internet][1] is to be believed, then the highest port supported by IPv4 addresses _should_ be `65_535`

While it be true that `port` for `AxonInfo` is correctly defined as `u16` (`pallets/subtensor/src/lib.rs:319`) I'm not fully certain of how things will behave if someone feeds a larger number during execution.

[1]: https://stackoverflow.com/questions/113224/what-is-the-largest-tcp-ip-network-port-number-allowable-for-ipv4#113228

S0AndS0 (approved):

pallets/subtensor/src/serving.rs
Rubberbandits: Will check for overflow, expect a commit momentarily.

Rubberbandits: Resolved

camfairchild (approved):
LGTM

PR #141, opened 2023-05-19
add contracts layer

cuteolaf:
To enable smart contracts on the Bittensor network, I've added pallet-contracts into the runtime.

@S0AndS0 @Rubberbandits @shibshib @camfairchild

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -4479,6 +4479,7 @@ dependencies = [
  "pallet-contracts-proc-macro",
  "parity-scale-codec",
  "rand 0.8.5",
+ "rand_pcg",
  "scale-info",
  "serde",
  "smallvec",
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2810,6 +2810,12 @@ dependencies = [
 ]
 
 [[package]]
+name = "indexmap-nostd"
+version = "0.4.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "8e04e2fd2b8188ea827b32ef11de88377086d690286ab35747ef7f9bf3ccb590"
+
+[[package]]
 name = "instant"
 version = "0.1.12"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -4190,6 +4196,8 @@ dependencies = [
  "frame-try-runtime",
  "pallet-aura",
  "pallet-balances",
+ "pallet-contracts",
+ "pallet-contracts-primitives",
  "pallet-grandpa",
  "pallet-insecure-randomness-collective-flip",
  "pallet-subtensor",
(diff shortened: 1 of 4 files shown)


unconst:
Hi @cuteolaf it looks like you ran a reformat over the entire lib.rs file making it difficult for us to evaluate what changes were made on that file. Could you revert the indentation changes you made please and resubmit?

Rubberbandits:
I've gone over the PR and there are two main concerns with implementing smart contracts at this point in time:

1. DoS via fee dynamics/heavy extrinsics. While under PoA, transaction fees are as minimal as possible, and while they are calculated based on extrinsic weight, the fee curve has been minimized as much as possible.
2. The CallFilter member in the config for the pallet is set to `Nothing`, which disallows smart contracts from calling any extrinsic. Smart contracts wouldn't be able to transfer balances or really do anything on the network with this filter type. CallFilter should act like a whitelist in order to give contracts enough functionality to be useful, but to also prevent them from calling extrinsics that could be used to harm the network if abused. The [pallet's documentation](https://paritytech.github.io/substrate/master/pallet_contracts/pallet/trait.Config.html#associatedtype.CallFilter) has a relevant section which explains why:

> The runtime must make sure that all dispatchables that are callable by contracts remain stable. In addition [Self::RuntimeCall](https://paritytech.github.io/substrate/master/pallet_contracts/pallet/trait.Config.html#associatedtype.RuntimeCall) itself must remain stable. This means that no existing variants are allowed to switch their positions. Note that dispatchables that are called via contracts do not spawn their own wasm instance for each call (as opposed to when called via a transaction). Therefore please make sure to be restrictive about which dispatchables are allowed in order to not introduce a new DoS vector like memory allocation patterns that can be exploited to drive the runtime into a panic.

Smart contract design and implementation needs to be properly fleshed out for Bittensor specifically before we can include it in the runtime. It's on our list of things to do, but this is a good starting point for us to deep-dive into how we want to implement smart contracts on the network.

cuteolaf:
> 

@unconst I've reverted the indentation changes.

S0AndS0:
> @cuteolaf
> To enable smart contracts on the Bittensor network, I've added pallet-contracts into the runtime.
> 
> @S0AndS0 ...

:wave: Howdy, I think it be really cool you're contributing to this project!... Though I believe it prudent, so ya don't feel ignored, to forewarn my attentions are currently elsewhere.

---

> @unconst
> ... reformat over the entire lib.rs file making it difficult for us to evaluate what changes were made on that file...

Git `diff` Tip, there be some semaphores that are useful for ignoring certain non-consequential changes;

**Relevant documentation**
- `man --pager='less --pattern="--ignore-all-space"' git-diff`
- `man -P 'less -p "--ignore-space-change"' git-diff`
- `man -P 'less -p "--ignore-blank-lines"' git-diff`

**Example** (what I use daily)
```bash
git diff \
  --ignore-all-space \
  --ignore-blank-lines \
  runtime/src/lib.rs
```

camfairchild:
Closing until this is decided. Feel free to open an issue we can use to track further

PR #145, opened 2023-06-05
Fix max-uids decrease test and add Jira ticket commit msg script

Rubberbandits:
## Description

Fix a broken test which prevented `cargo-husky` from installing a new git hook.
Include new git hook for `prepare-commit-msg` which inserts the current jira ticket's tag into your commit message based on branch name.

______

## Checklist

- [X] Tracked `rustfmt` changes
- [X] No _new_ compiler warnings

### DIFF
--- a/.cargo-husky/hooks/prepare-commit-msg
+++ b/.cargo-husky/hooks/prepare-commit-msg
@@ -1,0 +1,18 @@
+#!/bin/sh
+#
+# git prepare-commit-msg hook for automatically prepending an issue key
+# from the start of the current branch name to commit messages.
+
+# check if commit is merge commit or a commit ammend
+if [ $2 = "merge" ] || [ $2 = "commit" ]; then
+    exit
+fi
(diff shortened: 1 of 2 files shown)


.cargo-husky/hooks/prepare-commit-msg
```diff
@@ -0,0 +1,18 @@
+#!/bin/sh
```
eduardogr, line 1: this is very nice!! i will add this to some repositories i have! very nice   @Rubberbandits 

out of curiosity, how the final message looks like?

eduardogr (approved):

shibshib (approved):

Rubberbandits:
> this is very nice!! i will add this to some repositories i have! very nice   @Rubberbandits 

out of curiosity, how the final message looks like? 

Check out the messages in the commits of this PR! The first line gets the branch inserted automatically with the colon, then you just type your normal commit message and push.

PR #150, opened 2023-06-07
Create deterministic wasm builds and publish release on tag

Rubberbandits:


camfairchild (approved):

.github/workflows/publish-tag.yml
```diff
@@ -27,49 +27,45 @@ env:
...
           - x86_64-unknown-linux-gnu
-          - x86_64-apple-darwin
 
         os:
           - ubuntu-latest
-          - macos-latest
 
         include:
           - os: ubuntu-latest
-          - os: macos-latest
```
eduardogr: can't we build for both ??? and include them all in the release?

Rubberbandits: > can't we build for both ??? and include them all in the release?

Currently macos builds fail due to trying to install deps with apt. We should be using substrate's install script instead of manually installing deps as it will detect platform and install properly for us.

shibshib:
Quick note. I think the new nightlies are fixed and cargo should recompile, perhaps let's pick a newer version of nightly?

PR #152, opened 2023-06-23
Add Multisignature pallet to runtime

Rubberbandits:
This PR adds `pallet-multisig` to the runtime, enabling support for users to create multisig wallets through the `polkadot.js` explorer's Accounts page.

### DIFF
--- a/node/src/chain_spec.rs
+++ b/node/src/chain_spec.rs
@@ -178,7 +178,7 @@ pub fn finney_mainnet_config() -> Result<ChainSpec, String> {
 		Some("bittensor"),
 		None,
 		// Properties
-		None,
+		Some(properties),
 		// Extensions
 		None,
 	))
@@ -280,7 +280,7 @@ pub fn finney_testnet_config() -> Result<ChainSpec, String> {
 		Some("bittensor"),
 		None,
 		// Properties
-		None,
+		Some(properties),
 		// Extensions
 		None,
 	))
--- a/runtime/Cargo.toml
+++ b/runtime/Cargo.toml
@@ -46,6 +46,9 @@ sp-std = { version = "5.0.0", default-features = false, git = "https://github.co
 sp-transaction-pool = { version = "4.0.0-dev", default-features = false, git = "https://github.com/paritytech/substrate.git", branch = "polkadot-v0.9.39" }
 sp-version = { version = "5.0.0", default-features = false, git = "https://github.com/paritytech/substrate.git", branch = "polkadot-v0.9.39" }
 
+# Multisig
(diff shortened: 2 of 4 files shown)


camfairchild (approved):
LGTM

node/src/chain_spec.rs
```diff
@@ -178,7 +178,7 @@ pub fn finney_mainnet_config() -> Result<ChainSpec, String> {
 		Some("bittensor"),
 		None,
 		// Properties
-		None,
+		Some(properties),
```
shibshib, line 181: Does changing this still make a difference? I thought `finnet_mainnet_config` is really only relevant for the genesis block

runtime/Cargo.toml
```diff
@@ -46,6 +46,9 @@ sp-std = { version = "5.0.0", default-features = false, git = "https://github.co
 sp-transaction-pool = { version = "4.0.0-dev", default-features = false, git = "https://github.com/paritytech/substrate.git", branch = "polkadot-v0.9.39" }
 sp-version = { version = "5.0.0", default-features = false, git = "https://github.com/paritytech/substrate.git", branch = "polkadot-v0.9.39" }
 
+# Multisig
```
shibshib, line 49: ✅

runtime/src/lib.rs
```diff
@@ -320,6 +320,24 @@ impl pallet_sudo::Config for Runtime {
 	type RuntimeCall = RuntimeCall;
 }
 
+parameter_types! {
+	// One storage item; key size is 32; value is size 4+4+16+32 bytes = 56 bytes.
+	pub const DepositBase: Balance = (1) as Balance * 2_000 * 10_000_000 + (88 as Balance) * 100 * 1_000_000;
+	// Additional storage item size of 32 bytes.
+	pub const DepositFactor: Balance = (0) as Balance * 2_000 * 10_000_000 + (32 as Balance) * 100 * 1_000_000;
```
shibshib, line 327: Do we know what these two do? Guessing `DespositBase` is the base amount of currency needed to reserve to create a multisig?

shibshib (approved):

shibshib:
I'm good with this as well. Merge when ready @Rubberbandits

node/src/chain_spec.rs
Rubberbandits: Technically yes, but the properties value in the chain-spec isn't considered apart of genesis so it can be changed on the nodes without having to restart chain history.

runtime/src/lib.rs
Rubberbandits: Yes,  this is the implementation of the deposit function from polkadot's runtime. The deposit function has two arguments, items and bytes. It's an ease-of-use function which is used to calculate per-byte fees when storage data is used on chain.

PR #153, opened 2023-06-23
Add local chain to chain_spec for fast testing with localnet

Rubberbandits:
Previously using `test_finney` or `finney` chains when generating a chain-spec, Alice and Bob keys aren't inserted as initial authorities. This PR adds a third chain, `local` which has the correct initial authorities and balances for all development keys in `polkadot.js`

### DIFF
--- a/node/src/chain_spec.rs
+++ b/node/src/chain_spec.rs
@@ -479,7 +479,7 @@ fn localnet_genesis(
 			authorities: initial_authorities.iter().map(|x| (x.1.clone(), 1)).collect(),
 		},
 		sudo: SudoConfig {
-			key: Some(Ss58Codec::from_ss58check("5GpzQgpiAKHMWNSH3RN4GLf96GVTDct9QxYEFAY7LWcVzTbx").unwrap()),
+			key: Some(get_account_id_from_seed::<sr25519::Public>("Alice")),
 		},
 		transaction_payment: Default::default(),
 		subtensor_module: Default::default(),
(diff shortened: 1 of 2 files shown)


node/src/chain_spec.rs
```diff
@@ -363,3 +363,125 @@ fn finney_genesis(
...
+			]
+		},
+		aura: AuraConfig {
+			authorities: initial_authorities.iter().map(|x| (x.0.clone())).collect(),
+		},
+		grandpa: GrandpaConfig {
+			authorities: initial_authorities.iter().map(|x| (x.1.clone(), 1)).collect(),
+		},
+		sudo: SudoConfig {
+			key: Some(Ss58Codec::from_ss58check("5GpzQgpiAKHMWNSH3RN4GLf96GVTDct9QxYEFAY7LWcVzTbx").unwrap()),
```
camfairchild: Is this for local development? What key is this?

shibshib (approved):

node/src/chain_spec.rs
Rubberbandits: This is the same as finney testnet sudo

camfairchild:
```suggestion
			key: Some(get_account_id_from_seed::<sr25519::Public>("Alice")),
```

camfairchild (changes requested):

camfairchild (approved):

PR #157, opened 2023-07-07
Adjusts POW and BURN based on a moving average

unconst:
Adjusts the burn and POW via a moving average.


Rubberbandits (commented):
Does this change how unit testing should verify how much burn changes per register?

opentaco (approved):

opentaco (commented):
Some outstanding fixes for unit tests required, looking at it now.

PR #163, opened 2023-07-11
Move event deposits into delegated storage setters

Rubberbandits:
## Description

This change moves all storage-setter related event deposits into the delegated function, allowing events to be emitted when the runtime changes storage values outside of sudo calls.

______

## Checklist

- [X] No _new_ compiler warnings


camfairchild (changes requested):
We should add a check to each test for these calls that each event is emitted, as expected.
This was from ToB feedback

Rubberbandits:
> We should add a check to each test for these calls that each event is emitted, as expected. This was from ToB feedback

This is true, and is the standard for pallets from Paritytech. They have some code in the collective pallet for checking emitted event history which is super helpful.

fabhed:
Love that this is being worked on so promptly after I mentioned it in Discord. This will make indexing the chain much nicer!

camfairchild (approved):

PR #172, opened 2023-07-31
feat: add identity pallet to the runtime

cuteolaf:


cuteolaf:
@S0AndS0 @shibshib @Rubberbandits @camfairchild

camfairchild:
@unconst @shibshib what do you think? I think this may be a good feature

cuteolaf:
I believe that the community will like this feature

runtime/src/lib.rs
```diff
@@ -553,6 +559,30 @@ impl pallet_preimage::Config for Runtime {
     type ByteDeposit = PreimageByteDeposit;
 }
 
+parameter_types! {
+	pub const BasicDeposit: Balance = (1) as Balance * 2_000 * 10_000_000 + (258 as Balance) * 100 * 1_000_000;
```
camfairchild, line 563: Why is the calculation like this? Is there some documentation that can be added?

runtime/src/lib.rs
```diff
@@ -553,6 +559,30 @@ impl pallet_preimage::Config for Runtime {
...
+	pub const FieldDeposit: Balance = (0) as Balance * 2_000 * 10_000_000 + (66 as Balance) * 100 * 1_000_000;
+	pub const SubAccountDeposit: Balance = (1) as Balance * 2_000 * 10_000_000 + (53 as Balance) * 100 * 1_000_000;
+	pub const MaxSubAccounts: u32 = 100;
```
camfairchild, line 566: Same for here

runtime/src/lib.rs
```diff
@@ -553,6 +559,30 @@ impl pallet_preimage::Config for Runtime {
...
+	type Currency = Balances;
+	type BasicDeposit = BasicDeposit;
+	type FieldDeposit = FieldDeposit;
+	type SubAccountDeposit = SubAccountDeposit;
+	type MaxSubAccounts = MaxSubAccounts;
+	type MaxAdditionalFields = MaxAdditionalFields;
+	type MaxRegistrars = MaxRegistrars;
+	type Slashed = ();
+	type ForceOrigin = MoreThanHalfCouncil;
+	type RegistrarOrigin = MoreThanHalfCouncil;
```
camfairchild, line 582: So then (currently) adding a registrar is a 2 of 3 from the triumvirate? 
Wonder if using the Validator set makes more sense? Or perhaps an additional council group can be formed.

camfairchild: @unconst @shibshib

camfairchild (changes requested):
Left some comments. Thanks for the PR

runtime/src/lib.rs
cuteolaf: This is from the polkadot runtime.
https://github.com/paritytech/polkadot/blob/d74c60b6489f31f437edb1a19de0d0f6b8ad432e/runtime/polkadot/src/lib.rs#L620

runtime/src/lib.rs
cuteolaf: same
https://github.com/paritytech/polkadot/blob/d74c60b6489f31f437edb1a19de0d0f6b8ad432e/runtime/polkadot/src/lib.rs#L621-L622

runtime/src/lib.rs
cuteolaf: this should be a team decision, i believe.

PR #174, opened 2023-08-08
Testnet config

shibshib:
This creates a config for testing subtensor locally. Everything is set up with pre-made accounts and so blocks finalize right away, no need to insert keys and etc. Simply run it with 

```
./target/release/node-subtensor --dev --tmp
```

`--dev` will specify this localized testnet setup. 
`--tmp` will ensure the node is ephemeral and all its resources deleted when it's stopped.


Rubberbandits:
I implemented this within the governance update, it's included in the localnet script setup.

shibshib:
Closed as it's a dup.

PR #181, opened 2023-08-23
Add BondsPenalty hyperparam

opentaco:
### Add BondsPenalty hyperparam

Given recent issues with (partially) blacklisted validators getting poor dividends, we've been exploring the possibility of adjusting the validation bonds penalty to improve their dividends.

**Bonds penalty**: Weight higher than consensus gets clipped, and bonds calculated from this then also get clipped, which means dividends get clipped. Cabal that sets weights on its own poor servers have their bonds cut and receive less dividends.
**Validation advantage**: Setting weight on the highest performing servers, means that dividends will be relatively better. Cabal that sets weights on its own poor servers will get less dividends because those servers get less incentive.

**Suggestion**: Adjust bonds penalty per netuid, with a default zero penalty, keep validation advantage.
**Expected result**: Partially blacklisted validators that set more weights on a fewer set of servers will enjoy full bonds on those, instead of getting bonds cut like when validation penalty applies.

**Security guarantees**:
A foremost concern is how this will affect the consensus security guarantees, so we simulate that as part of the subtensor integration tests with the `_map_consensus_guarantees()` function, and do a comparison with/without bonds penalty.
We observe as slight erosion of guarantees, at one point quantified as a 73% -> 79% honest utility increase required @ 60% honest stake to retain at least 60% emission. This is acceptable given that we gain significant alleviation of poor dividends due to blacklisting.

### DIFF
--- a/pallets/subtensor/src/epoch.rs
+++ b/pallets/subtensor/src/epoch.rs
@@ -131,18 +131,22 @@ impl<T: Config> Pallet<T> {
         // Compute preranks: r_j = SUM(i) w_ij * s_i
         let preranks: Vec<I32F32> = matmul( &weights, &active_stake );
 
-        // Clip weights at majority consensus
-        let kappa: I32F32 = Self::get_float_kappa( netuid );  // consensus majority ratio, e.g. 51%.
+        // Consensus majority ratio, e.g. 51%.
+        let kappa: I32F32 = Self::get_float_kappa( netuid );
+        // Calculate consensus as stake-weighted median of weights.
         let consensus: Vec<I32F32> = weighted_median_col( &active_stake, &weights, kappa );
-        inplace_col_clip( &mut weights, &consensus );
-        let validator_trust: Vec<I32F32> = row_sum( &weights );
+        // Clip weights at majority consensus.
+        let mut clipped_weights: Vec<Vec<I32F32>> = weights.clone();
+        inplace_col_clip( &mut clipped_weights, &consensus );
+        // Calculate validator trust as sum of clipped weights set by validator.
+        let validator_trust: Vec<I32F32> = row_sum( &clipped_weights );
 
         // ====================================
         // == Ranks, Server Trust, Incentive ==
         // ====================================
 
         // Compute ranks: r_j = SUM(i) w_ij * s_i
-        let mut ranks: Vec<I32F32> = matmul( &weights, &active_stake );
+        let mut ranks: Vec<I32F32> = matmul( &clipped_weights, &active_stake );
 
         // Compute server trust: ratio of rank after vs. rank before.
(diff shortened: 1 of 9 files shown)


pallets/subtensor/src/epoch.rs
```diff
@@ -155,14 +159,21 @@ impl<T: Config> Pallet<T> {
         // == Bonds and Dividends ==
         // =========================
 
+        // Get validator bonds penalty in [0, 1].
+        let bonds_penalty: I32F32 = Self::get_float_bonds_penalty( netuid );
+        // Calculate weights for bonds, apply bonds penalty to weights.
+        // bonds_penalty = 0: weights_for_bonds = weights.clone()
+        // bonds_penalty = 1: weights_for_bonds = clipped_weights.clone()
+        let weights_for_bonds: Vec<Vec<I32F32>> = interpolate( &weights, &clipped_weights, bonds_penalty);
```
Eugene-hu, line 167: So if the bonds penalty = 0.5, will the interpolate function return an average between weights and clipped weights?

Eugene-hu (approved):
LGTM; when do you think we can release this update taco?

pallets/subtensor/src/epoch.rs
opentaco: Yes, that is correct, the unit tests for `interpolate` has examples of this.

opentaco:
> LGTM; when do you think we can release this update taco?

This is ready for release after outstanding approvals.
@shibshib please merge before the next chain upgrade.

opentaco:
> LGTM; when do you think we can release this update taco?

Once outstanding approvals are granted, it can be merged.
@shibshib please merge before the next chain upgrade.

pallets/subtensor/src/lib.rs
```diff
@@ -1959,6 +1969,18 @@ pub mod pallet {
         ) -> DispatchResult {
             Self::do_sudo_set_adjustment_alpha(origin, netuid, adjustment_alpha)
         }
+
+        #[pallet::call_index(59)]
+        #[pallet::weight((Weight::from_ref_time(14_000_000)
```
Rubberbandits, line 1974: Was this benchmarked? If not, will add it to the list of features that need benchmarking before we push latest chain upgrade.

camfairchild: ~~It was not~~ Edit: not my PR, oops

Rubberbandits (approved):

pallets/subtensor/src/lib.rs
opentaco: > Was this benchmarked? If not, will add it to the list of features that need benchmarking before we push latest chain upgrade.

No it wasn't benchmarked, but copied the time of another u16 hyperparam.

PR #183, opened 2023-08-24
Update `TotalStake` in genesis handler

cuteolaf:
`TotalStake` is not updated in genesis handler, so the value returned by `api.query.subtensorModule.totalStake()` is not correct.

### DIFF
--- a/pallets/subtensor/src/lib.rs
+++ b/pallets/subtensor/src/lib.rs
@@ -949,6 +949,9 @@ pub mod pallet {
                     // Update total issuance value
                     TotalIssuance::<T>::put(TotalIssuance::<T>::get().saturating_add(*stake));
 
+                    // Update total stake value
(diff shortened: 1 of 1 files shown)


cuteolaf:
@S0AndS0 @Rubberbandits @shibshib @camfairchild

pallets/subtensor/src/lib.rs
```diff
@@ -949,6 +949,9 @@ pub mod pallet {
                     // Update total issuance value
                     TotalIssuance::<T>::put(TotalIssuance::<T>::get().saturating_add(*stake));
 
+                    // Update total stake value
+                    TotalStake::<T>::put(TotalStake::<T>::get().saturating_add(*stake));
+
```
camfairchild, line 954: Might be good to leave a comment here mentioning that the genesis block was not produced using this. i.e. this was added afterwards, so the genesis state would be different for a new instance of the chain.

camfairchild (changes requested):

PR #189, opened 2023-10-12
POW Registration allowance/disallowance

Rubberbandits:
This PR adds a sudo call which separately manages whether or not POW registration is enabled for a subnet.

### DIFF
--- a/pallets/subtensor/src/registration.rs
+++ b/pallets/subtensor/src/registration.rs
@@ -131,7 +131,7 @@ impl<T: Config> Pallet<T> {
 
         // --- 3. Ensure the passed network allows registrations.
         ensure!(
-            Self::if_subnet_allows_registration(netuid),
+            Self::get_network_registration_allowed(netuid),
             Error::<T>::RegistrationDisabled
         );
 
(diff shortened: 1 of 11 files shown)


pallets/subtensor/src/registration.rs
```diff
@@ -317,7 +317,7 @@ impl<T: Config> Pallet<T> {
 
         // --- 3. Ensure the passed network allows registrations.
         ensure!(
-            Self::if_subnet_allows_registration(netuid),
+            Self::get_network_pow_registration_allowed(netuid),
```
eduardogr, line 320: is there some idiomatic reason for this name change?
I thjink the previous one has a good naming, maybe something like.... `Self::subnet_allows_registrations(netuid)` which could fit with the context and its semantics. What do you think?

Rubberbandits: if_subnet_allows_registration was a duplicate of another function which was already implemented.

ifrit98 (approved):
LGTM! Only question is on some failing tests. Assume these panics are from flaky tests or not updated?

image

Rubberbandits:
> LGTM! Only question is on some failing tests. Assume these panics are from flaky tests or not updated?
> 
> image

Yep, i do need to update these! Good catch

Rubberbandits:
Burn adjustment tests fixed

ifrit98 (approved):
LGTM! 

Only issue is the `test_registration_difficulty_adjustment` fails.

Rubberbandits:
> LGTM!
> 
> Only issue is the `test_registration_difficulty_adjustment` fails.

Fixed! Thank you for catching this.

shibshib (approved):

PR #191, opened 2023-10-24
Identity system for Subnets and Root-validators

Rubberbandits:
This PR implements `pallet-registry`, a new system which allows validators and subnet creators to commit information about themselves to the chain itself. This will replace the delegates.json list by enforcing all data on-chain.


ifrit98 (approved):
LGTM! 

Question: is the max data length for `create_identity_info` 32bytes? 


```python
let data = Data::Raw(vec![0; 32].try_into().unwrap());
```


I see that `pgp_fingerprint` is required to have 20bytes, are there other requirements for the rest of the fields? Want to make a note of any requirements on the bittensor API side.

ifrit98:
Small nit:

Assuming these tests just need to be updated?

```rs
error[E0432]: unresolved import `crate::mock`
 --> pallets/registry/src/tests.rs:1:13
  |
1 | use crate::{mock::*, Error, Event};
  |             ^^^^ could not find `mock` in the crate root

error[E0616]: field `id` of struct `UntrackedSymbol` is private
   --> pallets/registry/src/types.rs:369:44
    |
369 |         let type_info = registry.resolve(type_id.id).unwrap();
    |                                                  ^^ private field
    |
help: a method `id` also exists, call it with parentheses
    |
369 |         let type_info = registry.resolve(type_id.id()).unwrap();
    |                                                    ++

error[E0616]: field `type_def` of struct `Type` is private
   --> pallets/registry/src/types.rs:380:62
    |
380 |             if let scale_info::TypeDef::Variant(variant) = &type_info.type_def {
    |                                                                       ^^^^^^^^ private field
    |
help: a method `type_def` also exists, call it with parentheses
    |
380 |             if let scale_info::TypeDef::Variant(variant) = &type_info.type_def() {
    |                                                                               ++

error[E0616]: field `variants` of struct `TypeDefVariant` is private
   --> pallets/registry/src/types.rs:382:7
    |
382 |                     .variants
    |                      ^^^^^^^^ private field
    |
help: a method `variants` also exists, call it with parentheses
    |
382 |                     .variants()
    |                              ++

error[E0616]: field `len` of struct `TypeDefArray` is private
   --> pallets/registry/src/types.rs:393:12
    |
393 | ...                   arr.len
    |                           ^^^ private field
    |
help: a method `len` also exists, call it with parentheses
    |
393 |                             arr.len()
    |                                    ++
```

Rubberbandits:
> LGTM!
> 
> Question: is the max data length for `create_identity_info` 32bytes?
> 
> ```python
> let data = Data::Raw(vec![0; 32].try_into().unwrap());
> ```
> 
> I see that `pgp_fingerprint` is required to have 20bytes, are there other requirements for the rest of the fields? Want to make a note of any requirements on the bittensor API side.

The way the arguments work for the different fields are like so:

You have a struct of different fields, website, email, etc and each of them can be the type of `enum Data`
This Data enum has a bunch of different variants, all the way from Raw0 to Raw32 bytes, Keccak256, Blake256, and SHA256 hashes. In the specific case of a pgp fingerprint, you'd want to use enum type "Raw20"

shibshib (approved):

PR #195, opened 2023-11-10
Ostraca

unconst:


pallets/subtensor/src/lib.rs
```diff
@@ -2033,36 +2037,38 @@ pub mod pallet {
...
-                interval,
-            ));
+            log::info!("NetworkLockReductionInterval( interval: {:?} ) ", interval);
+            Self::deposit_event(Event::NetworkLockCostReductionIntervalSet(interval));
 
             Ok(())
         }
+
+        #[pallet::call_index(69)]
+        #[pallet::weight((Weight::from_ref_time(14_000_000)
```
Rubberbandits, line 2077: This will need to be benchmarked, as it seems the ostraca function is going to be pretty read/write heavy as it iterates through all stakers for a hotkey.

Rubberbandits (changes requested):

Rubberbandits (approved):
After adding dynamic weight to ostraca function, and fixing unit testing, LGTM

welikethestock (approved):
Good to go

PR #199, opened 2023-11-16
Adds a Liquid alpha switch

unconst:
Adds the ability to turn on liquid alpha which makes bonds speculative


Rubberbandits:
We need to add testing for this PR

welikethestock (dismissed):

PR #202, opened 2023-11-22
Fix broken test registry/types/manual_data_type_info

Rubberbandits:
This PR updates leading byte numbers in registry/types encode implementation, which while encoding data, resulted in broken decoding as it expected a different leading-byte to indicate the type.


ifrit98 (approved):
LGTM

Eugene-hu (approved):

PR #210, opened 2023-12-13
Localnet/specify init wallet balance

ifrit98:
* Add a default balance of million TAO to an ss58 address of choice provided in env var `BT_DEFAULT_TOKEN_WALLET`
* This env var can be set easily by parsing a wallet locally: `BT_DEFAULT_TOKEN_WALLET=$(cat ~/.bittensor/wallets/<WALLETNAME>/coldkeypub.txt | grep -oP '"ss58Address": "\K[^"]+')`
* Faucet reenabled by argument
* Features can now be enabled by argument in `localnet.sh`


runtime/src/lib.rs
```diff
@@ -625,7 +625,7 @@ parameter_types! {
     pub const SubtensorInitialSenateRequiredStakePercentage: u64 = 1; // 1 percent of total stake
     pub const SubtensorInitialNetworkImmunity: u64 = 7 * 7200;
     pub const SubtensorInitialMinAllowedUids: u16 = 128;
-    pub const SubtensorInitialMinLockCost: u64 = 1_000_000_000_000; // 1000 TAO
+    pub const SubtensorInitialMinLockCost: u64 = 100_000_000_000; // 100 TAO
```
Rubberbandits: This will be a problem for future runtime upgrades, we need to have two of these definitions with their own compiler flags to only insert them under specific conditions.

Rubberbandits (approved):

runtime/src/lib.rs
ifrit98: Ah, makes sense. I can revert if that's simples for now. With the `BT_DEFAULT_TOKEN_WALLET` env var, this becomes more convenience.

ifrit98: Reverted

camfairchild (approved):
lgtm :+1:

PR #213, opened 2023-12-15
update `return_per_1000` to return value in RAO

cuteolaf:
The RPC call `delegateInfo.getDelegates` provides information for delegates.

e.g.
```json
{
  "delegate_ss58": "5G3jKS7BJPgrsirMu8cXGQiPRZYC5ceDx9xVyswhSEG8tGUs",
  "take": 11796,
  "nominators": [
    ["5ExqSxwZWrPMuur8RBUjbfLMaGFkq3dmLedUvNWwhExkCxxF", 5755727],
    ["5FsZgSiWPJmm7Aj2H2fYPwMSWkG3LvHoo93m8WQNCJN6VrYx", 71999]
  ],
  "owner_ss58": "5FsZgSiWPJmm7Aj2H2fYPwMSWkG3LvHoo93m8WQNCJN6VrYx",
  "registrations": [0, 11, 21],
  "validator_permits": [],
  "return_per_1000": 2,
  "total_daily_return": 18777
},
{
  "delegate_ss58": "5EcisWbQAnkLyHneKgSrHN75vj4t9AnXVM3YPqhSMkbpBLKE",
  "take": 11796,
  "nominators": [],
  "owner_ss58": "5EHMc8mKZTzdXgx8gq9RKJ1tmatbyJg7gaAEasxNUCWyfLSE",
  "registrations": [],
  "validator_permits": [],
  "return_per_1000": 0,
  "total_daily_return": 0
},
```

The problem is that `return_per_1000` is 0 for most of the delegates.
It's because the actual value is in the range of (0, 1).
Those delegates with `return_per_1000` > 1 are mostly weight copiers, who are not actually contributing to the network, but only trying to get rewarded.

To fix this problem, I'd like to suggest updating `return_per_1000` to show the value in RAO(`10^-9` TAO).

That being said, this formula can be updated.

**CURRENT**:
`return_per_1000 = emissions_per_day *  0.82 / (total_stake / 1000);`

**PROPOSED**:
`return_per_1000 = emissions_per_day * 0.82 / (total_stake / 1000) * 10 ^ 9`  -->
`return_per_1000 = emissions_per_day * 0.82 / total_stake * 10 ^ 12`

I've already tested this approach in our project and it works well. 😀

### DIFF
--- a/pallets/subtensor/src/delegate_info.rs
+++ b/pallets/subtensor/src/delegate_info.rs
@@ -61,7 +61,7 @@ impl<T: Config> Pallet<T> {
         let mut return_per_1000: U64F64 = U64F64::from_num(0);
         
         if total_stake > U64F64::from_num(0) {
-            return_per_1000 = ( emissions_per_day *  U64F64::from_num(0.82)) / (total_stake /  U64F64::from_num(1000));
+            return_per_1000 = emissions_per_day *  U64F64::from_num(820_000_000_000) / total_stake;
         }
         
         return DelegateInfo {


cuteolaf:
@Rubberbandits @S0AndS0 @unconst @camfairchild @shibshib

sam0x17 (changes requested):
@cuteolaf 

This is a great idea, but I am also concerned about rug-pulling anyone who might already be using this RPC call and might not know about the change, since changing by this order of magnitude would be.. quite a change if your code isn't expecting it.

I would prefer we make a new RPC call, maybe `return_per_1000_in_rao`, and then add a deprecation notice to the old call with the intention that we will eventually remove it

does that sound reasonable? cc @unconst 

I suppose we _could_ forgo this step if we are truly certain no one is able to make use of this RPC call as is because the result is currently useless across the board, but you never know who is using what at the end of the day.

also there is now a merge conflict ;)

cuteolaf:
this has been merged into the main branch already.

cuteolaf:
my bad, it's not actually merged yet.
apologies for the confusion.

sam0x17:
@cuteolaf did you have any thoughts on my comment?

PR #240, opened 2024-02-23
Automatically set 0 to deregistered subnets

shibshib:
This will ensure that when subnets are deregistered, its weights are automagically set to 0.

### DIFF
--- a/pallets/subtensor/src/root.rs
+++ b/pallets/subtensor/src/root.rs
@@ -823,7 +823,7 @@ impl<T: Config> Pallet<T> {
         let _ = Keys::<T>::clear_prefix(netuid, u32::max_value(), None);
         let _ = Bonds::<T>::clear_prefix(netuid, u32::max_value(), None);
       
-        // --- 3. Iterate over stored weights and fill the matrix.
+        // --- 9. Iterate over stored weights and fill the matrix.
         for (uid_i, weights_i) in
         <Weights<T> as IterableStorageDoubleMap<u16, u16, Vec<(u16, u16)>>>::iter_prefix(
             Self::get_root_netuid(),
@@ -844,7 +844,7 @@ impl<T: Config> Pallet<T> {
             Weights::<T>::insert(Self::get_root_netuid(), uid_i, modified_weights);
         }
 
-        // --- 9. Remove various network-related parameters.
+        // --- 10. Remove various network-related parameters.
         Rank::<T>::remove(netuid);
         Trust::<T>::remove(netuid);
         Active::<T>::remove(netuid);
@@ -857,7 +857,7 @@ impl<T: Config> Pallet<T> {
         ValidatorPermit::<T>::remove(netuid);
         ValidatorTrust::<T>::remove(netuid);
 
-        // --- 10. Erase network parameters.
+        // --- 11. Erase network parameters.
         Tempo::<T>::remove(netuid);
         Kappa::<T>::remove(netuid);
         Difficulty::<T>::remove(netuid);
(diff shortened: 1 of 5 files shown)


pallets/subtensor/src/root.rs
```diff
@@ -821,9 +822,29 @@ impl<T: Config> Pallet<T> {
...
+            let mut modified_weights = Vec::new();
+            // Iterate over each weight entry in `weights_i` to potentially update it.
+            for (subnet_id, weight) in weights_i.iter() {
+                if subnet_id == &netuid {
+                    // If the condition matches, modify the weight as needed and push it to the new vector.
+                    modified_weights.push((*subnet_id, 0)); // Set weight to 0 for the matching subnet_id.
+                } else {
+                    // For all other entries, just copy them to the new vector.
+                    modified_weights.push((*subnet_id, *weight));
+                }
```
sam0x17: This is inefficient for a few reasons:
1. If we use `Vec::new()` every time, but we actually know the final size of the Vec, we are relying on the auto-resizing mechanics of `Vec` when we could just make the whole allocation up-front, resulting in a few intermediate needless allocations
2. It turns out though, we are better off just cloning the original vec, since we are just taking the original `weights_i` vec and making some minor modifications. `clone()` for `Vec` uses the `memcpy` system call, which is much faster than manually copying via a for loop
3. Then in the actual for loop, we simply set the weight to zero for entries where the `subnet_id` matches `netuid`:

```suggestion
            let mut modified_weights = weights_i.clone();
            // Iterate over each weight entry to potentially update it.
            for (subnet_id, weight) in modified_weights.iter_mut() {
                if subnet_id == &netuid {
                    // If the condition matches, modify the weight
                    *weight = 0; // Set weight to 0 for the matching subnet_id.
                }
```

that said, I do wonder if this can be made even more efficient. If we could know ahead of time the index/indices where `subnet_id` is equal to `netuid`, then we could avoid looping over the whole vec, which would be much more efficient. What is the structure of `weights_i` going into this loop? I imagine we would only be stuck with this setup if `weights_i` has no pattern to it at all in terms of the `subnet_id`, is unsorted, and has an unknown number of duplicates. For example, if we know minimally that a given `subnet_id` only appears once, we can break early at least. If we know a `subnet_id` only appears once, and at a specific position, we can skip the loop entirely and index into that position, etc

sam0x17 (commented):
Looks good in general, one loop I think could be improved significantly

pallets/subtensor/src/root.rs
sam0x17: on the flipside, if it really is unsorted and with an unknown number of duplicates, the original code with `Vec::with_capacity(weights_i.len())` would be more efficient than what I have written or what was previously there ;)

shibshib:
Thanks @sam0x17 this was super helpful, updated the loop after testing.

sam0x17:
@shibshib did you have any thoughts regarding what I was saying about "is there anything we can do to avoid looping over the whole vec" -- I don't have enough context to know the topology of the `weights_i` vec but I'm curious what the details are here, like can an id be repeated, is the order random, etc

sam0x17 (approved):

shibshib:
> @shibshib did you have any thoughts regarding what I was saying about "is there anything we can do to avoid looping over the whole vec" -- I don't have enough context to know the topology of the `weights_i` vec but I'm curious what the details are here, like can an id be repeated, is the order random, etc

We unfortunately have to loop through the whole vec. The vector itself contains the weights set by validators for all subnets. If a subnet's weight must be reset to 0, it has to be reset by every validator as well. Hence, it's hard to optimize that further, I don't think. Open to suggestions though!

sam0x17:
should be good as-is then 👍🏻

unconst (approved):
LGTM

PR #241, opened 2024-02-29
Weights set stake limit

unconst:
Adds a stake min to setting weights.


shibshib:
Don't forget to update the spec_version

shibshib (approved):

PR #251, opened 2024-03-11
Allow the setting of specific subnet hyperparameters on chain.

shibshib:
This PR adds and fixes 4 subnet hyper parameter settings for subnet owners. Owners are now able to explicitly set Rho, Kappa, Difficulty, and bonds moving average on chain for their own subnet.

Users can see hyperparams for subnets using btcli. They can also sudo set hyperparams on subnets they own using btcli sudo set.


sam0x17 (approved):
looks good + makes sense!

Eugene-hu (approved):

unconst (approved):

sam0x17 (approved):

PR #254, opened 2024-03-12
Adds a migration to reset the Total Issuance value

unconst:
The general process is the following:

1. Fix the incorrect places where a burn is occuring in the code.
2. Add a migration which will re set the total issuance by summing up the balances, stake values and lock values.
3. Adds tests here for all three types of account balances.


pallets/subtensor/src/utils.rs
```diff
@@ -294,6 +294,9 @@ impl<T: Config> Pallet<T> {
     pub fn burn_tokens(amount: u64) {
         TotalIssuance::<T>::put(TotalIssuance::<T>::get().saturating_sub(amount));
     }
+    pub fn mint_tokens(amount: u64 ){
```
shibshib: This sort of naming might freak people out. Maybe let's just rename it to `update_totalissuance` or something

sam0x17 (approved):
LGTM 🎉

pallets/subtensor/src/block_step.rs
```diff
@@ -148,9 +148,9 @@ impl<T: Config> Pallet<T> {
                     &Self::get_subnet_owner(netuid),
                     Self::u64_to_balance(cut.to_num::<u64>()).unwrap(),
                 );
-                TotalIssuance::<T>::put(
-                    TotalIssuance::<T>::get().saturating_add(cut.to_num::<u64>()),
-                );
+
+                // We are minting tokens here for the owner.
+                Self::mint_tokens( cut.to_num::<u64>() );
```
shibshib: Same comment as below, 'minting' should be reworded to something less defi-sounding.

pallets/subtensor/src/registration.rs
```diff
@@ -401,9 +401,10 @@ impl<T: Config> Pallet<T> {
 
         // --- 5. Add Balance via faucet.
         let balance_to_add: u64 = 100_000_000_000;
+        Self::mint_tokens( 100_000_000_000 ); // Accounting here for the newly created tokens.
```
shibshib: Why is the total issuance being updated here?

unconst: because this is the faucet.

pallets/subtensor/src/block_step.rs
unconst: ok. legal.

pallets/subtensor/src/utils.rs
unconst: ok big brother

sam0x17: `update_total_issuance` imo

pallets/subtensor/src/migration.rs
```diff
@@ -22,6 +22,46 @@ pub mod deprecated_loaded_emission_format {
...
+        weight = weight.saturating_add(T::DbWeight::get().reads_writes(Stake::<T>::iter().count() as u64, 0));
+
+        // Calculate the sum of all stake values
+        let locked_sum: u64 = SubnetLocked::<T>::iter()
+            .fold(0, |accumulator, (_, locked_value)| accumulator.saturating_add(locked_value));
+        weight = weight.saturating_add(T::DbWeight::get().reads_writes(SubnetLocked::<T>::iter().count() as u64, 0));
+
+        // Retrieve the total balance sum
+        let total_balance = T::Currency::total_issuance();
+        let total_balance_sum: u64 = total_balance.try_into().unwrap_or_else(|_| panic!("Conversion must be within range"));
```
sam0x17, line 48: honestly this panic is really scary, could we just keep with the type of `T::Currency::total_issuance()` instead or does it not support the ops we need? (another reason to use my currencies crate in the future hahaha)

sam0x17 (changes requested):

sam0x17 (approved):

shibshib (approved):

PR #280, opened 2024-03-21
Bring back halving

JohnReedV:
This PR reimplements the block emission halving logic.


pallets/subtensor/src/block_step.rs
```diff
@@ -82,6 +84,43 @@ impl<T: Config> Pallet<T> {
...
 
+    // Determines if due for a halving at the current block and
+    // updates block emissions if so.
+    //
+    pub fn check_halving() {
+        let total_issuance = Self::get_total_issuance();
+        let current_emission = Self::get_block_emission();
+        let last_halving_issuance = Self::get_last_halving_issuance();
+
+        let is_halving: bool = match total_issuance {
```
distributedstatemachine: Since the halving occurs every 10_500_000 TAOs , do you think we can avoid the matches and do this instead?

```
pub fn check_halving() {
    let total_issuance = Self::get_total_issuance();
    let current_emission = Self::get_block_emission();
    let last_halving_issuance = Self::get_last_halving_issuance();

    // Assuming halving occurs every 10,500,000 tokens.
    const HALVING_INTERVAL: u64 = 10_500_000;
    // Calculate the number of halvings that should have occurred by now.
    let expected_halvings = total_issuance / HALVING_INTERVAL;
    // Calculate the number of halvings that had occurred by the last recorded issuance.
    let actual_halvings = last_halving_issuance / HALVING_INTERVAL;

    // Check if a new halving event should occur.
    if expected_halvings > actual_halvings {
        if current_emission == 0 {
            // Halvings are already complete, or no emission to halve.
            return;
        }
        // Halve the current emission and update the last halving issuance.
        let new_emission = current_emission / 2;
        Self::set_block_emission(new_emission);
        // Update the last halving issuance to the current total issuance.
        Self::set_last_halving_issuance(total_issuance);
        log::info!(
            "Halving event at total issuance {}: New block emission is {}",
            total_issuance,
            new_emission
        );
    }
}
```

JohnReedV: yeah this approach is better

pallets/subtensor/src/utils.rs
```diff
@@ -598,4 +599,40 @@ impl<T: Config> Pallet<T> {
...
+                result *= b;
+            }
+            b *= b;
+            exp /= 2;
+        }
+        result
+    }
+    pub fn floor(x: f64) -> f64 {
+        x as i64 as f64
+    }
```
sam0x17: so I should have thought about this before I helped you get f64 stuff working, but probably a bad idea to have anything on-chain rely on an f64 calculation

I would recommend instead using the `rust_decimal` crate with default features disabled, which disables the `std` feature allowing for `no_std` support which will let you use it in pallets etc. This way we will get predictable / deterministic outputs and eliminate floating point weirdness being a source of inconsistency

pallets/subtensor/src/block_step.rs
```diff
@@ -82,6 +84,38 @@ impl<T: Config> Pallet<T> {
...
+        const TOTAL_SUPPLY: f64 = 21_000_000_000_000_000.0; //Rao => 21_000_000 Tao
+        if total_issuance >= TOTAL_SUPPLY {
+            return 0.0;
+        }
+
+        let h = Self::log2(1.0 / (1.0 - total_issuance / (2.0 * 11_000_000_000_000_000.0)));
+        let h = Self::floor(h);
+        Self::powf(2.0, -h)
+    }
+
```
sam0x17: yeah really really really don't want f64 stuff on chain, sorry again I didn't flag that sooner

sam0x17 (changes requested):
so yeah, I think we should use `rust_decimal` in `no_std` mode over `f64` for safety / consistency reasons. On WASM it ironically might actually be faster anyway since it's backed by u64s

pallets/subtensor/src/block_step.rs
```diff
@@ -82,6 +84,38 @@ impl<T: Config> Pallet<T> {
         LoadedEmission::<T>::get(netuid).unwrap()
     }
 
+    // Checks the current block emission and
+    // updates the block emission if the recalculated value is lower
+    //
```
keithtensor, line 99:
```suggestion
    /// Checks the current block emission and
    /// updates the block emission if the recalculated value is lower
```

pallets/subtensor/src/block_step.rs
```diff
@@ -82,6 +84,38 @@ impl<T: Config> Pallet<T> {
...
+    pub fn check_halving() {
+        const ORIGINAL_EMISSION: f64 = 1_000_000_000.0; //Rao => 1 Tao
```
keithtensor: I'm surprised that this isn't already defined somewhere. In fact, it might not be a good idea to define it here, and instead should be a config item where the runtime gets to define what the original emission is. We'd then be able to reuse it here via `T::ORIGINAL_EMISSION`.

pallets/subtensor/src/block_step.rs
```diff
@@ -82,6 +84,38 @@ impl<T: Config> Pallet<T> {
...
+        let current_emission: u64 = Self::get_block_emission();
+
+        if current_emission > new_emission {
+            Self::set_block_emission(new_emission);
+        }
+    }
+
+    // Calculates the emission rate based on the total issuance,
+    // where the emission decreases as the total issuance approaches the supply cap.
+    //
```
keithtensor, line 112:
```suggestion
    /// Calculates the emission rate based on the total issuance,
    /// where the emission decreases as the total issuance approaches the supply cap.
```

pallets/subtensor/src/block_step.rs
```diff
@@ -82,6 +84,38 @@ impl<T: Config> Pallet<T> {
...
+    pub fn get_emission_from_issuance(total_issuance: f64) -> f64 {
+        const TOTAL_SUPPLY: f64 = 21_000_000_000_000_000.0; //Rao => 21_000_000 Tao
```
keithtensor: Same comment as `ORIGINAL_EMISSION` -- define this as a config item, and have the runtime supply this number.

keithtensor:
Make sure to address @sam0x17's point about using `rust_decimal` rather than `f64`.

pallets/subtensor/src/block_step.rs
shibshib: It is already defined in https://github.com/opentensor/subtensor/blob/09e0270028d3cb616aa781df80eae3467f8e6f9b/pallets/subtensor/src/lib.rs#L211

You should take advantage of that.

sam0x17:
> Make sure to address @sam0x17's point about using `rust_decimal` rather than `f64`.

they're gonna go with an i64-based decimal implementation they already use elsewhere apparently, but ye

distributedstatemachine:
Closing in favour of https://github.com/opentensor/subtensor/pull/290

PR #286, opened 2024-03-26
Add proxy pallet

camfairchild:
Integrates the proxy pallet

[Needs testing]

# TODO
- Add tests for each `ProxyType` filter
    - Can't do transactions not in filter
- Verify works on UI
- Verify fee and deposit


vstam1:
What is the progress on this? How soon can we get the proxy pallet merged?

vstam1:
@camfairchild Would love to see this feature merged, as it would allow us to work with cold wallets that can stake/unstake funds using a hot wallet. As it is a standard polkadot-sdk pallet that is included in many chains, it should be relatively easy to merge, right? 🙏

open-junius:
The PR is ready for review. I think the proxy types defined is flexible enough for token related, sub tensor staking and other operations. like to know more comments what's kind of type is needed from business logic

runtime/src/lib.rs
```diff
@@ -473,9 +481,9 @@ impl pallet_sudo::Config for Runtime {
 
 parameter_types! {
     // One storage item; key size is 32; value is size 4+4+16+32 bytes = 56 bytes.
-    pub const DepositBase: Balance = (1) as Balance * 2_000 * 10_000 + (88 as Balance) * 100 * 10_000;
+    pub const DepositBase: Balance = deposit(1, 88);
```
distributedstatemachine: Where does this figure come from ? The calculation doesnt seem to match the comments

runtime/src/lib.rs
```diff
@@ -489,6 +497,121 @@ impl pallet_multisig::Config for Runtime {
...
+} // allow all Calls; required to be most permissive
+impl InstanceFilter<RuntimeCall> for ProxyType {
+    fn filter(&self, c: &RuntimeCall) -> bool {
+        match self {
+            ProxyType::Any => true,
+            ProxyType::NonTransfer => !matches!(c, RuntimeCall::Balances(..)),
+            ProxyType::NonFungibile => !matches!(
+                c,
+                RuntimeCall::Balances(..)
+                    | RuntimeCall::SubtensorModule(pallet_subtensor::Call::add_stake { .. })
```
distributedstatemachine, line 555: Note to self: Rebase stao off development so we can capture this

distributedstatemachine (changes requested):
Left some comments . Is there a way to add tests to this ?

runtime/src/lib.rs
open-junius: From multisig pallet, I get the code as follows. the comments from substrate upstream. I will check if we should change it.
/// This is held for an additional storage item whose value size is
/// `4 + sizeof((BlockNumber, Balance, AccountId))` bytes and whose key size is
/// `32 + sizeof(AccountId)` bytes.
#[pallet::constant]
type DepositBase: Get<BalanceOf<Self>>;

runtime/src/lib.rs
open-junius: After rebase, the NonFungibile excludes all extrinsic related to balance in Subtensor pallet.

distributedstatemachine: ha.. this was a note for me !

open-junius:
> Left some comments . Is there a way to add tests to this ?

It is tricky to test the runtime, we can skip it for now since the pallet from upstream

runtime/src/lib.rs
open-junius: pls add stao extrinsic in filter

distributedstatemachine (approved):

JuanCortez93:
Hello Guys! Thank you very much for taking care of this. I´m glad to know you are testing this proxy pallet :) 

Do you have an estimation of the time it can take to push this to production?

PR #287, opened 2024-03-26
Add rate limit on staking transactions

JohnReedV:
This PR adds a rate limit to staking & unstaking transactions.


pallets/subtensor/src/lib.rs
```diff
@@ -1817,14 +1831,32 @@ where
...
-            }),
-            Some(Call::remove_stake { .. }) => Ok(ValidTransaction {
-                priority: Self::get_priority_vanilla(),
-                ..Default::default()
-            }),
+            Some(Call::add_stake { hotkey, .. }) => {
+                let stakes_this_interval = Pallet::<T>::get_stakes_this_interval_for_hotkey(hotkey);
+                let max_stakes_per_interval = Pallet::<T>::get_target_stakes_per_interval();
+
+                if stakes_this_interval >= max_stakes_per_interval {
```
cuteolaf, line 1838:
```suggestion
                if stakes_this_interval > max_stakes_per_interval {
```

pallets/subtensor/src/lib.rs
```diff
@@ -1817,14 +1831,32 @@ where
...
+                }
+
+                Ok(ValidTransaction {
+                    priority: Self::get_priority_vanilla(),
+                    ..Default::default()
+                })
+            }
+            Some(Call::remove_stake { hotkey, .. }) => {
+                let stakes_this_interval = Pallet::<T>::get_stakes_this_interval_for_hotkey(hotkey);
+                let max_stakes_per_interval = Pallet::<T>::get_target_stakes_per_interval();
```
cuteolaf, line 1849: Is this correct?
I don't think this limits the amount of unstaking in an interval.

pallets/subtensor/src/lib.rs
```diff
@@ -1817,14 +1831,32 @@ where
...
+
+                if stakes_this_interval >= max_stakes_per_interval {
```
cuteolaf, line 1851:
```suggestion
                if stakes_this_interval > max_stakes_per_interval {
```

pallets/subtensor/src/lib.rs
keithtensor, line 1866: When is this `validate` function called? We're essentially doing 2 storage lookups whenever the call is `add_stake` or `remove_stake`, not sure if we're OK with that...

pallets/subtensor/src/utils.rs
```diff
@@ -140,6 +140,16 @@ impl<T: Config> Pallet<T> {
...
     }
+    pub fn set_target_stakes_per_interval(target_stakes_per_interval: u64) {
+        TargetStakesPerInterval::<T>::set(target_stakes_per_interval)
+    }
+    pub fn set_stakes_this_interval_for_hotkey(hotkey: &T::AccountId, stakes_this_interval: u64) {
+        TotalHotkeyStakesThisInterval::<T>::insert(hotkey, stakes_this_interval);
+    }
+    pub fn reset_stakes_and_unstakes_this_interval() {
+        // This removes all key-value pairs from the storage map
+        let _ = TotalHotkeyStakesThisInterval::<T>::clear(u32::MAX, None);
```
keithtensor: Need to be careful about this one, how many hotkey stakes do we expect to have in each interval? If it's a lot, it may consume a lot of compute from the chain.

pallets/subtensor/src/lib.rs
JohnReedV: remove/add stake can only be called once per hotkey per interval at the moment.

pallets/subtensor/src/utils.rs
JohnReedV: Each hotkey can only stake or unstake once per interval, defined as a duration of 360 blocks or equivalently, a 72-minute period, given a block time of 12 seconds.

My assumption writing this is that there won't be "too many" stakes in any given 72 minuet window especially considering the financial cost to the staker per stake.

sam0x17 (approved):
looks good to me, though we do have to be very careful about these state changes in `validate`, but I think for the purpose of preventing spam it is necessary

JohnReedV:
The reason for changing the state during validation is to prevent spamming of the pool.

If `set_stakes_this_interval_for_hotkey()` were called at execution or in `pre_dispatch()`, a staker could potentially submit multiple staking transactions to the pool before the first transaction is executed and the state updated.

Though the additional checks at execution would prevent these subsequent staking transactions from completing, the primary aim of the PR is to prevent such spamming.

Edit: I was wrong. The small time window paired with the financial cost to the staker should be sufficient in preventing pool spam before the stake is executed.

sam0x17 (approved):

distributedstatemachine (approved):

sam0x17 (approved):

sam0x17 (approved):

PR #290, opened 2024-03-28
Emission schedule

unconst:


pallets/subtensor/tests/root.rs
```diff
@@ -697,3 +697,95 @@ fn test_weights_after_network_pruning() {
...
+            (20_999_920_000_000_000, 3_814),     // Eighteenth halving event
+            (20_999_960_000_000_000, 1_907),     // Nineteenth halving event
+            (20_999_980_000_000_000, 953),       // Twentieth halving event
+            (20_999_990_000_000_000, 476),       // Twenty-first halving event
+            (20_999_990_500_000_000, 476),
+            (20_999_995_000_000_000, 238), // Twenty-second halving event
+            (20_999_998_000_000_000, 119), // Twenty-third halving event
+            (20_999_999_000_000_000, 59),  // Twenty-fourth halving event
+            (21_000_000_000_000_000, 0),   // Total supply reached, emissions stop
+            (21_100_000_000_000_000, 0)    // Just for fun
```
sam0x17: nit: get these comments all vertically aligned

pallets/subtensor/tests/root.rs
```diff
@@ -697,3 +697,95 @@ fn test_weights_after_network_pruning() {
...
+            };
+            assert_eq!(
+                SubtensorModule::get_block_emission(),
+                expected_emission,
+                "Issuance: {}",
+                issuance_f64
+            );
+            step = expected_emission as usize;
+        }
+    });
```
sam0x17: I guess as long as this consistently passes this is fine but I suppose if there is ever a hardware difference w.r.t floating points vs the safe implementation this test _might_ fail on some random hardware. Probably fine tho

my preference tho would be that this uses safe math as well if possible

sam0x17 (approved):
lgtm, few nits

pallets/subtensor/src/utils.rs
```diff
@@ -599,3 +597,37 @@ impl<T: Config> Pallet<T> {
...
+    let mut b = base;
+    while exp > 0 {
+        if exp % 2 == 1 {
+            result *= b;
+        }
+        b *= b;
+        exp /= 2;
+    }
+    result
+}
```
sam0x17: as const mentioned, first one is built into `I96F32` apparently, and second one you can just self-multiply in a loop since we don't have fractional exponent. In other words don't need `powf` really just need `pow` which is a lot simpler to implement

sam0x17 (changes requested):

pallets/subtensor/tests/root.rs
JohnReedV: aligning these makes cargo fmt fail :(

sam0x17 (approved):
looks good, pending answer from @unconst about that defensive guard for the scenario where supply exceeds 21 million

PR #302, opened 2024-04-08
connect grandpa rpc

open-junius:
The pr is from MacieRaj, rebased with development branch.


open-junius:
The PR looks good to me, the grandpa rpc is supported.

node/src/service.rs
```diff
@@ -222,26 +222,44 @@ pub fn new_full(config: Configuration) -> Result<TaskManager, ServiceError> {
...
+						shared_authority_set: shared_authority_set.clone(),
+						justification_stream: justification_stream.clone(),
+						subscription_executor: subscription_executor.clone(),
+						finality_provider: finality_proof_provider.clone(),
+					},
+					backend: rpc_backend.clone(),
+				};
+			crate::rpc::create_full(deps).map_err(Into::into)
+		})
+	};
```
sam0x17, line 262: lot of boxing and cloning here, I wonder if it is actually all necessary?

sam0x17 (approved):

node/src/service.rs
open-junius: I just checked, all clones are needed. can't fully understand the grammar yet, like need a new variable then box can pin them

PR #312, opened 2024-04-10
add zero check when stake

open-junius:
When checking the status of network, found out lots of stake with 0 records. We can avoid it via checking the stake amount


sam0x17 (approved):
💯

open-junius:
close the PR, as agreed with @gztensor . the code change will be part of other PR.

PR #317, opened 2024-04-11
set root weights with coldkey

JohnReedV:
This PR changes the requirement from signing root weights with a hot key to using a cold key instead.

### DIFF
--- a/pallets/subtensor/src/staking.rs
+++ b/pallets/subtensor/src/staking.rs
@@ -181,7 +181,8 @@ impl<T: Config> Pallet<T> {
         );
 
         // --- 8. Ensure the remove operation from the coldkey is a success.
-        let actual_amount_to_stake = Self::remove_balance_from_coldkey_account(&coldkey, stake_as_balance.unwrap())?;
+        let actual_amount_to_stake =
+            Self::remove_balance_from_coldkey_account(&coldkey, stake_as_balance.unwrap())?;
 
         // --- 9. If we reach here, add the balance to the hotkey.
         Self::increase_stake_on_coldkey_hotkey_account(&coldkey, &hotkey, actual_amount_to_stake);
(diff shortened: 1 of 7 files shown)


pallets/subtensor/src/staking.rs
```diff
@@ -190,11 +191,7 @@ impl<T: Config> Pallet<T> {
         Self::set_last_tx_block(&coldkey, block);
 
         // --- 10. Emit the staking event.
-        Self::set_stakes_this_interval_for_hotkey(
-            &hotkey,
-            stakes_this_interval + 1,
-            block,
-        );
+        Self::set_stakes_this_interval_for_hotkey(&hotkey, stakes_this_interval + 1, block);
```
distributedstatemachine, line 194: nit: Is your editor automatically formatting ? If you are using vscode , it migght make sense for you to turn the formatOnSave off in your settings.json

distributedstatemachine (commented):
LGTM!

pallets/subtensor/src/staking.rs
sam0x17: we definitely want format on save, pretty much everything it does will be CI enforced soon if it isn't already

sam0x17:
> LGTM!

@distributedstatemachine can you add an approval, you just commented ;)

sam0x17 (approved):
just a note that we may want to think eventually consider the implications of cold key compromise or loss. It might be beneficial to have a recovery mechanism or a way to rotate keys securely without disrupting the network operations or user experience.

distributedstatemachine (approved):

pallets/subtensor/src/staking.rs
distributedstatemachine: In the past , formatOnSave has really screwed me over , especially when your settings are different from the repo , you loads of changes in PRs which are due to differences in formatting settings and lets not even mention rebases. 

I think it might better to have a make / just file and people run make/ just lint before every commit , so that its standardised

distributedstatemachine: https://discord.com/channels/799672011265015819/1176889593136693339/1229443862909423697

PR #322, opened 2024-04-12
Fixes bootnode ID for docker containers

shibshib:
Currently new docker test nodes will fail because they are trying to connect to the test boot node but using the wrong node key, this PR fixes that.


prop-opentensor (approved):

sam0x17 (approved):

birkskyum:
> Currently new docker test nodes

Just to clarify, anyone using the script for test nodes, be it `-e binary` or  `-e docker` will be affected.

birkskyum:
@shibshib , it appear the peerID in here also has to be updated accordingly:
https://docs.bittensor.com/subtensor-nodes/using-source#lite-node-on-testchain

birkskyum:
This is affecting people compiling from source as well, thought he `-e binary`

PR #325, opened 2024-04-13
Align chain flag in test nodes

birkskyum:
Of these places with test node configuration:

- **subtensor/docker-compose** (raw_testspec)
- **subtensor/scripts/run/subtensor.sh** (raw_spec)
- **subtensor/docs/running-subtensor-locally.md** (raw_spec)
- **developer-docs/docs/subtensor-nodes/using-source.md** (raw_spec)

Only the docker-compose.yml would set chain to raw_testspec.json instead of raw_spec.json. This aligns the docker-compose to the rest. Let me know if it's actually the 3 other places that should be aligned to the docker-compose.yml instead.

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -1,4 +1,4 @@
-version: '3.8'
+version: "3.8"
 
 volumes:
   mainnet-lite-volume:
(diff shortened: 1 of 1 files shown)


distributedstatemachine:
Thanks for opening this issue! 

With regards to changes in this PR , I dont think we need them , as the changes you introduce are changes the testnet docket compose builds to use the mainnet spec, which is not what we want.

I will defer to @rajkaramchedu to comment on the other docs

birkskyum:
Interesting, I have had a lot of issues of hitting the main unintended lately, so if the raw_testspec is up-to-date and working as intended it would make sense if it's all the other places that needs altering (see #329).

PR #327, opened 2024-04-15
chore: replace localnet_setup with rust installation

distributedstatemachine:


scripts/install_rust.sh
```diff
@@ -0,0 +1,42 @@
...
+fi
+
+curl https://sh.rustup.rs -sSf | sh -s -- -y
+source "$HOME/.cargo/env"
+rustup default stable
+
+rustup update nightly
+rustup target add wasm32-unknown-unknown --toolchain nightly
+
+echo "*** Rust installation complete"
```
sam0x17:
```suggestion
echo "*** Rust installation complete"

```

scripts/install_rust.sh
```diff
@@ -0,0 +1,42 @@
...
+    brew update
+    brew install openssl cmake llvm
+    elif [[ "$(uname)" == "Linux" ]]; then
+    if [[ -f "/etc/arch-release" ]]; then
+        # Arch Linux
+        sudo pacman -Syu --noconfirm
+        sudo pacman -S --noconfirm cmake pkgconf openssl git gcc clang
+    else
+        # Ubuntu (and other Debian-based distributions)
+        sudo apt update
```
sam0x17: note that usage of plain `apt` in scripts is unstable, still supposed to use `apt-get` because it has a stable CLI that won't i.e. spawn some UI sometimes
```suggestion
        sudo apt-get update
```

scripts/install_rust.sh
```diff
@@ -0,0 +1,42 @@
...
+        sudo apt install -y cmake pkg-config libssl-dev git gcc build-essential clang libclang-dev
```
sam0x17:
```suggestion
        sudo apt-get install -y cmake pkg-config libssl-dev git gcc build-essential clang libclang-dev
```

scripts/install_rust.sh
```diff
@@ -0,0 +1,42 @@
+#!/bin/bash
```
sam0x17: would prefer like this so it is POSIX compliant and will work with things like zsh so we don't require bash

```suggestion
#!/bin/sh
```

sam0x17 (approved):
few nits, looks good if addressed

PR #333, opened 2024-04-16
fix: feature propagation

orriin:
Propagates missing features with `zepter`, and adds a CI check to prevent further breakages.

This fixes `try-runtime` and `std`  builds.

### DIFF
--- a/.github/workflows/check-rust.yml
+++ b/.github/workflows/check-rust.yml
@@ -4,7 +4,6 @@ concurrency:
   group: ci-${{ github.ref }}
   cancel-in-progress: true
 
-
 on:
   ## Run automatically for all PRs against main, regardless of what the changes are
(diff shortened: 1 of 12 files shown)


.github/workflows/check-rust.yml
```diff
@@ -331,3 +330,25 @@ jobs:
...
+        uses: actions-rs/toolchain@v1
+        with:
+          profile: minimal
+          toolchain: stable
+
+      - name: Install Zepter
+        run: cargo install --locked -q zepter && zepter --version
+
+      - name: Checkout
+        uses: actions/checkout@v3
```
sam0x17, line 349:
```suggestion
        uses: actions/checkout@v4
```

sam0x17 (approved):
💯

PR #338, opened 2024-04-17
fix: account migration

orriin:
Fixes minor issues with the last account migration and vastly improves test coverage / sanity checks.

## Issues fixed

1. I was decoding `Account` straight into 

```rust
        pub struct AccountDataStruct<Balance> {
            pub free: Balance,
            pub reserved: Balance,
            pub misc_frozen: Balance,
            pub fee_frozen: Balance,
        }
```
instead of
```rust
        pub struct AccountStruct<Balance> {
            pub nonce: u32,
            pub consumers: u32,
            pub providers: u32,
            pub sufficients: u32,
            pub data: AccountDataStruct<Balance>,
        }
```
which caused `free` to be set to `nonce`, `reserved` to be set to `consumers`, etc.

2. I was updating storage using `::mutate` while the state was still undecodable, leading to the new state being set in storage with everything except `data` reset to defaults (zero). Switching to `::insert` and passing the entire new state resolves this issue.

## Other changes

- Moved migration to dedicated file, now that it's getting large
- Greatly improved thoroughness of `pre_upgrade` and `post_upgrade` checks


shibshib (approved):

runtime/src/migrations.rs
```diff
@@ -0,0 +1,162 @@
...
+            frame_system::pallet::Pallet<T>,
+            Blake2_128Concat,
+            AccountId,
+            AccountStruct<Balance>,
+            ValueQuery,
+        >;
+    }
+
+    const TARGET: &'static str = "runtime::account_data_migration";
+    pub struct Migration;
```
keithtensor: Might wanna give this a more descriptive name.

orriin: I mean it is the only Migration in the `account_data_migration` module :P

runtime/src/lib.rs
```diff
@@ -1147,121 +1150,6 @@ pub type SignedExtra = (
...
-            let account_ids = prev::Account::<Runtime>::iter_keys().collect::<Vec<_>>();
-            log::info!(target: TARGET, "pre-upgrade");
-
-            Ok(account_ids.encode())
-        }
-
-        /// Ensures post-upgrade that
-        /// 1. Number of accounts has not changed.
-        /// 2. Each account exists in storage and decodes.
-        /// 3. Each account has a provider val >0.
```
keithtensor, line 1192: Was this something you wrote? I was thinking how it was even possible to get an old lifecycle runtime upgrade hook from upstream to execute in our chain, but if you wrote a brand new one, it does the job just fine.

runtime/src/migrations.rs
```diff
@@ -0,0 +1,162 @@
...
+                frame_system::pallet::Account::<Runtime>::insert(acc_id.clone(), new_account);
+
+                // Ensure upgraded
+                pallet_balances::Pallet::<Runtime, ()>::ensure_upgraded(&acc_id);
+            }
+
+            log::info!(target: TARGET, "Migrated {} accounts ✅", accounts.len());
+
+            // R/W not important for solo chain.
+            return <Runtime as frame_system::Config>::DbWeight::get().reads_writes(0u64, 0u64);
```
keithtensor: You're better than this, Mr. Rust expert:

```suggestion
            <Runtime as frame_system::Config>::DbWeight::get().reads_writes(0u64, 0u64)
```

runtime/src/lib.rs
orriin: Yeah I wrote this from scratch.

runtime/src/migrations.rs
orriin: Huh I'm surprised clippy didn't shout at me about that

runtime/src/migrations.rs
```diff
@@ -0,0 +1,162 @@
+pub(crate) mod account_data_migration {
```
keithtensor: Oh, I didn't know you did this weird thing. Do you mind pulling this mod definition up to `lib.rs` and rename this file?

keithtensor (approved):

sam0x17 (approved):

PR #341, opened 2024-04-18
fix: update `scripts/localhost.sh`

orriin:
`--ws-port` was removed in v0.9.43 (https://github.com/paritytech/polkadot/releases/tag/v0.9.43)


orriin:
Oh, I see already done in https://github.com/opentensor/subtensor/pull/339

PR #343, opened 2024-04-19
fix: try-runtime

orriin:
- Aligns on-chain and in-code storage versions
- Adds idempotency check for `account_data_migration` 
- Adds `try-runtime` CI to prevent future breakage


sam0x17 (approved):
💯

PR #345, opened 2024-04-20
Min nominator stake

unconst:
The change does the following:

1. Introduces a function int the block_step which clears all nomination staking account balances (returning the amounts to the coldkey balance) if they are below a min value.

2. The operation runs every 1000 blocks. 

3. The code includes a test of this operation

4. The code includes a sudo operation to set the min and a test for the operation.

5. The code introduces a new Storage item NominatorMinRequiredStake


pallets/subtensor/src/block_step.rs
```diff
@@ -24,10 +24,39 @@ impl<T: Config> Pallet<T> {
...
+        }
         // Return ok.
         Ok(())
     }
 
+    /// Clears small nominations that are below the minimum required stake.
+    /// This function is designed to run at specific intervals, every 1000 blocks, to ensure that
+    /// the network does not retain nominator accounts with stakes that are too low. This helps in
+    /// reduce the computational complexity of running epochs.
+    ///
```
orriin, line 40: In what scenarios can the stake go beneath the minimum stake? If only when `NominatorMinRequiredStake` increases or a nominators individual stake decreases, can we check the min stake at those moments rather than on a fixed interval?

It should be O(1) cost to check a nominators individual stake when it decreases, so that's fine. It's O(n) to check everyone when `NominatorMinRequiredStake` increases, but that should hopefully happen very infrequently and should be much preferable compared to running on a fixed interval I think?   

If this is high priority to merge because it resolves a more severe performance issue, perhaps we could merge this as-is but create an issue to remove the fixed interval O(n) loop in a follow up PR?

```suggestion
    /// Clears small nominations that are below the minimum required stake.
    /// This function is designed to run at specific intervals, every 1000 blocks, to ensure that
    /// the network does not retain nominator accounts with stakes that are too low. This helps in
    /// reduce the computational complexity of running epochs.
```

pallets/subtensor/src/block_step.rs
```diff
@@ -24,10 +24,39 @@ impl<T: Config> Pallet<T> {
...
+    pub fn clear_small_nominations() {
+        // Loop through all staking accounts to identify and clear nominations below the minimum stake.
+        for (hotkey, coldkey, stake) in Stake::<T>::iter() {
+            // Verify if the account is a nominator account by checking ownership of the hotkey by the coldkey.
+            if !Self::coldkey_owns_hotkey(&coldkey, &hotkey) {
+                // If the stake is below the minimum required, it's considered a small nomination and needs to be cleared.
+                if stake < Self::get_nominator_min_required_stake() {
+                    // Remove the stake from the nominator account. (this is a more forceful unstake operation which )
+                    // Actually deletes the staking account.
```
orriin, line 49:
```suggestion
                    // Remove the stake from the nominator account (this is a more forceful unstake operation which
                    // actually deletes the staking account).
```

pallets/subtensor/src/lib.rs
```diff
@@ -369,6 +369,10 @@ pub mod pallet {
         0
     }
     #[pallet::type_value]
+    pub fn DefaultNominatorMinRequiredStake<T: Config>() -> u64 {
+        0
```
orriin, line 373: Why is 0 the default? My understanding is that it should be some significant amount, like 10TAO.

orriin (commented):
- I have a question about the frequency of checking the staked amount of every staker, and whether that is necessary 
- Need to run `cargo +nightly-2024-03-05 fmt`
- Tests do not compile

pallets/subtensor/src/block_step.rs
unconst: Yes better would be a check in stake and unstake ops O(1) + running the O(n) when we change the value.

orriin:
Superseded by https://github.com/opentensor/subtensor/pull/361

PR #356, opened 2024-04-22
Update test for per-subnet take emission

gztensor:
## Description
Update the test_different_subnet_take_different_distribution test so that it accounts for dynamic emission

## Related Issue(s)

n/a

## Type of Change
- [ ] Documentation update
- [x] Other (please describe): Tests

## Breaking Change

n/a

## Checklist


pallets/subtensor/tests/staking.rs
```diff
@@ -3205,6 +3205,14 @@ fn test_can_set_different_take_per_subnet() {
     });
 }
 
+fn assert_substake(coldkey: &U256, hotkey: &U256, netuid: u16, amount: u64) {
```
distributedstatemachine: Do you think it might be better if we moved these to the end of the file ? Or better yet , create a test utils file where we can add all the helper functions

pallets/subtensor/tests/staking.rs
```diff
@@ -3222,14 +3230,37 @@ fn test_different_subnet_take_different_distribution() {
...
-        SubtensorModule::add_balance_to_coldkey_account(&coldkey1, 100000);
+        SubtensorModule::add_balance_to_coldkey_account(&coldkey0, 100000000000000u64);
+        SubtensorModule::add_balance_to_coldkey_account(&coldkey1, 100000000000000u64);
 
         // Register the 2 neurons to new networks.
-        add_network(netuid1, 0, 0);
-        add_network(netuid2, 0, 0);
-        register_ok_neuron(netuid1, hotkey0, coldkey0, 124124);
-        register_ok_neuron(netuid2, hotkey0, coldkey0, 124124);
+        let lock_cost_1 = SubtensorModule::get_network_lock_cost();
```
distributedstatemachine: I think these should be tested separated. Is it possible to isolate subnet takes v dtao ?

pallets/subtensor/tests/staking.rs
distributedstatemachine: Over better still , we should move it to mocks

gztensor: I think helper macros are the best approach. If a function fails, we only see the failed assertion in this function, which may be called in 100s of lines in the test code. But if a assertion macro fails, we get pointed to a particular test.

pallets/subtensor/tests/staking.rs
gztensor: Yes, and it is very needed. I'll have a note about it in the test plan.

distributedstatemachine (approved):

PR #357, opened 2024-04-22
Increase base transaction fee to 0.00012 TAO

unconst:
## Description
Increases the base transaction fee from 350 RAO per transaction to 0.00012 TAO per transaction.

### DIFF
--- a/runtime/src/lib.rs
+++ b/runtime/src/lib.rs
@@ -268,7 +268,7 @@ impl pallet_utility::Config for Runtime {
 }
 
 // Existential deposit.
-pub const EXISTENTIAL_DEPOSIT: u64 = 2_000_000;
+pub const EXISTENTIAL_DEPOSIT: u64 = 500;
 
 impl pallet_balances::Config for Runtime {
     type MaxLocks = ConstU32<50>;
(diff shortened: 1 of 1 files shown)


sam0x17:
rebased to be off of `main`

sam0x17 (approved):
pending any changes regarding ED or last minute tweaks to the tx fee

keithtensor (approved):

distributedstatemachine (approved):

shibshib (approved):

runtime/src/lib.rs
```diff
@@ -268,7 +268,7 @@ impl pallet_utility::Config for Runtime {
 }
 
 // Existential deposit.
-pub const EXISTENTIAL_DEPOSIT: u64 = 500;
+pub const EXISTENTIAL_DEPOSIT: u64 = 2_000_000;
```
shibshib: Do we know the implications of this change? If ED is increased, would this affect wallets that have staked tao in their hotkey but 0 tao in their coldkey? It would be problematic for users if all of their tao is staked and then suddenly their cold wallet is destroyed, orphaning their hot wallet. 

I suppose it's always possible to regain the wallet if you have the mnemonic, but it can just be annoying.

shibshib (changes requested):

shibshib (approved):

PR #348, opened 2024-04-21
Feat/nominator stake changes

distributedstatemachine:
## Description

This supercedes https://github.com/opentensor/subtensor/pull/345 , and has the development branch changes. We might also want to consider add stao? dynamic changes to this
This PR : 

- Enforces a minimum threshold on stake/ unstake operations for nominators
- Clears nominations below the threshold whenever it is set. 

## TODO

- [x] Fix broken tests 
- [x] Ensure ED changes are carried along

## Related Issue(s)

- Closes https://github.com/opentensor/subtensor/issues/347 (partially)

## Type of Change

- [ ] Bug fix (non-breaking change which fixes an issue)
- [x] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] Documentation update
- [ ] Other (please describe):

## Breaking Change

If this PR introduces a breaking change, please provide a detailed description of the impact and the migration path for existing applications.

## Checklist

- [x] I have performed a self-review of my own code
- [x] I have commented my code, particularly in hard-to-understand areas
- [x] I have run `cargo fmt` and `cargo clippy` to ensure my code is formatted and linted correctly
- [ ] I have made corresponding changes to the documentation
- [x] My changes generate no new warnings
- [x] I have added tests that prove my fix is effective or that my feature works
- [x] New and existing unit tests pass locally with my changes
- [ ] Any dependent changes have been merged and published in downstream modules

PR #361, opened 2024-04-23
hotfix: minimum nominator stake

orriin:
Supersedes https://github.com/opentensor/subtensor/pull/345 and https://github.com/opentensor/subtensor/pull/348.

- Creates new `admin-util` call `sudo_set_nominator_min_required_stake`, which clears small nomination stakes
- Updates `add_stake` and `remove_stake` calls to enforce the new nominator minimum stake

### DIFF
--- a/pallets/admin-utils/src/lib.rs
+++ b/pallets/admin-utils/src/lib.rs
@@ -53,7 +53,6 @@ pub mod pallet {
     }
 
     #[pallet::event]
-    #[pallet::generate_deposit(pub(super) fn deposit_event)]
     pub enum Event<T: Config> {}
 
     // Errors inform users that something went wrong.
@@ -765,6 +764,27 @@ pub mod pallet {
             T::Subtensor::set_weights_min_stake(min_stake);
             Ok(())
         }
+
+        /// Sets the minimum stake required for nominators, and clears small nominations
+        /// that are below the minimum required stake.
+        #[pallet::call_index(43)]
+        #[pallet::weight((0, DispatchClass::Operational, Pays::No))]
+        pub fn sudo_set_nominator_min_required_stake(
+            origin: OriginFor<T>,
+            // The minimum stake required for nominators.
+            min_stake: u64,
+        ) -> DispatchResult {
+            ensure_root(origin)?;
+            let prev_min_stake = T::Subtensor::get_nominator_min_required_stake();
+            log::trace!("Setting minimum stake to: {}", min_stake);
+            T::Subtensor::set_nominator_min_required_stake(min_stake);
+            if min_stake > prev_min_stake {
+                log::trace!("Clearing small nominations");
+                T::Subtensor::clear_small_nominations();
+                log::trace!("Small nominations cleared");
+            }
+            Ok(())
+        }
     }
 }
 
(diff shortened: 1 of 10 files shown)


pallets/admin-utils/src/lib.rs
```diff
@@ -765,6 +764,27 @@ pub mod pallet {
...
+        pub fn sudo_set_nominator_min_required_stake(
+            origin: OriginFor<T>,
+            // The minimum stake required for nominators.
+            min_stake: u64,
+        ) -> DispatchResult {
+            ensure_root(origin)?;
+            let prev_min_stake = T::Subtensor::get_nominator_min_required_stake();
+            log::trace!("Setting minimum stake to: {}", min_stake);
+            T::Subtensor::set_nominator_min_required_stake(min_stake);
+            if min_stake > prev_min_stake {
```
shibshib, line 781: Do we always want to increase the min required stake? What if we messed something up and want to reduce it?

pallets/subtensor/src/migration.rs
```diff
@@ -22,30 +23,37 @@ pub mod deprecated_loaded_emission_format {
...
+        weight = weight.saturating_add(
+            T::DbWeight::get().reads_writes(SubnetLocked::<T>::iter().count() as u64, 0),
+        );
 
         // Retrieve the total balance sum
         let total_balance = T::Currency::total_issuance();
-        let total_balance_sum: u64 = total_balance.try_into().unwrap_or_else(|_| panic!("Conversion must be within range"));
+        let total_balance_sum: u64 = total_balance
+            .try_into()
+            .unwrap_or_else(|_| panic!("Conversion must be within range"));
```
shibshib: Do we want a panic during a migration? We should avoid these as much as possible in the critical path.

unconst (approved):

pallets/subtensor/src/staking.rs
```diff
@@ -158,12 +158,25 @@ impl<T: Config> Pallet<T> {
...
 
+        // --- 7. If this is a nomination stake, check if total stake after adding will be above
+        // the minimum required stake.
+        if !Self::coldkey_owns_hotkey(&coldkey, &hotkey) {
+            let total_stake_after_add =
+                Stake::<T>::get(&hotkey, &coldkey).saturating_add(stake_to_be_added);
+
+            ensure!(
+                total_stake_after_add >= NominatorMinRequiredStake::<T>::get(),
+                Error::<T>::NomStakeBelowMinimumThreshold
```
shibshib, line 178: can we perhaps add some more documentation as to why we're ensuring that coldkey does NOT own the hotkey?

sam0x17 (approved):

pallets/subtensor/src/migration.rs
orriin: Nope, don't want to panic where possible. 

I didn't write that code, just my formatter formatted it. I'll remove the panic just in case.

distributedstatemachine (approved):

pallets/admin-utils/src/lib.rs
orriin: We can still reduce it, but when we reduce it we don't need to run the check to clear small nominations because no nominations could have dropped below the new min value.

pallets/subtensor/src/migration.rs
sam0x17: awesome, want to eliminate these as much as possible 👍🏻

PR #387, opened 2024-05-01
Upgrade to polkadot-sdk-v1.10.0

orriin:
- [x] Refactor to use workspace deps
- [x] Upgrade runtime 
- [x] Upgrade to latest jsonrpsee
- [x] Upgrade node
- [x] Upgrade tests
- [x] Upgrade chain spec
- [x] Upgrade benchmarks
- [x] Upgrade migrations


sam0x17 (approved):
💯

integration-tests/Cargo.toml
```diff
@@ -12,12 +12,10 @@ workspace = true
 
 [dependencies]
 
+[dev-dependencies]
```
distributedstatemachine: Should we delete this crate ? Its a no op at the moment

Cargo.toml
```diff
@@ -11,6 +11,99 @@ resolver = "2"
 [workspace.lints.clippy]
 type_complexity = "allow"
 
+[workspace.dependencies]
```
distributedstatemachine, line 13: Lisan Al Gaib!

node/src/service.rs
```diff
@@ -13,6 +13,10 @@ use sc_transaction_pool_api::OffchainTransactionPoolFactory;
 use sp_consensus_aura::sr25519::AuthorityPair as AuraPair;
 use std::{sync::Arc, time::Duration};
 
+/// The minimum period of blocks on which justifications will be
+/// imported and generated.
+const GRANDPA_JUSTIFICATION_PERIOD: u32 = 512;
```
distributedstatemachine, line 18: How do we arrive at this value ? Is 512 best practice , or is it implict in our existing code base and exposed in v1.10?

distributedstatemachine (approved):

integration-tests/Cargo.toml
orriin: Yah sounds good

node/src/service.rs
orriin: It was the period used previously, it was just hardcoded rather than in a variable. It was moved to a variable because it's repeated in a couple places now.

PR #398, opened 2024-05-06
ci: clippy deny warnings

orriin:
- Makes clippy warnings fail by passing `--deny warnings`
- Runs clippy twice, once with features once without


sam0x17:
looks like some clippys need fixing with that change now @orriin

sam0x17 (approved):
looks good to me once CI is passing

PR #429, opened 2024-05-16
Remove more direct indexing

keithtensor:
Removes more cases of direct indexing.

### DIFF
--- a/pallets/subtensor/src/registration.rs
+++ b/pallets/subtensor/src/registration.rs
@@ -2,7 +2,6 @@ use super::*;
 use frame_support::storage::IterableStorageDoubleMap;
 use sp_core::{Get, H256, U256};
 use sp_io::hashing::{keccak_256, sha2_256};
-use sp_runtime::MultiAddress;
 use system::pallet_prelude::BlockNumberFor;
 
 const LOG_TARGET: &str = "runtime::subtensor::registration";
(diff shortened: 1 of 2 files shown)


pallets/subtensor/src/registration.rs
```diff
@@ -528,18 +528,14 @@ impl<T: Config> Pallet<T> {
         hash_as_vec
     }
 
-    #[allow(clippy::indexing_slicing)]
     pub fn hash_block_and_hotkey(block_hash_bytes: &[u8; 32], hotkey: &T::AccountId) -> H256 {
-        // Get the public key from the account id.
-        let hotkey_pubkey: MultiAddress<T::AccountId, ()> = MultiAddress::Id(hotkey.clone());
-        let binding = hotkey_pubkey.encode();
-        // Skip extra 0th byte.
-        let hotkey_bytes: &[u8] = binding[1..].as_ref();
```
keithtensor, line 537: This one I'm fairly certain but not 100% sure -- it's actually trying to skip the discriminant byte from the `MultiAddress` enum, right? If so, we can just encode the `hotkey` directly, as I've done in my PR.

camfairchild: This seems right to me

camfairchild (approved):
Looks good here.

distributedstatemachine (approved):

sam0x17 (approved):

PR #439, opened 2024-05-20
Add fuzzing to matrix functions

### LINKED ISSUE #421: Fuzzing
We should add automated fuzzing to the CI particularly to catch bugs around matrix operations and unanticipated combinations of extrinsics potentially.

AC:
- [ ] pick a fuzzing library/crate
- [ ] get fuzzing working
- [ ] get fuzzing working automatically in the CI
- [ ] profit?

keithtensor:
Fixes #421.


sam0x17 (approved):
lgtm

pallets/subtensor/src/math.rs
```diff
@@ -206,6 +206,7 @@ pub fn sigmoid_safe(input: I32F32, rho: I32F32, kappa: I32F32) -> I32F32 {
 
 // Returns a bool vector where an item is true if the vector item is in topk values.
 #[allow(dead_code, clippy::indexing_slicing)]
+#[test_fuzz::test_fuzz]
```
distributedstatemachine, line 230: Is it possible to add bounds to these?  My fear is that the fuzzer would run iterations on values outside what is realistic.

sam0x17: I mean that's probably a feature rather than a bug unless it causes a false positive someday

PR #479, opened 2024-05-27
feat: bad blocks chain extension

### LINKED ISSUE #467: `bad_blocks` chain extension
- [x] Currently `local`, `testnet`, `finney` and common chain spec logic is all in 1 file, which is somewhat unorganised. Split up into dedicated files for better organisation and easier maintenance.
- [x] Add `bad_blocks` chain extensions.
- [x] Test `bad_blocks` locally to ensure it works.

This is a necessary part of our chain recovery SOP (https://github.com/opentensor/subtensor/issues/438).

orriin:
Closes https://github.com/opentensor/subtensor/issues/467

- Improves chain spec logic organisation 
- Adds option to specify `bad_blocks` in chain specs

Tested:
 
Screenshot 2024-05-27 at 08 32 05

Node will log attempts to import bad blocks a few times before banning the peer trying to propagate the bad block.


distributedstatemachine:
Nice ! @orriin  can you please add instructions to test this ? It be also be helpful to try using this to sync an archive node on finney

> Node will log attempts to import bad blocks a few times before banning the peer trying to propagate the bad block.

What happens if all the peers have the badblocks? Does this mean it wont be able to sync as it is banning a peers?

scripts/localnet.sh
```diff
@@ -8,7 +8,7 @@ BASE_DIR="$SCRIPT_DIR/.."
 
 : "${CHAIN:=local}"
 : "${BUILD_BINARY:=1}"
-: "${FEATURES:=pow-faucet}"
+: "${FEATURES:="pow-faucet runtime-benchmarks"}"
```
distributedstatemachine: Do we need the `runtime-benchmarks` for localnet ?

orriin: Yeah it didn't work, got the same error that you DM'd me about

orriin:
> Nice ! @orriin can you please add instructions to test this ? It be also be helpful to try using this to sync an archive node on finney

1. Sync Alice and Bob some blocks
2. Stop Bob
3. Add latest block hash to bad blocks
4. Regen chain spec
5. Start Alice again

> What happens if all the peers have the badblocks? Does this mean it wont be able to sync as it is banning a peers?

Yeah, it can't sync from them anyway because they have bad blocks. Dw though, the ban list is cleared when the node restarts.

distributedstatemachine:
@orriin please resolve conflicts

distributedstatemachine (approved):

PR #480, opened 2024-05-27
Enable exponential backoff for AuRa

keithtensor:
This PR enables exponential backoff for the block authoring gadget AuRa. It uses the default backoff settings, i.e. it never waits more than 100 blocks before authoring, it starts considering backing off after 50 or more unfinalized blocks, and it is biased to consider the unfinalized head suffix length to grow half as fast as in actuality.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -4645,6 +4645,7 @@ dependencies = [
  "sc-consensus-aura",
  "sc-consensus-grandpa",
  "sc-consensus-grandpa-rpc",
+ "sc-consensus-slots",
  "sc-executor",
  "sc-keystore",
  "sc-network",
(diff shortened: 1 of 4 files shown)


camfairchild (approved):

orriin (approved):

gztensor:
Wow, this is great we have it now! How do we test exponential backup? Should we add some enormous size test migration?

distributedstatemachine:
> Wow, this is great we have it now! How do we test exponential backup? Should we add some enormous size test migration? 

@gztensor would be pretty hard to recreate , but keen on ideas. We need a large existing state , with  a lot of nodes , similar to finney. Might be worth recreating during the finney dry run, and using a pre v.1 snapshot

PR #485, opened 2024-05-28
Ban unsafe arithmetic operations

### LINKED ISSUE #303: deny unsafe math operations
Part of #300, we need to deny the ability to perform potentially-panicking and/or overflowing operations on number types in subtensor

AC:
- [ ] fix existing offending overflow operations
- [ ] fix existing offending float usages, if any
- [ ] enforce `#[deny(arithmetic_overflow)]`
- [ ] enforce `clippy::integer_arithmetic`
- [ ] disallow the use of `f32` / `f64` entirely

keithtensor:
Devnet Companion: #529

This PR bans raw arithmetic operations such as those with the `+ - / *` symbols, and instead replaces them with either the saturating equivalent, or the checked equivalent.

Resolves #303.

### DIFF
--- a/pallets/collective/src/lib.rs
+++ b/pallets/collective/src/lib.rs
@@ -54,7 +54,7 @@ use frame_support::{
 use scale_info::TypeInfo;
 use sp_io::storage;
 use sp_runtime::traits::Dispatchable;
-use sp_runtime::{traits::Hash, RuntimeDebug};
+use sp_runtime::{traits::Hash, RuntimeDebug, Saturating};
 use sp_std::{marker::PhantomData, prelude::*, result};
 
 #[cfg(test)]
@@ -119,7 +119,7 @@ impl DefaultVote for MoreThanMajorityThenPrimeDefaultVote {
         _no_votes: MemberCount,
         len: MemberCount,
     ) -> bool {
-        let more_than_majority = yes_votes * 2 > len;
+        let more_than_majority = yes_votes.saturating_mul(2) > len;
         more_than_majority || prime_vote.unwrap_or(false)
     }
 }
@@ -545,7 +545,9 @@ pub mod pallet {
                 Error::<T, I>::DurationLowerThanConfiguredMotionDuration
             );
 
-            let threshold = (T::GetVotingMembers::get_count() / 2) + 1;
+            let threshold = T::GetVotingMembers::get_count()
+                .saturating_div(2)
+                .saturating_add(1);
 
             let members = Self::members();
             let (proposal_len, active_proposals) =
@@ -716,10 +718,15 @@ impl<T: Config<I>, I: 'static> Pallet<T, I> {
             })?;
 
         let index = Self::proposal_count();
-        <ProposalCount<T, I>>::mutate(|i| *i += 1);
+        <ProposalCount<T, I>>::try_mutate(|i| {
+            *i = i
+                .checked_add(1)
+                .ok_or(Error::<T, I>::TooManyActiveProposals)?;
+            Ok::<(), Error<T, I>>(())
+        })?;
         <ProposalOf<T, I>>::insert(proposal_hash, proposal);
         let votes = {
-            let end = frame_system::Pallet::<T>::block_number() + duration;
+            let end = frame_system::Pallet::<T>::block_number().saturating_add(duration);
             Votes {
                 index,
                 threshold,
@@ -862,10 +869,10 @@ impl<T: Config<I>, I: 'static> Pallet<T, I> {
         // default voting strategy.
         let default = T::DefaultVote::default_vote(prime_vote, yes_votes, no_votes, seats);
 
-        let abstentions = seats - (yes_votes + no_votes);
+        let abstentions = seats.saturating_sub(yes_votes.saturating_add(no_votes));
         match default {
-            true => yes_votes += abstentions,
-            false => no_votes += abstentions,
+            true => yes_votes = yes_votes.saturating_add(abstentions),
+            false => no_votes = no_votes.saturating_add(abstentions),
         }
         let approved = yes_votes >= voting.threshold;
 
@@ -981,7 +988,7 @@ impl<T: Config<I>, I: 'static> Pallet<T, I> {
         Voting::<T, I>::remove(proposal_hash);
         let num_proposals = Proposals::<T, I>::mutate(|proposals| {
             proposals.retain(|h| h != &proposal_hash);
-            proposals.len() + 1 // calculate weight based on original length
+            proposals.len().saturating_add(1) // calculate weight based on original length
         });
         num_proposals as u32
     }
@@ -1154,7 +1161,7 @@ impl<
     type Success = ();
     fn try_origin(o: O) -> Result<Self::Success, O> {
         o.into().and_then(|o| match o {
-            RawOrigin::Members(n, m) if n * D > N * m => Ok(()),
+            RawOrigin::Members(n, m) if n.saturating_mul(D) > N.saturating_mul(m) => Ok(()),
             r => Err(O::from(r)),
         })
     }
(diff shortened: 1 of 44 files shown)


sam0x17:
looks like some conflicts + CI failures now @keithtensor

sam0x17 (approved):

pallets/collective/src/lib.rs
```diff
@@ -983,7 +988,7 @@ impl<T: Config<I>, I: 'static> Pallet<T, I> {
         Voting::<T, I>::remove(proposal_hash);
         let num_proposals = Proposals::<T, I>::mutate(|proposals| {
             proposals.retain(|h| h != &proposal_hash);
-            proposals.len() + 1 // calculate weight based on original length
+            proposals.len().saturating_add(1) // calculate weight based on original length
```
orriin, line 991:
```suggestion
            proposals.len().saturating_inc() // calculate weight based on original length
```

pallets/subtensor/src/block_step.rs
```diff
@@ -27,9 +28,10 @@ impl<T: Config> Pallet<T> {
         Ok(())
     }
 
-    // Helper function which returns the number of blocks remaining before we will run the epoch on this
-    // network. Networks run their epoch when (block_number + netuid + 1 ) % (tempo + 1) = 0
-    //
+    #[allow(clippy::arithmetic_side_effects)]
```
orriin, line 31: Why allowed?

runtime/src/lib.rs
```diff
@@ -646,7 +650,11 @@ impl PrivilegeCmp<OriginCaller> for OriginPrivilegeCmp {
                     r_yes_votes,
                     r_count,
                 )), // Equivalent to (l_yes_votes / l_count).cmp(&(r_yes_votes / r_count))
-            ) => Some((l_yes_votes * r_count).cmp(&(r_yes_votes * l_count))),
+            ) => Some(
+                l_yes_votes
+                    .saturating_mul(*r_count)
+                    .cmp(&r_yes_votes.saturating_mul(*l_count)),
```
camfairchild, line 672:
```suggestion
                    .cmp(&r_yes_votes.saturating_mul(l_count)),
```

Should not be a dereference, I believe

keithtensor: That's incorrect, both `left` and `right` are references, and so when they're matched against, all their internal fields are also references. If I don't dereference, I'd be multiplying a reference here, not the actual number.

All saturating math functions aren't like the regular math operators where there's an auto-deref of the operands other than the operand on the callee, so this deref is necessary.

More importantly, this is the kind of thing that the Rust compiler catches, so if your belief is true, then I would expect the compiler to throw errors as it is now.

sam0x17: yeah CI would definitely catch this if it was incorrect

pallets/subtensor/src/lib.rs
```diff
@@ -1147,15 +1147,21 @@ pub mod pallet {
 
                     Stake::<T>::insert(hotkey.clone(), coldkey.clone(), stake);
 
-                    next_uid += 1;
+                    next_uid = next_uid.checked_add(1).expect(
+                        "should not have total number of hotkey accounts larger than u16::MAX",
+                    );
```
orriin, line 1188: u16::MAX is only 65535. is this enough? I think we're close to that, if not already exceeded it?

pallets/collective/src/lib.rs
keithtensor: I don't think this works, since `saturating_inc` doesn't return the sum afterwards.

pallets/subtensor/src/block_step.rs
keithtensor: Because if not, it spits out weird errors about how there's a side effect in this function, but I can't seem to find where it is.

pallets/subtensor/src/lib.rs
keithtensor: If so, how does this line currently work without my changes?

sam0x17: yeah this is fine

pallets/subtensor/src/block_step.rs
sam0x17: yeah looking at that function, there is simply no way that could have side effects, and gpt agrees, so probably something to do with macro expansion confusing clippy. Seems legit 👍🏻

sam0x17 (dismissed):
I had gpt 4o go over the diff of this in detail with me as well and it is satisfied all of these preserve the original behavior after a long convo

pallets/subtensor/src/lib.rs
```diff
@@ -1147,15 +1147,21 @@ pub mod pallet {
 
                     Stake::<T>::insert(hotkey.clone(), coldkey.clone(), stake);
 
-                    next_uid += 1;
+                    next_uid = next_uid.checked_add(1).expect(
+                        "should not have total number of hotkey accounts larger than u16::MAX",
```
orriin, line 1187:
```suggestion
                        "genesis state should not have a total number of hotkey accounts larger than u16::MAX",
```

pallets/subtensor/src/staking.rs
```diff
@@ -343,7 +343,7 @@ impl<T: Config> Pallet<T> {
         Self::set_stakes_this_interval_for_coldkey_hotkey(
             &coldkey,
             &hotkey,
-            stakes_this_interval + 1,
+            stakes_this_interval.saturating_add(1),
```
distributedstatemachine, line 349: inconsistencies between using `saturating_inc` and `saturating_add` for increments of 1

pallets/subtensor/src/staking.rs
```diff
@@ -456,7 +456,7 @@ impl<T: Config> Pallet<T> {
         Self::set_stakes_this_interval_for_coldkey_hotkey(
             &coldkey,
             &hotkey,
-            unstakes_this_interval + 1,
+            unstakes_this_interval.saturating_add(1),
```
distributedstatemachine, line 455: same as above

pallets/subtensor/src/uids.rs
```diff
@@ -60,7 +60,7 @@ impl<T: Config> Pallet<T> {
         );
 
         // 2. Get and increase the uid count.
-        SubnetworkN::<T>::insert(netuid, next_uid + 1);
+        SubnetworkN::<T>::insert(netuid, next_uid.saturating_add(1));
```
distributedstatemachine, line 62: saturating_add v inc

distributedstatemachine (dismissed):
Approved but lets wait for @sam0x17 on merging this

sam0x17 (dismissed):
changes look good, next step would be for a companion PR to open, with a link to this branch, that merges a this branch into `devnet-ready`

JohnReedV (dismissed):

sam0x17:
So yeah again, should have a companion that merges this branch into `devnet-ready`. Not sure what this `opentensor:development` is or why people keep opening PRs into it 😕

sam0x17:
but yeah if this is already tested on devnet, should add the `devnet-pass` label, if this is already tested on testnet, should add the `testnet-pass` label, at which point it should be able to merge into `main`

Mananitade:
I'm curious about the outright ban of the operations + - / *, since these are rather common operations. Have you considered an alternative of catching overflow exceptions? To me, it seems like an overkill, but I'm open to being convinced otherwise. Is it considered good practice in the Rust community to ban these operations? What are the tradeoffs (how much slower does it make things run)?

sam0x17 (dismissed):

sam0x17:
> I'm curious about the outright ban of the operations + - / *, since these are rather common operations. Have you considered an alternative of catching overflow exceptions? To me, it seems like an overkill, but I'm open to being convinced otherwise. Is it considered good practice in the Rust community to ban these operations? What are the tradeoffs (how much slower does it make things run)?

main reason being panicking in an extrinsic is enough to brick a substrate-based chain. On parity's end they have always said you should only use checked math in extrinsics.

JohnReedV (dismissed):

sam0x17 (approved):

distributedstatemachine (approved):

JohnReedV (approved):

PR #329, opened 2024-04-15
hotfix(docker): Use raw_testspec for test nodes

birkskyum:
See https://github.com/opentensor/subtensor/pull/325

PR #493, opened 2024-05-30
fix: chainspec irregularities 

### LINKED ISSUE #489: Chainspec consistency
The raw chain specs have not been regenerated since upgrading to v1.0, and Medulla team is having issues with them.

Need to 
- [x] Check if any manual changes have been made to the raw chain specs
- [x] Regenerate the chain specs to align with the definitions in the code
- [x] Add bootnodes
- [x] Add CI to enforce the raw chain spec is consistent

orriin:
Closes https://github.com/opentensor/subtensor/issues/489

- Sets SS58 format in Rust chain spec definitions to match chain spec JSON
- Sets missing `protocol_id` in Rust chain spec definitions
- Rebuilds `plain_spec_finney.json`, `plain_spec_testfinney.json`, `raw_spec_finney.json`, `raw_spec_testfinney.json`
- Removes `finney_plain_spec.json`, `raw_spec.json`, `raw_testspec.json` since they appeared stale and I do not understand their purpose 
- Adds CI that builds and commits updated chain specs on push to branches `development`, `staging`, `main`

TODO
- [x] Check if we need bootnodes
- [x] Create script to easily update chainspecs
- [x] Add CI to enforce JSON chain spec stays consistent with the Rust definitions


prop-opentensor:
bootnodes:

finney: `/dns/bootnode.finney.chain.opentensor.ai/tcp/30333/ws/p2p/12D3KooWRwbMb85RWnT8DSXSYMWQtuDwh4LJzndoRrTDotTR5gDC`
testnet: 
`/dns/bootnode.test.chain.opentensor.ai/tcp/30333/p2p/12D3KooWPM4mLcKJGtyVtkggqdG84zWrd7Rij6PGQDoijh1X86Vr`


fyi: `raw_testspec.json` is currently in use on devnet validators - since we plan to replace devnet with the testnet mirror, imo this can still be removed

orriin:
> fyi: raw_testspec.json is currently in use on devnet validators - since we plan to replace devnet with the testnet mirror, imo this can still be removed

Cool, yeah I'm removing all of `finney_plain_spec.json`, `raw_spec.json`, `raw_testspec.json`, they seem to just be duplicates.

birkskyum:
related - i just noticed that the mainnet spec files was also used for people wanting to start at testnet docker container, but it might be obsoleted if those spec files are now deleted
- https://github.com/opentensor/subtensor/pull/329

Also, these raw_spec.json is mentioned in docs [all over the place](https://github.com/opentensor/developer-docs/issues/221), which would be good to have synced with this change

birkskyum:
out of curiosity, what is the difference between raw_spec and plain_spec?

orriin:
> out of curiosity, what is the difference between raw_spec and plain_spec?

See the 'Raw chain specifications' section at these docs https://docs.substrate.io/build/chain-spec/

birkskyum:
@orriin , can you update the docker-compose and scripts/run/subtensor.sh too in a way that aligns with this change? they use the raw_spec.json / raw_testspec.json that are removed in this PR and will thus break.

Also, [running a lite node on testchain with docker](https://docs.bittensor.com/subtensor-nodes/using-docker#run-a-lite-node-on-testchain) breaks because the subtensor.sh doesn't conditionally use a spec for the testnet when setting the test flag, which I resolved in ( #329 ) , but since this PR obsoletes that work it would be reasonable to include the fix here with the name of the new main/test spec files.

Dockerfile
```diff
@@ -60,5 +60,4 @@ FROM $BASE_IMAGE AS subtensor
 
 COPY --from=builder /subtensor/snapshot.json /
 COPY --from=builder /subtensor/raw_spec.json /
```
birkskyum: isn't raw_spec removed as well?

birkskyum: This need adjustment too, both for mainnet/testnet flags - otherwise docker won't work

https://github.com/opentensor/subtensor/blob/e8eae823ebf581bc8b90498eaa22bbdae976701f/scripts/run/subtensor.sh#L45

sam0x17 (approved):

sam0x17 (approved):

camfairchild (approved):

keithtensor (approved):

rajkaramchedu:
We don't use the word "finney" in the published docs. We use "mainnet" or "testnet" or "devnet" or "local". Given this, for clarifying what's what in my docs, are the below two statements accurate?

- The `raw_spec_finney.json` spec is for mainnet.
- The `raw_spec_testfinney.json` spec is for testnet.

Thanks

orriin:
> We don't use the word "finney" in the published docs. We use "mainnet" or "testnet" or "devnet" or "local". Given this, for clarifying what's what in my docs, are the below two statements accurate?
> 
> * The `raw_spec_finney.json` spec is for mainnet.
> * The `raw_spec_testfinney.json` spec is for testnet.
> 
> Thanks

This is correct :)

sam0x17 (approved):

PR #533, opened 2024-06-13
prevent accidental changes to storage structs

### LINKED ISSUE #513: prevent accidental changes to storage structs
diffs like the following can be extremely tricky/insidious in the context of a substrate-based blockchain:
```diff
+    /// Data structure for Axon information.
     #[derive(Encode, Decode, Default, TypeInfo, Clone, PartialEq, Eq, Debug)]
     pub struct AxonInfo {
-        pub block: u64,       // --- Axon serving block.
-        pub version: u32,     // --- Axon version
-        pub ip: u128,         // --- Axon u128 encoded ip address of type v6 or v4.
-        pub port: u16,        // --- Axon u16 encoded port.
-        pub ip_type: u8,      // --- Axon ip type, 4 for ipv4 and 6 for ipv6.
-        pub protocol: u8,     // --- Axon protocol. TCP, UDP, other.
-        pub placeholder1: u8, // --- Axon proto placeholder 1.
-        pub placeholder2: u8, // --- Axon proto placeholder 2.
-    }
-
-    // --- Struct for Prometheus.
+       ///  Axon serving block.
+       pub block: u64,
+       ///  Axon version
+       pub version: u32,
+       ///  Axon u128 encoded ip address of type v6 or v4.
+       pub port: u16,
+       ///  Axon u16 encoded port.
+       pub ip: u128,
+       ///  Axon ip type, 4 for ipv4 and 6 for ipv6.
+       pub ip_type: u8,
+       ///  Axon protocol. TCP, UDP, other.
+       pub protocol: u8,
+       ///  Axon proto placeholder 1.
+       pub placeholder1: u8,
+       ///  Axon proto placeholder 2.
+       pub placeholder2: u8,
    }
```

In this particular case, the relative order of `ip` and `port` was quietly changed, resulting in decoding errors on the bittensor side since the order of these struct fields determines how they are laid out in memory, and thus, how they must be decoded/encoded.

Similarly, a frequent gotcha in substrate-based chains is subtle changes to storage structs without a matching migration to enact this change / storage upgrade.

These types of issues make up a sizeable percentage of critical issues that make it to testnet or beyond before they are detected, and can be extremely damaging since they often render the underlying storage field(s) undecodable or corrupted, which can have a wide variety of exotic and severe consequences.

The main problems is we want developers to be aware when they have accidentally made a non-trivial change to a storage struct. Ideally they should be prompted in some way to create a storage migration or at least be made aware if they re-order, rename, add, or remove fields on a storage struct.

## Solution

```rust
#[freeze_layout("13c78c707b010724")]
#[derive(Encode, Decode, Default, TypeInfo, Clone, PartialEq, Eq, Debug)]
pub struct AxonInfo {
    ...
}
```

I propose a simple solution that will prevent people from accidentally making changes to storage structs. The idea is to create an attribute macro called `freeze_layout` that can be attached to storage structs and other change-sensitive rust items. This attribute will take a single argument which is a hash code of the _non-trivial_ (ignoring doc comments) AST nodes that make up the item the attribute is attached to.

If the hash code is not provided: A compiler error will be displayed providing the programmer with the proper hash code which they can then paste in.

If the hash code does not match: A compiler error will be displayed informing the programmer that they have made a non-trivial/structural change to a storage struct and that they will likely need to write a migration to enact this change. The correct hash code will also be displayed.

This is based largely on my "[audited](https://forum.polkadot.network/t/rfc-trustless-in-code-auditing-annotations/988/5)" idea originally posted on the polkadot forums, though in this case it would be a lot simpler.

## AC

- [x] attribute macro implementation, including a syn visitor pattern that calculates the hash code for the underlying AST nodes, ignoring trivial nodes like doc comments. Should be sensitive to node order/structure. My code here is a really good starting point: https://github.com/sam0x17/audited
- [x] apply `freeze_layout` to all storage structs in subtensor
- [ ] some way of _requiring_ `freeze_layout` on storage structs? clippy? quick and dirty custom visitor running in a bin target that becomes a ci job?

sam0x17:
Fixes #513 

devnet companion: #550, #554


camfairchild (dismissed):

open-junius:
Look good! I have a question, how to get the hash value in freeze_struct()?

sam0x17:
> Look good! I have a question, how to get the hash value in freeze_struct()?

from the compile error

open-junius (dismissed):

distributedstatemachine (dismissed):

distributedstatemachine (approved):

support/macros/tests/tests.rs
```diff
@@ -0,0 +1,7 @@
+use subtensor_macros::freeze_struct;
+
+#[freeze_struct("ecdcaac0f6da589a")]
```
kianenigma, line 3: How can one calculate this hash in an easy way?

sam0x17: so right now the dev ux is basically you are in one of two scenarios:
1. you are adding it to a new struct, in which case you just do `#[freeze_struct]` with zero args and the compiler error will immediately tell you the proper hash code
2. you are updating an existing struct that already has a hash code, in which case the compiler error the second you change the struct will tell you the new expected hash code, and also encourages you to write a migration etc. if applicable

sam0x17: people running RA will see it literally in their IDE as well

sam0x17: also worth noting it is immune to doc comment changes, they aren't part of the hash code calculation, and arbitrary additional nodes to ignore could also be added to the visitor pattern if desired

kianenigma: I see, that's cool! but will this work also for generic structs? It seems like it will also break if you rename a generic or field, even if it is the same encoding? 

I think it is certainly a cool idea, and if no similar rust crate exists, we can make it an independent Rust crate, and reference it in our docs.

sam0x17: Yeah it could be modified to be immune to name changes. I wasn't sure if that was safe in all cases so I left that out of this one

sam0x17: Same goes for generics, whatever you want it to ignore in the hash code calculation can be added to the visitor pattern

open-junius (approved):

JohnReedV (approved):

support/macros/src/lib.rs
```diff
@@ -0,0 +1,68 @@
...
+            format!("You must provide a hashcode in the `freeze_struct` attribute to freeze this struct.\n\n\
+            expected hashcode: `#[freeze_struct(\"{calculated_hash_hex}\")]`"),
+        ));
+    }
+
+    let parsed_attr = parse2::<LitStr>(attr)?;
+    let provided_hash_hex = parsed_attr.value().to_lowercase();
+
+    let mut visitor = CleanDocComments::new();
+    visit_item_struct_mut(&mut visitor, &mut item_clone);
```
keithtensor: I'm still not seeing how `item_clone` is later used in this function.

sam0x17:
```
        let mut item_clone = item.clone();
        let mut cleaner = CleanDocComments;
        cleaner.visit_item_mut(&mut item_clone);

        // Calculate the hash of the cleaned item
        let hash_after = generate_hash(&item_clone);
```

sam0x17: the visit_item_mut mutates `item_clone` and then we calculate the hash code from it and store that in `hash_after`

orriin (approved):

support/macros/src/lib.rs
sam0x17: you are totally right, fixing

runtime/src/lib.rs
```diff
@@ -139,7 +139,7 @@ pub const VERSION: RuntimeVersion = RuntimeVersion {
     //   `spec_version`, and `authoring_version` are the same between Wasm and native.
     // This value is set to 100 to notify Polkadot-JS App (https://polkadot.js.org/apps) to use
     //   the compatible custom types.
-    spec_version: 154,
+    spec_version: 155,
```
keithtensor: Should we just leave this unchanged?

keithtensor (approved):

PR #539, opened 2024-06-15
fix can_commit to use get_weights_set_rate_limit

surcyf123:
## Description
Previously, the can_commit function checked to see if a whole commit_reveal_weights_interval has passed before being allowed to commit weights again. Instead, we should keep this logic for when we can reveal_weights (as is intended), but the neuron should be able to commit weights multiple times within a single commit_reveal_weights_interval according to the weights_rate_limit hyperparameter. If not, the weights_rate_limit is essentually useless because the can_commit function overrides the weights_rate_limit. 

Furthermore, it introduces an edge case where if commit_reveal_weights_interval if > activity_cutoff, all weights go away and the network stalls because neurons are now setting weights > once per activity_cutoff. This PR fixes this too.

## Type of Change

- [ ] Documentation update

## Checklist

## Additional Notes

### DIFF
--- a/.github/workflows/check-rust.yml
+++ b/.github/workflows/check-rust.yml
@@ -65,101 +65,101 @@ jobs:
       - name: cargo fmt
         run: cargo fmt --check --all
 
-  # cargo-clippy-default-features:
-  #   name: cargo clippy
-  #   runs-on: SubtensorCI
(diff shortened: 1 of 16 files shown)


surcyf123:
forked from wrong base branch

PR #550, opened 2024-06-19
[devnet companion] prevent accidental changes to storage structs

sam0x17:
devnet companion for #533


open-junius (dismissed):

sam0x17:
note to reviewers: some extra changes are included that match keith's changes to appease clippy for the panic stuff

distributedstatemachine (approved):

support/macros/src/lib.rs
```diff
@@ -0,0 +1,68 @@
...
+            format!("You must provide a hashcode in the `freeze_struct` attribute to freeze this struct.\n\n\
+            expected hashcode: `#[freeze_struct(\"{calculated_hash_hex}\")]`"),
+        ));
+    }
+
+    let parsed_attr = parse2::<LitStr>(attr)?;
+    let provided_hash_hex = parsed_attr.value().to_lowercase();
+
+    let mut visitor = CleanDocComments::new();
+    visit_item_struct_mut(&mut visitor, &mut item_clone);
```
keithtensor, line 54: What's the purpose of the `item_clone` variable? It doesn't seem to have served any purpose other than to have been visited by this visitor.

sam0x17: the visitor mutates it to have all blank doc comments, so we don't want to apply that to the real AST

keithtensor: But where is the `item_clone` then used? It's been modified, but then I don't see it being referenced anymore after this line.

sam0x17: you are actually right, it should be `item_clone` is used to generate the hash, very nice subtle find. Apparently it all works without this, but this explains some weird results I was getting with things that had #[cfg_attr]

sam0x17: resolved in #554

PR #553, opened 2024-06-21
Implement safe mode pallet

### LINKED ISSUE #535: Add Safe Mode pallet to runtime
We would like to add the Safe Mode pallet to our subtensor runtime. This pallet allows a highly privileged user such as root to put the entire chain into "safe mode" for a set period of time, whereby only a select few extrinsics can be executed, while all others will be rejected. While in safe mode, the privileged user can also choose to extend the duration of the safe mode period.

# Acceptance Criteria
- [x] Figure out and configure who should have the authority to put the chain into and out of safe mode (root)
- [ ] Figure out and configure how long safe mode can last
- [ ] Figure out and configure how long an extension can last
- [ ] Add the configured pallet to the runtime

keithtensor:
Devnet Companion: ~#558~ #559  

Resolves #535.

### DIFF
--- a/runtime/src/lib.rs
+++ b/runtime/src/lib.rs
@@ -14,9 +14,9 @@ use frame_support::{
     dispatch::DispatchResultWithPostInfo,
     genesis_builder_helper::{build_config, create_default_config},
     pallet_prelude::{DispatchError, Get},
-    traits::{fungible::HoldConsideration, LinearStoragePrice},
+    traits::{fungible::HoldConsideration, Contains, LinearStoragePrice},
 };
-use frame_system::{EnsureNever, EnsureRoot, RawOrigin};
+use frame_system::{EnsureNever, EnsureRoot, EnsureRootWithSuccess, RawOrigin};
 use pallet_commitments::CanCommit;
 use pallet_grandpa::{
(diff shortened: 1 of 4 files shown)


runtime/src/lib.rs
```diff
@@ -284,6 +284,49 @@ impl pallet_utility::Config for Runtime {
...
+    fn contains(call: &RuntimeCall) -> bool {
+        matches!(
+            call,
+            RuntimeCall::Sudo(_)
+                | RuntimeCall::System(_)
+                | RuntimeCall::SafeMode(_)
+                | RuntimeCall::Timestamp(_)
+        )
+    }
+}
```
sam0x17, line 310: just wanted to double check we don't want to whitelist any other calls, and that we can perform a runtime upgrade when in safemode as that seems a likely safemode activity

keithtensor: I think this is a safe default, the calls here are basic operations critical for the blockchain, and as such runtime upgrades can still be performed. No other pallets that are irrelevant are included here.

sam0x17 (approved):

distributedstatemachine (approved):

JohnReedV (approved):

Issue #563, opened 2024-06-24
Total issuance discrepancy 


orriin:
`balance.totalIssuance`: 1,195,894,993,084,044 (which should be effectively the number of coins in circulation)
`subtensorModule.totalissuance`: 7,010,451,064,912,511
`subtensorModule.totalstake`: 5,780,329,269,388,411

mogmachine:
I think this stems from the way burned and recycled tokens are or are not counted in the various totals, but this is critical we get right as exchanges are relying on us for the truth and we are showing discrepancies in our data.

orriin:
## Initial analysis

<img width="423" alt="Screenshot 2024-06-25 at 16 44 51" src="https://github.com/opentensor/subtensor/assets/167025436/96d7fb0e-4656-40c6-a5c1-a55583b3c4a7">

Good news:
- Balances `TotalIssuance` amount is consistent with all free + reserved tokens in existence
- Subtensor `TotalStaked` amount is consistent with sum of staked amounts across all accounts
- TAOStats reports the correct TI (6.97M)

Inconsistency:
- Subtensor `TotalIssuance` is `35651447307410` (approximately 35k TAO or 0.5%) higher than it should be.

## Next Steps

1. Identify which code path/s decrease the stake or account balance without making a corresponding change to Subtensor TI. Will collab with @mogmachine on that, it sounds like he already has an idea.
2. Fix the issue, so TI changes no longer deviate from stake or acc balance changes
3. Adjust the TI back to its correct value
4. Add a `try-state` invariant to immediately notify us if state becomes inconsistent again

## Long Term

We should migrate our staking system to use Holds instead of the current `Stake` storage, this can be achieved as a lazy migration. Once migrated to Holds, we can remove all the "double accounting" that we currently do to track total issuance.

orriin:
Confirmed the cause is with subnet registration fees and tx fees.

Subnet registration makes up for the overwhelming majority of the 35k discrepancy, only ~7.5 TAO comes from tx fees.

## Causes
 
### Subnet registration

When a user registers a new network here:

https://github.com/opentensor/subtensor/blob/c097ff8fb0d434a5541b62ab71a9c406e0daa05f/pallets/subtensor/src/lib.rs#L2016-L2023

balance is "burnt" from the registrar without a corresponding update to the subtensor total issuance:

[...]

PR #595, opened 2024-06-28
Machinery to fix total issuance accounting and keep it fixed

orriin:
# DO NOT MERGE UNTIL CONST HAS APPROVED

Machinery for #563 

- Adds documentation for `pallet_subtensor::TotalIssuance`
- Introduces new storage item `TotalSubnetLocked` to track the total amount of tokens locked for subnet registration
- Introduces migration to initialize `TotalSubnetLocked` and then rejig the total issuance to be the sum of issued tokens, staked tokens, and tokens locked for subnet registration
- Adjusts `set_subnet_locked_balance` to update `TotalSubnetLocked` to reflect changes in `SubnetLocked`
- Creates custom `on_nonzero_unbalanced` handler for transaction fees, which will remove the burnt transaction fee amount from `pallet_subtensor::TotalIssuance` before finally dropping it
  - Tested on localnet 
- Adds `try-state` checks for
  - `TotalStake`
  - `TotalSubnetLock`
  - `TotalIssuance`


pallets/admin-utils/src/lib.rs
```diff
@@ -26,7 +26,7 @@ pub mod pallet {
 
     /// Configure the pallet by specifying the parameters and types on which it depends.
     #[pallet::config]
-    pub trait Config: frame_system::Config {
+    pub trait Config: frame_system::Config + pallet_subtensor::Config {
```
open-junius, line 29: For TotalSubnetLocked, we can set it via a function as other hyperparameter. Why introduce the pallet_subtensor::Config this time. is anything special?

open-junius: If it is a temp solution, we can easily remove it later?

pallets/subtensor/src/lib.rs
```diff
@@ -2051,6 +2099,42 @@ pub mod pallet {
...
+        /// total amount staked + the total amount locked in subnets.
+        #[pallet::call_index(63)]
+        #[pallet::weight((
+            Weight::default()
+                .saturating_add(T::DbWeight::get().reads(3))
+                .saturating_add(T::DbWeight::get().writes(1)),
+            DispatchClass::Normal,
+            Pays::Yes
+        ))]
+        pub fn rejig_total_issuance(origin: OriginFor<T>) -> DispatchResult {
```
open-junius: Can we put it in on_finalize, then it can update total issuance automatically

pallets/admin-utils/src/lib.rs
orriin: It's the idiomatic way to achieve this type of pallet coupling in Substrate, it allows us to directly access items from the subtensor pallet without needing all the boilerplate of specifying and implementing the interface. Notice I didn't need any additional boilerplate in my changes here, I just directly access pallet_subtensor storage items. I'll be doing a show and tell on this in a couple weeks :)

pallets/subtensor/src/lib.rs
orriin: No need, we only need to call it once. After that, it will stay synced on its own (unless there's an edge case I missed 😅).

pallets/subtensor/src/lib.rs
```diff
@@ -322,6 +323,16 @@ pub mod pallet {
     pub type MinTake<T> = StorageValue<_, u16, ValueQuery, DefaultMinTake<T>>;
     #[pallet::storage] // --- ITEM ( global_block_emission )
     pub type BlockEmission<T> = StorageValue<_, u64, ValueQuery, DefaultBlockEmission<T>>;
+
+    /// The Subtensor [`TotalIssuance`] represents the total issuance of tokens on the Bittensor network.
+    ///
+    /// It comprised of three parts:
```
camfairchild:
```suggestion
    /// It is comprised of three parts:
```

pallets/subtensor/src/lib.rs
```diff
@@ -655,6 +666,9 @@ pub mod pallet {
     #[pallet::storage] // --- MAP ( netuid ) --> subnet_locked
     pub type SubnetLocked<T: Config> =
         StorageMap<_, Identity, u16, u64, ValueQuery, DefaultSubnetLocked<T>>;
+    /// The total amount of locked tokens in subnets.
```
camfairchild:
```suggestion
    /// The total amount of locked tokens in subnets for subnet registration.
```

open-junius (dismissed):

distributedstatemachine (dismissed):

pallets/subtensor/src/lib.rs
```diff
@@ -322,6 +323,16 @@ pub mod pallet {
...
+
+    /// The Subtensor [`TotalIssuance`] represents the total issuance of tokens on the Bittensor network.
+    ///
+    /// It is comprised of three parts:
+    /// - The total amount of issued tokens, tracked in the TotalIssuance of the Balances pallet
+    /// - The total amount of tokens staked in the system, tracked in [`TotalStake`]
+    /// - The total amount of tokens locked up for subnet reg, tracked in [`TotalSubnetLocked`]
+    ///
+    /// Eventually, Bittensor should migrate to using Holds afterwhich time we will not require this
+    /// seperate accounting.
```
camfairchild:
```suggestion
    /// separate accounting.
```

pallets/subtensor/src/utils.rs
```diff
@@ -324,6 +324,13 @@ impl<T: Config> Pallet<T> {
     }
 
     pub fn set_subnet_locked_balance(netuid: u16, amount: u64) {
+        let prev_total = TotalSubnetLocked::<T>::get();
+        let prev = SubnetLocked::<T>::get(netuid);
+
+        // Deduct the prev amount and add the new amount to the total
```
camfairchild:
```suggestion
        // Deduct the previous amount and add the new amount to the total
```

camfairchild (commented):
Do we have a test for the total issuance fees hook? If not, that would be great to add.

orriin:
> Do we have a test for the total issuance fees hook? If not, that would be great to add.

It will be tested every pr with live state in the existing try-runtime tests

open-junius (dismissed):

camfairchild (dismissed):

sam0x17 (dismissed):

pallets/subtensor/src/lib.rs
```diff
@@ -2402,6 +2334,46 @@ pub mod pallet {
             }
             true
         }
+
+        #[cfg(feature = "try-runtime")]
```
distributedstatemachine, line 2338: I like this alot . We should probably add more maps to this.

runtime/src/migrations/initialise_ti.rs
```diff
@@ -0,0 +1,46 @@
+use frame_support::traits::{fungible, OnRuntimeUpgrade};
+
+use crate::*;
+
+pub struct Migration;
+
+impl OnRuntimeUpgrade for Migration {
```
distributedstatemachine: What is the advantage of having this migration in the runtime? I find it confusing that we have some migrations in the pallet , and some in the runtime. Can we move it to the subtensor module?

distributedstatemachine (changes requested):

pallets/admin-utils/src/lib.rs
```diff
@@ -1035,6 +1035,18 @@ pub mod pallet {
             T::Subtensor::ensure_subnet_owner_or_root(origin.clone(), netuid)?;
             T::Subtensor::do_set_alpha_values(origin, netuid, alpha_low, alpha_high)
         }
+
+        /// Sets the [`pallet_subtensor::TotalSubnetLocked`] value.
+        #[pallet::call_index(52)]
+        #[pallet::weight(T::WeightInfo::sudo_set_commit_reveal_weights_enabled())]
+        pub fn sudo_set_total_subnet_locked(origin: OriginFor<T>, amount: u64) -> DispatchResult {
```
distributedstatemachine: Remove

pallets/admin-utils/tests/tests.rs
```diff
@@ -1183,6 +1183,22 @@ fn test_sudo_set_target_stakes_per_interval() {
     });
 }
 
+#[test]
+fn test_set_total_subnet_locked_ok() {
```
distributedstatemachine: Remove

pallets/admin-utils/tests/tests.rs
```diff
@@ -1219,6 +1235,21 @@ fn test_set_alpha_values_dispatch_info_ok() {
     });
 }
 
+#[test]
+fn test_set_total_subnet_locked_not_sudo() {
+    new_test_ext().execute_with(|| {
```
distributedstatemachine: Remove all tests

distributedstatemachine (changes requested):
Can we also add tests for the migrations ?

runtime/src/migrations/initialise_ti.rs
orriin: They should be in the runtime actually, it allows pre/post checks and better integration with try-runtime-cli.

orriin:
DO NOT MERGE UNTIL CONST HAS APPROVED

runtime/src/migrations/initialise_ti.rs
orriin: I moved the migration definition to the pallet.

It's best practice for the migrations to be passed via the runtime so that try-runtime-cli can run pre/post hooks, do addition checks, and give better error messages.

PR #604, opened 2024-07-01
Make weighted_median iterative

keithtensor:
Devnet companion: #628 

This PR removes the recursive algorithm in `weighted_median` and replaces it with a iterative one.


gztensor (approved):
Existing unit tests cause a lot of recursion on both branches of if statement in the older version. I am confident this fix doesn't break it.

sam0x17 (approved):
looks good to me!

JohnReedV (approved):

PR #620, opened 2024-07-07
Schedule Coldkey Swap

distributedstatemachine:
## Description

This PR introduces a new extrinsic called [schedule_coldkey_swap](https://github.com/opentensor/subtensor/blob/arbitrage_coldkeys/pallets/subtensor/src/lib.rs#L2082) that allows users to swap their potentially compromised coldkeys while implementing a time-based arbitration mechanism for resolving multiple swap requests.

## Motivation for change

The primary motivation for this change is to address a recent security incident where coldkeys were potentially compromised. This feature enhances the security and flexibility of the Subtensor network by providing users with a mechanism to replace compromised coldkeys. It's crucial for maintaining the integrity of user accounts and protecting their assets in case of security breaches or key compromises.

## Behaviour pre-change

Before this change, users had no built-in mechanism to swap their coldkeys if they were compromised. This lack of functionality could lead to potential security risks and loss of assets if a coldkey was compromised without any means of replacement. In light of the recent security incident, this limitation posed a significant risk to affected users.

## Behaviour post-change

After implementing this feature:

1. Users can initiate a coldkey swap by calling the [schedule_coldkey_swap](https://github.com/opentensor/subtensor/blob/arbitrage_coldkeys/pallets/subtensor/src/lib.rs#L2082) function, providing the new coldkey and a proof of work.

2. The system implements an arbitration period to handle multiple swap requests:
   - If it's the first swap request for a coldkey, it's scheduled for the next arbitration period.
   - If there are already two destination coldkeys for the old coldkey, the request is rejected.

3. The difficulty of the proof of work increases exponentially with each subsequent swap attempt, adding an extra layer of security.

4. After the arbitration period, the system either performs the swap if there's only one destination coldkey or extends the arbitration period if there are multiple requests.

5. The swap process includes transferring all associated data and balances from the old coldkey to the new one, including stakes, owned hotkeys, and subnet ownership.

This new functionality allows users affected by the recent security incident to secure their accounts and assets by replacing potentially compromised coldkeys.

## Alternatives considered

1. Immediate swaps without arbitration: This was deemed too risky as it could lead to potential abuse or unauthorized swaps, especially in the context of the recent security incident.

2. Manual review process: While more secure, this would be less efficient and require more administrative overhead, which could delay the resolution for affected users.

## TODO:

- [ ] Fix broken tests
- [ ] Ensure ColdkeyDestinationMap isnt emptied

## Related Issue(s)

- Closes #[issue number]

## Type of Change

- [ ] Documentation update

## Breaking Change

## Checklist

## Additional Notes


pallets/subtensor/src/benchmarks.rs
```diff
@@ -428,4 +428,29 @@ reveal_weights {
     let _ = Subtensor::<T>::commit_weights(<T as frame_system::Config>::RuntimeOrigin::from(RawOrigin::Signed(hotkey.clone())), netuid, commit_hash);
 
   }: reveal_weights(RawOrigin::Signed(hotkey.clone()), netuid, uids, weight_values, salt, version_key)
+
+  schedule_coldkey_swap {
```
distributedstatemachine, line 432: Need to update

pallets/subtensor/src/lib.rs
```diff
@@ -1301,21 +1334,29 @@ pub mod pallet {
         // 	* 'n': (BlockNumberFor<T>):
         // 		- The number of the block we are initializing.
         fn on_initialize(_block_number: BlockNumberFor<T>) -> Weight {
+            // Unstake all and transfer pending coldkeys
+            let swap_weight = match Self::swap_coldkeys_this_block() {
```
keithtensor: We do need to be careful about how long this operation will take on average, otherwise it may brick the chain.

pallets/subtensor/src/swap.rs
```diff
@@ -115,47 +122,287 @@ impl<T: Config> Pallet<T> {
...
+    /// * `Weight` - The total weight consumed by the operation.
+    pub fn swap_coldkeys_this_block() -> Result<Weight, &'static str> {
+        let mut weight = frame_support::weights::Weight::from_parts(0, 0);
+
+        // Get the block number
+        let current_block: u64 = Self::get_current_block_as_u64();
+        log::debug!("Swapping coldkeys for block: {:?}", current_block);
+
+        // Get the coldkeys to swap here and then remove them.
+        let source_coldkeys: Vec<T::AccountId> = ColdkeysToSwapAtBlock::<T>::get(current_block);
```
keithtensor, line 348: What we could do is check whether `source_coldkeys` is greater than `MAX_ALLOWED_COLDKEYS_TO_SWAP_PER_BLOCK`, and if it is, then we truncate this vector, putting the remaining coldkeys back into the `ColdKeysToSwapAtBlock`, albeit in `current_block + 1`.

orriin:
> If there are already two destination coldkeys for the old coldkey, the request is rejected.

Does this mean an attacker could register 2 coldkeys immediately, preventing the true owner from registering a claim?

pallets/subtensor/src/errors.rs
```diff
@@ -146,5 +146,13 @@ mod errors {
         NoBalanceToTransfer,
         /// Same coldkey
         SameColdkey,
+        /// The coldkey is in arbitration
+        ColdkeyIsInArbitration,
+        /// The new coldkey is already registered for the drain
+        DuplicateColdkey,
+        /// Error thrown on a coldkey swap.
+        SwapError,
```
orriin: Can we make swap errors more specific? Otherwise may be difficult to pinpoint exactly what went wrong

pallets/subtensor/src/staking.rs
```diff
@@ -851,113 +869,74 @@ impl<T: Config> Pallet<T> {
...
-        // Emit the event
-        Self::deposit_event(Event::AllBalanceUnstakedAndTransferredToNewColdkey {
-            current_coldkey: current_coldkey.clone(),
-            new_coldkey: new_coldkey.clone(),
-            total_balance,
-        });
-
-        Ok(())
-    }
+    // pub fn x( coldkey_a: &T::AccountId, coldkey_b: &T::AccountId ) -> Weight {
```
orriin: Why is all this stuff commented out?

pallets/subtensor/src/swap.rs
```diff
@@ -115,47 +122,287 @@ impl<T: Config> Pallet<T> {
...
+        if total_balance > 0 {
+            // Attempt to transfer the entire total balance to coldkeyB.
+            if let Err(e) = T::Currency::transfer(
+                old_coldkey,
+                new_coldkey,
+                total_balance,
+                Preservation::Expendable,
+            ) {
+                return Err(e.into());
+            }
```
orriin: Can just use ? syntax I think

pallets/subtensor/tests/swap.rs
```diff
@@ -1380,3 +1386,25 @@ fn test_coldkey_has_associated_hotkeys() {
         ));
     });
 }
+
+// #[test]
```
orriin, line 1669: Is this intentionally commented out?

runtime/src/lib.rs
```diff
@@ -312,26 +312,9 @@ impl Contains<RuntimeCall> for SafeModeWhitelistedCalls {
                 | RuntimeCall::SafeMode(_)
                 | RuntimeCall::Timestamp(_)
                 | RuntimeCall::SubtensorModule(
-                    pallet_subtensor::Call::add_stake { .. }
```
orriin, line 315: Are these intentionally removed?

pallets/subtensor/src/lib.rs
sam0x17: 33k IIRC

pallets/subtensor/src/swap.rs
```diff
@@ -115,47 +122,287 @@ impl<T: Config> Pallet<T> {
...
+    pub fn swap_coldkeys_this_block() -> Result<Weight, &'static str> {
+        let mut weight = frame_support::weights::Weight::from_parts(0, 0);
+
+        // Get the block number
+        let current_block: u64 = Self::get_current_block_as_u64();
+        log::debug!("Swapping coldkeys for block: {:?}", current_block);
+
+        // Get the coldkeys to swap here and then remove them.
+        let source_coldkeys: Vec<T::AccountId> = ColdkeysToSwapAtBlock::<T>::get(current_block);
+        ColdkeysToSwapAtBlock::<T>::remove(current_block);
```
gztensor, line 349: We should check if `source_coldkeys` length is not zero before performing this `remove` because it adds an unnecessary read on every block otherwise.

pallets/subtensor/src/lib.rs
distributedstatemachine: We require caller to submit proofs of work which grow expotentially with each call. In addition, the calling coldkey has to have 1 TAO , which makes DDOS this function very expensive

sam0x17:
> > If there are already two destination coldkeys for the old coldkey, the request is rejected.
> 
> Does this mean an attacker could register 2 coldkeys immediately, preventing the true owner from registering a claim?

yeah this one concerns me, I am still partial to my original idea which is you can schedule your key to be swapped, the swap takes X days, and there is a sudo extrinsic which can cancel any particular swap that only we can submit, so once a swap starts either you authorized it, in which case you are good, or you have X days to notice that an attacker did it on your behalf and report that to us. This solution is a bit more centralized but it:
1) reduces the work on our end down to just cases where a swap is contested, which with any luck could be 0 anyway
2) gives us full control over which swaps get cancelled (the cancel extrinsic would require sudo) and gives and the user a long window (X days) in which to validate contested swap requests
3) the attacker's only chance at this point becomes the user not noticing that the attacker has started a swap request before the user
4) the medium for contesting would be opening a thread in discord, so we wouldn't have to worry about on-chain resources being overwhelmed by many contesting requests, and it would be quite easy to have the community help vouch for people, versus doing this on-chain

pallets/subtensor/src/errors.rs
distributedstatemachine: done

orriin:
TODO for me: swap `on_initialize` hook for `on_poll`

pallets/subtensor/tests/swap.rs
open-junius: uncommented, the test case works

pallets/subtensor/src/lib.rs
```diff
@@ -83,6 +84,9 @@ pub mod pallet {
     /// order of migrations. (i.e. always increasing)
     const STORAGE_VERSION: StorageVersion = StorageVersion::new(6);
 
+    /// Minimum balance required to perform a coldkey swap
+    pub const MIN_BALANCE_TO_PERFORM_COLDKEY_SWAP: u64 = 1_000_000_000; // 1 TAO in RAO
```
sam0x17: I get that this will prevent spam from arbitrary keys, but what will we do if/when someone reaches out to us who has less than 1 TAO but wants to upgrade their key? Just want to make sure we have thought through that edge case. If our plan is to tell them they need 1 TAO to upgrade their key and if they have less they are out of luck, we should definitely be clear that is our stance in the announcement

pallets/subtensor/src/swap.rs
```diff
@@ -112,50 +119,308 @@ impl<T: Config> Pallet<T> {
...
+        weight_used = weight_used.saturating_add(T::DbWeight::get().reads_writes(1, 1));
+
+        let mut keys_swapped = 0u64;
+        for coldkey_i in source_coldkeys.iter() {
+            // TODO: need a sane way to terminate early without locking users in.
+            // we should update the swap time
+            // if weight_used.ref_time() > weight_limit.ref_time() {
+            //     log::warn!("Could not finish swapping all coldkeys this block due to weight limit, breaking after swapping {} keys.", keys_swapped);
+            //     break;
+            // }
```
sam0x17, line 359: ideally we should benchmark a single swap and use the resulting weight as our means for estimating the weight of n swaps. I think breaking early makes sense but we have to be certain we aren't underestimating weight, so perhaps we double it

gztensor (changes requested):
It's a small nit, but if tempo is 0, it means we have 7.2 emissions per day, not 0, because of how `blocks_until_next_epoch` function is defined. The code in `get_delegate_by_existing_account` should be:

```rust
if tempo > U64F64::from_num(0) {
    let epochs_per_day: U64F64 = U64F64::from_num(7200).saturating_div(tempo);
    emissions_per_day =
        emissions_per_day.saturating_add(emission.saturating_mul(epochs_per_day));
} else {
    emissions_per_day = emissions_per_day.saturating_add(U64F64.from_num(7.2))
}
```

pallets/subtensor/src/swap.rs
distributedstatemachine: done

distributedstatemachine:
> It's a small nit, but if tempo is 0, it means we have 7.2 emissions per day, not 0, because of how `blocks_until_next_epoch` function is defined. The code in `get_delegate_by_existing_account` should be:
> 
> ```rust
> if tempo > U64F64::from_num(0) {
>     let epochs_per_day: U64F64 = U64F64::from_num(7200).saturating_div(tempo);
>     emissions_per_day =
>         emissions_per_day.saturating_add(emission.saturating_mul(epochs_per_day));
> } else {
>     emissions_per_day = emissions_per_day.saturating_add(U64F64.from_num(7.2))
> }
> ```

Really? IIRC , if the tempo is zero , the epoch doesnt run @gztensor

distributedstatemachine:
@gztensor 

```
    pub fn blocks_until_next_epoch(netuid: u16, tempo: u16, block_number: u64) -> u64 {
        // tempo | netuid | # first epoch block
        //   1        0               0
        //   1        1               1
        //   2        0               1
        //   2        1               0
        //   100      0              99
        //   100      1              98
        // Special case: tempo = 0, the network never runs.
        if tempo == 0 {
            return 1000;
        }
        (tempo as u64).saturating_sub(
            block_number.saturating_add(netuid as u64).saturating_add(1)
                % (tempo as u64).saturating_add(1),
        )
    }
   ```

pallets/subtensor/src/lib.rs
distributedstatemachine: This is a great point , and we have addressed this partially in the schedule_coldkey swap method , and it now includes a check for delegates with delegated stakes > 500 , and subnet owners not requiring a minimum tao balance to call it .

pallets/subtensor/src/staking.rs
distributedstatemachine: Old cold , removing

pallets/subtensor/src/swap.rs
distributedstatemachine: Not sure how to deal with this at the moment , that doesnt involve rewriting the codebase , but I am open to suggestions , as for now , it is a risk we accept

pallets/subtensor/src/lib.rs
keithtensor: Wait, how does the caller factor in here? This `swap_coldkeys_this_block` function is being run in `on_initialize`, which doesn't have a caller and gets run on every block.

pallets/subtensor/src/swap.rs
distributedstatemachine: By virtue of the transaction gettting to this fucntion , I dont think its possible for ColdkeysToSwapAtBlock to be 0..

pallets/subtensor/src/lib.rs
distributedstatemachine: in the previous call , `schedule_coldkey_swap` , it registers a arbitration block number, and then this call with swap those keys once they reach the arbitration block number.

keithtensor (approved):

gztensor (approved):

JohnReedV (approved):

camfairchild (approved):

runtime/src/lib.rs
distributedstatemachine: yes

PR #671, opened 2024-07-23
Added `--no-purge` flag to `scripts/localnet.sh` to be able to save state when restarting a script.

basfroman:
When working with localnet, there is often a need to restart the script. To save the state and not waste time on settings, you need to comment out some lines. I suggest introducing a flag that will bypass local state cleanup.

### DIFF
--- a/scripts/localnet.sh
+++ b/scripts/localnet.sh
@@ -1,5 +1,14 @@
 #!/bin/bash
 
+# Check if `--no-purge` passed as a parameter
+NO_PURGE=0
+for arg in "$@"; do
+    if [ "$arg" = "--no-purge" ]; then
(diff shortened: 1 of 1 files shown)


sam0x17:
Hey roman just an FYI we no longer use the `development` branch. All PRs should be into `main`. See https://github.com/opentensor/subtensor/blob/main/CONTRIBUTING.md for more info

basfroman:
> Hey roman just an FYI we no longer use the `development` branch. All PRs should be into `main`. See https://github.com/opentensor/subtensor/blob/main/CONTRIBUTING.md for more info

Hey Sam, I see. Thanks for the update.

PR #675, opened 2024-07-23
Added `--no-purge` flag to `scripts/localnet.sh` to be able to save state when restarting a script.

basfroman:
When working with localnet, there is often a need to restart the script. To save the state and not waste time on settings, you need to comment out some lines. I suggest introducing a flag that will bypass local state cleanup.


unconst (dismissed):

runtime/src/lib.rs
```diff
@@ -139,7 +139,7 @@ pub const VERSION: RuntimeVersion = RuntimeVersion {
     //   `spec_version`, and `authoring_version` are the same between Wasm and native.
     // This value is set to 100 to notify Polkadot-JS App (https://polkadot.js.org/apps) to use
     //   the compatible custom types.
-    spec_version: 191,
+    spec_version: 192,
```
open-junius, line 142: curious if the code change just to pass CI. no need to update runtime spec version.

basfroman: If I don't do this then one checker failed. Also this action is mentioned in the [contribution instruction](https://github.com/opentensor/subtensor/blob/main/CONTRIBUTING.md).

JohnReedV (approved):

sam0x17:
this PR should be into main as per https://github.com/opentensor/subtensor/blob/main/CONTRIBUTING.md

PR #699, opened 2024-07-30
Feat/childkey take

distributedstatemachine:
## Description

This PR add childkey takes , and more tests for childkeys


rajkaramchedu:
When this lands, we should add a test case to the Google doc we went over, @gus-opentensor.

pallets/subtensor/src/lib.rs
```diff
@@ -603,7 +625,12 @@ pub mod pallet {
     #[pallet::storage]
     /// MAP ( hot ) --> take | Returns the hotkey delegation take. And signals that this key is open for delegation.
     pub type Delegates<T: Config> =
-        StorageMap<_, Blake2_128Concat, T::AccountId, u16, ValueQuery, DefaultDefaultTake<T>>;
+        StorageMap<_, Blake2_128Concat, T::AccountId, u16, ValueQuery, DefaultDelegateTake<T>>;
+    #[pallet::storage]
+    /// DMAP ( hot, netuid ) --> take | Returns the hotkey childkey take for a specific subnet
+    pub type ChildkeyTake<T: Config> =
+        StorageMap<_, Blake2_128Concat, (T::AccountId, u16), u16, ValueQuery>;
```
gztensor: Does it need to be single map with tuple key instead of double map?

pallets/subtensor/src/staking/set_children.rs
```diff
@@ -190,4 +190,102 @@ impl<T: Config> Pallet<T> {
...
+        );
+
+        // Ensure the hotkey passes the rate limit
+        ensure!(
+            Self::passes_rate_limit_on_subnet(&TransactionType::SetChildkeyTake, &hotkey, netuid),
+            Error::<T>::TxChildkeyTakeRateLimitExceeded
+        );
+
+        // Set the new childkey take value for the given hotkey and network
+        ChildkeyTake::<T>::insert((hotkey.clone(), netuid), take);
```
gztensor: We also need to remove this value when we remove children for a parent key to avoid chain pollution.

runtime/src/lib.rs
```diff
@@ -858,7 +858,9 @@ parameter_types! {
     pub const SubtensorInitialPruningScore : u16 = u16::MAX;
     pub const SubtensorInitialBondsMovingAverage: u64 = 900_000;
     pub const SubtensorInitialDefaultTake: u16 = 11_796; // 18% honest number.
-    pub const SubtensorInitialMinTake: u16 = 5_898; // 9%
+    pub const SubtensorInitialMinDelegateTake: u16 = 5_898; // 9%
+    pub const SubtensorInitialDefaultChildKeyTake: u16 = 11_796; // 18% honest number.
```
gztensor: I may remember this wrong from a voice conversation, but didn't we decide to make the default to be zero so that we don't deincentivise parent key delegation?

pallets/subtensor/src/lib.rs
distributedstatemachine: I think it might be better off as a double map to allow greater flexibility when querying it.

runtime/src/lib.rs
distributedstatemachine: done

pallets/admin-utils/tests/mock.rs
```diff
@@ -77,12 +77,16 @@ parameter_types! {
     pub const InitialBondsMovingAverage: u64 = 900_000;
     pub const InitialStakePruningMin: u16 = 0;
     pub const InitialFoundationDistribution: u64 = 0;
-    pub const InitialDefaultTake: u16 = 11_796; // 18% honest number.
+    pub const InitialDefaultDelegateTake: u16 = 11_796; // 18% honest number.
+    pub const InitialMinDelegateTake: u16 = 5_898; // 9%;
+    pub const InitialDefaultChildKeyTake: u16 = 11_796; // 18% honest number.
```
gztensor: We probably want to use the same default value as in Runtime for default childkey take.

gztensor (changes requested):
We should be removing ChildkeyTake entries for old children in do_set_children in this loop (only if old children are not included in the new set of children):

```rust
// --- 6.0. Iterate over all my old children and remove myself from their parent's map.
        for (_, old_child_i) in old_children.clone().iter() {
            // --- 6.1. Get the old child's parents on this network.
            let my_old_child_parents: Vec<(u64, T::AccountId)> =
                ParentKeys::<T>::get(old_child_i.clone(), netuid);
            // --- 6.2. Filter my hotkey from my old children's parents list.
            let filtered_parents: Vec<(u64, T::AccountId)> = my_old_child_parents
                .into_iter()
                .filter(|(_, parent)| *parent != hotkey)
                .collect();
            // --- 6.3. Update the parent list in storage
            ParentKeys::<T>::insert(old_child_i, netuid, filtered_parents);
        }
```

pallets/subtensor/src/staking/set_children.rs
```diff
@@ -130,6 +130,11 @@ impl<T: Config> Pallet<T> {
         // --- 7.1. Insert my new children + proportion list into the map.
         ChildKeys::<T>::insert(hotkey.clone(), netuid, children.clone());
 
+        if children.is_empty() {
+            // If there are no children, remove the ChildkeyTake value
+            ChildkeyTake::<T>::remove(hotkey.clone(), netuid);
```
gztensor: We should not remove `hotkey` childket take here: `hotkey` is a parent. If this parent is someone's child, even if it has no children, it should keep childkey take if it exists.

distributedstatemachine: On second thoughts , I dont think we should be clearing these values out regardless , as everyone has the ability to set a childkey take , regardless of whether or not a parent is allocating to them.

gztensor (dismissed):

pallets/subtensor/tests/children.rs
```diff
@@ -32,6 +34,7 @@ fn test_do_set_child_singular_success() {
     });
 }
 
+// 2: Attempt to set child in non-existent network
 // SKIP_WASM_BUILD=1 RUST_LOG=info cargo test --test children -- test_do_set_child_singular_network_does_not_exist --exact --nocapture
 #[test]
 fn test_do_set_child_singular_network_does_not_exist() {
```
open-junius, line 40: It is duplicated with test_do_revoke_child_singular_network_does_not_exist, test_do_set_children_multiple_network_does_not_exist, test_do_revoke_children_multiple_network_does_not_exist. just test the network not exists.

pallets/subtensor/tests/children.rs
```diff
@@ -862,6 +1220,10 @@ fn test_do_revoke_children_multiple_success() {
     });
 }
 
+// 29: Test revoking children when network does not exist
+// This test verifies the behavior when attempting to revoke children on a non-existent network:
+// - Attempts to revoke children on a network that doesn't exist
+// - Verifies that the operation fails with the correct error
 // SKIP_WASM_BUILD=1 RUST_LOG=info cargo test --test children -- test_do_revoke_children_multiple_network_does_not_exist --exact --nocapture
 #[test]
 fn test_do_revoke_children_multiple_network_does_not_exist() {
```
open-junius, line 1229: The test code not aligned with test name.

pallets/subtensor/tests/children.rs
```diff
@@ -560,6 +612,10 @@ fn test_do_set_children_multiple_success() {
     });
 }
 
+// 17: Test setting multiple children in a non-existent network
+// This test ensures that attempting to set multiple children in a non-existent network results in an error:
+// - Attempts to set children in a network that doesn't exist
+// - Verifies that the appropriate error is returned
 // SKIP_WASM_BUILD=1 RUST_LOG=info cargo test --test children -- test_do_set_children_multiple_network_does_not_exist --exact --nocapture
 #[test]
 fn test_do_set_children_multiple_network_does_not_exist() {
```
open-junius, line 621: The test code not aligned with test name. just one child and the number of child has no difference on a non-existed network.

pallets/subtensor/tests/children.rs
```diff
@@ -453,6 +485,10 @@ fn test_do_revoke_child_singular_success() {
     });
 }
 
+// 13: Test revoking a child in a non-existent network
+// This test verifies that attempting to revoke a child in a non-existent network results in an error:
+// - Attempts to revoke a child in a network that doesn't exist
+// - Checks that the appropriate error is returned
 // SKIP_WASM_BUILD=1 RUST_LOG=info cargo test --test children -- test_do_revoke_child_singular_network_does_not_exist --exact --nocapture
 #[test]
 fn test_do_revoke_child_singular_network_does_not_exist() {
```
open-junius, line 494: can't see any child revoke.

pallets/subtensor/tests/children.rs
```diff
@@ -583,6 +639,11 @@ fn test_do_set_children_multiple_network_does_not_exist() {
...
 }
 
+// 18: Test setting multiple children with an invalid child
+// This test verifies that attempting to set multiple children with an invalid child (same as parent) results in an error:
+// - Sets up a network and registers a hotkey
+// - Attempts to set a child that is the same as the parent hotkey
+// - Checks that the appropriate error is returned
 // SKIP_WASM_BUILD=1 RUST_LOG=info cargo test --test children -- test_do_set_children_multiple_invalid_child --exact --nocapture
 #[test]
 fn test_do_set_children_multiple_invalid_child() {
```
open-junius, line 649: just one child in the extrinsic.

pallets/subtensor/tests/children.rs
distributedstatemachine: Setting an empty vec for children is the revoke operation.

unconst (dismissed):

unconst (approved):

Issue #578, opened 2024-06-25
crates.io releases


sam0x17:
If we start releasing the `subtensor` workspace to crates.io ~~it would dramatically simplify the user experience of running a node (could even get to `cargo install subtensor` level potentially), but I see this as a side-mission / goal, the real objective here would~~ be getting rust docs released to docs.rs/lib.rs/etc so when people google subtensor-related things they actually come up with up-to-date docs.

AC:
- [x] consolidate workspace so it is crates.io-friendly, ensure globally unique crate names, etc
- [ ] set up a secure crates.io account we can use for releases, get a key for pushing to it added to CI
- [x] publish to crates.io (requires a github action / script especially since crates need to be published in a very particular order) #720
- [ ] ability to publish to crates.io automatically as part of gh action release process
- [x] ~~ability to spin up a new node via `cargo install`? or something we don't care about?~~

ppolewicz:
Running a node is already easy via docker-compose and crates.io won't make it any easier for ordinary user.

I don't understand what benefit it would give to non-developers who need to have devenv set up already.

sam0x17:
> Running a node is already easy via docker-compose and crates.io won't make it any easier for ordinary user.
> 
> I don't understand what benefit it would give to non-developers who need to have devenv set up already.

probably none, I just mentioned it in case anyone thought it would be useful for some specific reason

gztensor:
On-hold until reproducible builds are done.

keithtensor:
It looks like we're actually quite close in completing this, we only seem to need a specific crates.io account and a GH action release process. Am I missing anything?

sam0x17:
major blocker is we need to 100% eliminate git dependencies from our workspace. All crates.io crates have to only have crates.io dependencies

PR #720, opened 2024-08-06
Create publish script to publish crates in the correct order

keithtensor:
Work towards #578.


support/tools/src/bump_version.rs
```diff
@@ -0,0 +1,49 @@
...
+    "pallets/registry",
+    "pallets/subtensor",
+    "pallets/subtensor/runtime-api",
+    "pallets/admin-utils",
+    "runtime",
+    "node",
+];
+
+fn main() -> anyhow::Result<()> {
+    let mut version_file = fs::File::options().read(true).write(true).open("VERSION")?;
```
sam0x17: would be better if we could read this from STDIN then we don't even need a version file you can just specify it as an input to the program and use a workflow input on the github action that lets you specify the tag

distributedstatemachine (approved):

unconst (approved):

PR #724, opened 2024-08-08
custom linting framework + freeze struct enforcement lint

### LINKED ISSUE #606: freeze_struct improvements / custom linting framework
#513 introduced the `#[freeze_struct]` attribute which allows you to "lock" a storage struct such that any changes to it result in a new hash code that is enforced at compile time.

A few additional features would enhance this:
- [ ] ignore changes to generic name in hash code calculation (verify that this is actually harmless)
- [ ] ignore changes to field names in hash code calculation (verify that this is actually harmless)
- [x] custom linting framework
- [x] an automatic lint requiring this on storage structs

sam0x17:
custom linting framework + custom lint that forces us to add `#[freeze_struct]` to ALL structs in the workspace that have `Encode` and/or `Decode`. Re-adds `#[freeze_struct]` to a number of structs that were missing it including 4-5 that previously had it but it got removed

Fixes #606

Screenshot 2024-08-08 at 12 41 05 AM


sam0x17:
closing since swap insertion isn't really an issue anymore 🎉 


will pull out some of the changes here since there were some unrelated improvements

sam0x17:
whoops

scripts/localnet.sh
```diff
@@ -48,26 +48,26 @@ fi
 
 if [[ $BUILD_BINARY == "1" ]]; then
   echo "*** Building substrate binary..."
-  cargo build --release --features "$FEATURES" --manifest-path "$BASE_DIR/Cargo.toml"
+  cargo build --workspace --profile=production --features "$FEATURES" --manifest-path "$BASE_DIR/Cargo.toml"
```
distributedstatemachine: We do not want these for local net builds , as it just makes it alot longer. I think the original release build is sufficient

support/linting/src/dummy_lint.rs
```diff
@@ -0,0 +1,11 @@
+use syn::File;
+
+use super::*;
+
+pub struct DummyLint;
+
+impl Lint for DummyLint {
```
distributedstatemachine: Why do we need DummyLint?

distributedstatemachine (changes requested):

support/linting/src/dummy_lint.rs
sam0x17: we can probably remove it now good call out lol

distributedstatemachine (approved):

unconst (approved):

PR #750, opened 2024-08-24
add neuron certificate handling

andreea-popescu-reef:
## Description

implemented logic for neuron certificate discovery and storage:
- add `neuronInfo_getNeuronCertificate: ( netuid, hotkey ) --> certificate` pallet storage for neuron certificates
- add `fn get_neuron_certificate(netuid, neuron_uid)` rpc to be called via bittensor library to fetch the certificate for a particular neuron
- add `fn serve_axon_tls(..., certificate)` as an alternative for existing `serve_axon` taking an additional certificate parameter and storing it

compatible with older versions of bittensor as it doesn't modify the behaviour of existing rpc calls.

## Checklist
- [ ] Added tests for intended behaviors
- [ ] Added tests for error and/or panic states
- [ ] Updated relevant documentation
- [ ] Tracked `rustfmt` changes
- [ ] No _new_ compiler warnings

### DIFF
--- a/pallets/subtensor/rpc/src/lib.rs
+++ b/pallets/subtensor/rpc/src/lib.rs
@@ -41,13 +41,6 @@ pub trait SubtensorCustomApi<BlockHash> {
     fn get_neurons(&self, netuid: u16, at: Option<BlockHash>) -> RpcResult<Vec<u8>>;
     #[method(name = "neuronInfo_getNeuron")]
     fn get_neuron(&self, netuid: u16, uid: u16, at: Option<BlockHash>) -> RpcResult<Vec<u8>>;
-    #[method(name = "neuronInfo_getNeuronCertificate")]
-    fn get_neuron_certificate(
-        &self,
-        netuid: u16,
-        uid: u16,
-        at: Option<BlockHash>,
-    ) -> RpcResult<Vec<u8>>;
     #[method(name = "subnetInfo_getSubnetInfo")]
     fn get_subnet_info(&self, netuid: u16, at: Option<BlockHash>) -> RpcResult<Vec<u8>>;
     #[method(name = "subnetInfo_getSubnetsInfo")]
@@ -191,20 +184,6 @@ where
 
         api.get_neuron(at, netuid, uid)
             .map_err(|e| Error::RuntimeError(format!("Unable to get neuron info: {:?}", e)).into())
-    }
-
-    fn get_neuron_certificate(
-        &self,
-        netuid: u16,
-        uid: u16,
-        at: Option<<Block as BlockT>::Hash>,
-    ) -> RpcResult<Vec<u8>> {
-        let api = self.client.runtime_api();
-        let at = at.unwrap_or_else(|| self.client.info().best_hash);
-
-        api.get_neuron_certificate(at, netuid, uid).map_err(|e| {
-            Error::RuntimeError(format!("Unable to get neuron certificate: {:?}", e)).into()
-        })
     }
 
     fn get_subnet_info(
--- a/pallets/subtensor/rpc/src/lib.rs
+++ b/pallets/subtensor/rpc/src/lib.rs
@@ -41,7 +41,13 @@ pub trait SubtensorCustomApi<BlockHash> {
     fn get_neurons(&self, netuid: u16, at: Option<BlockHash>) -> RpcResult<Vec<u8>>;
     #[method(name = "neuronInfo_getNeuron")]
     fn get_neuron(&self, netuid: u16, uid: u16, at: Option<BlockHash>) -> RpcResult<Vec<u8>>;
-
+    #[method(name = "neuronInfo_getNeuronCertificate")]
+    fn get_neuron_certificate(
+        &self,
+        netuid: u16,
+        uid: u16,
+        at: Option<BlockHash>,
+    ) -> RpcResult<Vec<u8>>;
     #[method(name = "subnetInfo_getSubnetInfo")]
     fn get_subnet_info(&self, netuid: u16, at: Option<BlockHash>) -> RpcResult<Vec<u8>>;
     #[method(name = "subnetInfo_getSubnetsInfo")]
@@ -185,6 +191,20 @@ where
 
         api.get_neuron(at, netuid, uid)
             .map_err(|e| Error::RuntimeError(format!("Unable to get neuron info: {:?}", e)).into())
+    }
+
+    fn get_neuron_certificate(
+        &self,
+        netuid: u16,
+        uid: u16,
+        at: Option<<Block as BlockT>::Hash>,
+    ) -> RpcResult<Vec<u8>> {
+        let api = self.client.runtime_api();
+        let at = at.unwrap_or_else(|| self.client.info().best_hash);
+
+        api.get_neuron_certificate(at, netuid, uid).map_err(|e| {
+            Error::RuntimeError(format!("Unable to get neuron certificate: {:?}", e)).into()
+        })
     }
 
     fn get_subnet_info(
--- a/pallets/subtensor/src/lib.rs
+++ b/pallets/subtensor/src/lib.rs
@@ -73,6 +73,7 @@ pub mod pallet {
     use sp_runtime::traits::{Dispatchable, TrailingZeroInput};
     use sp_std::vec;
     use sp_std::vec::Vec;
+	use subtensor_macros::freeze_struct;
 
     #[cfg(not(feature = "std"))]
     use alloc::boxed::Box;
@@ -132,6 +133,7 @@ pub mod pallet {
     /// Struct for NeuronCertificate.
     pub type NeuronCertificateOf = NeuronCertificate;
     /// Data structure for NeuronCertificate information.
+    #[freeze_struct("e6193a76002d491")]
     #[derive(Decode, Encode, Default, TypeInfo, PartialEq, Eq, Clone, Debug)]
     pub struct NeuronCertificate {
         ///  The neuron certificate.
--- a/pallets/subtensor/src/lib.rs
+++ b/pallets/subtensor/src/lib.rs
@@ -129,6 +129,15 @@ pub mod pallet {
         pub placeholder2: u8,
     }
 
+    /// Struct for NeuronCertificate.
+    pub type NeuronCertificateOf = NeuronCertificate;
+    /// Data structure for NeuronCertificate information.
+    #[derive(Decode, Encode, Default, TypeInfo, PartialEq, Eq, Clone, Debug)]
+    pub struct NeuronCertificate {
+        ///  The neuron certificate.
+        pub certificate: Vec<u8>,
+    }
+
     ///  Struct for Prometheus.
     pub type PrometheusInfoOf = PrometheusInfo;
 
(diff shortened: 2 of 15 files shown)


pallets/subtensor/rpc/src/lib.rs
```diff
@@ -41,7 +41,13 @@ pub trait SubtensorCustomApi<BlockHash> {
     fn get_neurons(&self, netuid: u16, at: Option<BlockHash>) -> RpcResult<Vec<u8>>;
     #[method(name = "neuronInfo_getNeuron")]
     fn get_neuron(&self, netuid: u16, uid: u16, at: Option<BlockHash>) -> RpcResult<Vec<u8>>;
-
+    #[method(name = "neuronInfo_getNeuronCertificate")]
```
distributedstatemachine: We are deprecating the addition of new RPCs. going forwards ,  clients should make query state directly , for what they nee. That being said , since this is a part of the Axon , it can be fetched from neuron info which the cli already uses to populate the metagraph

pallets/subtensor/src/macros/dispatches.rs
```diff
@@ -507,6 +507,89 @@ mod dispatches {
...
+        /// 	- The numerically encoded ip address does not resolve to a proper ip.
+        ///
+        /// * 'ServingRateLimitExceeded':
+        /// 	- Attempting to set prometheus information withing the rate limit min.
+        ///
+        #[pallet::call_index(40)]
+        #[pallet::weight((Weight::from_parts(46_000_000, 0)
+		.saturating_add(T::DbWeight::get().reads(4))
+		.saturating_add(T::DbWeight::get().writes(1)), DispatchClass::Normal, Pays::No))]
+        pub fn serve_axon_tls(
```
distributedstatemachine, line 577: Do we need a new method for this ? Can we just have serve Axon ? with Certificate optional ?

distributedstatemachine (changes requested):

pallets/subtensor/src/macros/dispatches.rs
```diff
@@ -507,6 +507,89 @@ mod dispatches {
...
+            origin: OriginFor<T>,
+            netuid: u16,
+            version: u32,
+            ip: u128,
+            port: u16,
+            ip_type: u8,
+            protocol: u8,
+            placeholder1: u8,
+            placeholder2: u8,
+            certificate: Vec<u8>,
```
distributedstatemachine, line 587: Making this an option removes the need for `serve_axon_tls`

distributedstatemachine:
On seconds thoughts, adding the new extrinsic ensures backwards compatibility , so a bit undecided here

pallets/subtensor/src/macros/dispatches.rs
andreea-popescu-reef: wouldn't that break the bittensor lib once the subtensor is updated?

pallets/subtensor/rpc/src/lib.rs
andreea-popescu-reef: [It's not really part of the Axon](https://github.com/opentensor/subtensor/pull/750/files#diff-9d85ff26db441ec381d86188b389b226449eee4e9847bf0ce3301c0bb6cf93a0R99) - it's a separate storage. I thought that adding it to the aoxn  (can't use any the placeholders due to the type) would break the encoding/decoding of the AxonInfo on subtensor update - right?

distributedstatemachine: Got it , leaning towards seperate methods. Also we could introduce an Axonv2 , if we dont want to break existing integration. 

 The recommendation to remove rpcs still hold. Client should access this data via [direct state queries](https://gist.github.com/JohnReedV/de955830d499979e80da212ab03fa7b7)

andreea-popescu-reef: removed the rpc

pallets/subtensor/src/lib.rs
```diff
@@ -116,6 +115,15 @@ pub mod pallet {
         pub placeholder2: u8,
     }
 
+    /// Struct for NeuronCertificate.
+    pub type NeuronCertificateOf = NeuronCertificate;
+    /// Data structure for NeuronCertificate information.
+    #[derive(Decode, Encode, Default, TypeInfo, PartialEq, Eq, Clone, Debug)]
+    pub struct NeuronCertificate {
+        ///  The neuron certificate.
+        pub certificate: Vec<u8>,
```
gztensor: Can we make it a bounded vector instead? There should be a limit of how much data can an axon store on-chain.

pallets/subtensor/src/subnets/serving.rs
```diff
@@ -86,6 +90,15 @@ impl<T: Config> Pallet<T> {
             Error::<T>::ServingRateLimitExceeded
         );
 
+        // Check certificate
```
gztensor, line 93: We should be cleaning up stored certificates when the neuron is deregistered, there needs to be a matching deletion for this insertion.

gztensor (changes requested):

pallets/subtensor/src/lib.rs
open-junius: It is better to add freeze_struct, even for this simple struct.

pallets/subtensor/src/macros/dispatches.rs
open-junius: other solution is make the certificate as Option, and mark the serve_axon is deprecated. In the future, we just accept the extrinsic serve_axon_tls, and remove serve_axon.

open-junius: we can avoid two extrinsic with same function, and almost the same arguments.

pallets/subtensor/src/lib.rs
andreea-popescu-reef: will add freeze_struct

pallets/subtensor/src/rpc_info/neuron_info.rs
```diff
@@ -147,6 +147,23 @@ impl<T: Config> Pallet<T> {
         Some(neuron)
     }
 
+    pub fn get_neuron_certificate(netuid: u16, uid: u16) -> Option<NeuronCertificate> {
```
distributedstatemachine: if we dont have rpcs, do we still need this ?

andreea-popescu-reef: cleaned up

pallets/subtensor/src/lib.rs
distributedstatemachine: @andreea-popescu-reef what about the using a bounded vec ? Have you tested with actual certificate sizes to see how much data it is ?

andreea-popescu-reef: for an ECDSA public key we would need 256bits. To support multiple algorithms 512bits should be enough for the key plus some additional bits to determine the algoritthm so let's reserve 520bits.
I'd also rename the `certificate field` - `public_key` would be more appropiate
does that sound good?

distributedstatemachine: Its not about the naming , but more about people being able to dump arbitrary data (potentialy 5 MB worth)  into the field. This is why we are suggesting a bounded vec. It would also be nice to see some tests showing those bounds are enforced

andreea-popescu-reef: made the changes

distributedstatemachine (dismissed):

runtime/src/lib.rs
```diff
@@ -1369,13 +1369,7 @@ impl_runtime_apis! {
...
 
         fn get_delegate(delegate_account_vec: Vec<u8>) -> Vec<u8> {
-            let _result = SubtensorModule::get_delegate(delegate_account_vec);
-            if _result.is_some() {
-                let result = _result.expect("Could not get DelegateInfo");
-                result.encode()
-            } else {
-                vec![]
-            }
+            SubtensorModule::get_delegate(delegate_account_vec).map(|r| r.encode()).unwrap_or_default()
```
ppolewicz:
```suggestion
            let _result = SubtensorModule::get_delegate(delegate_account_vec);
            if _result.is_some() {
                let result = _result.expect("Could not get DelegateInfo");
                result.encode()
            } else {
                vec![]
            }
```

runtime/src/lib.rs
```diff
@@ -1391,13 +1385,7 @@ impl_runtime_apis! {
...
 
         fn get_neuron_lite(netuid: u16, uid: u16) -> Vec<u8> {
-            let _result = SubtensorModule::get_neuron_lite(netuid, uid);
-            if _result.is_some() {
-                let result = _result.expect("Could not get NeuronInfoLite");
-                result.encode()
-            } else {
-                vec![]
-            }
+            SubtensorModule::get_neuron_lite(netuid, uid).map(|r| r.encode()).unwrap_or_default()
```
ppolewicz:
```suggestion
            let _result = SubtensorModule::get_neuron_lite(netuid, uid);
            if _result.is_some() {
                let result = _result.expect("Could not get NeuronInfoLite");
                result.encode()
            } else {
                vec![]
            }
```

runtime/src/lib.rs
```diff
@@ -1406,25 +1394,13 @@ impl_runtime_apis! {
...
 
         fn get_neuron(netuid: u16, uid: u16) -> Vec<u8> {
-            let _result = SubtensorModule::get_neuron(netuid, uid);
-            if _result.is_some() {
-                let result = _result.expect("Could not get NeuronInfo");
-                result.encode()
-            } else {
-                vec![]
-            }
+            SubtensorModule::get_neuron(netuid, uid).map(|r| r.encode()).unwrap_or_default()
```
ppolewicz:
```suggestion
            let _result = SubtensorModule::get_neuron(netuid, uid);
            if _result.is_some() {
                let result = _result.expect("Could not get NeuronInfo");
                result.encode()
            } else {
                vec![]
            }
```

runtime/src/lib.rs
```diff
@@ -1406,25 +1394,13 @@ impl_runtime_apis! {
...
     impl subtensor_custom_rpc_runtime_api::SubnetInfoRuntimeApi<Block> for Runtime {
         fn get_subnet_info(netuid: u16) -> Vec<u8> {
-            let _result = SubtensorModule::get_subnet_info(netuid);
-            if _result.is_some() {
-                let result = _result.expect("Could not get SubnetInfo");
-                result.encode()
-            } else {
-                vec![]
-            }
+            SubtensorModule::get_subnet_info(netuid).map(|r| r.encode()).unwrap_or_default()
```
ppolewicz:
```suggestion
            let _result = SubtensorModule::get_subnet_info(netuid);
            if _result.is_some() {
                let result = _result.expect("Could not get SubnetInfo");
                result.encode()
            } else {
                vec![]
            }
```

runtime/src/lib.rs
```diff
@@ -1448,13 +1424,7 @@ impl_runtime_apis! {
...
 
         fn get_subnet_hyperparams(netuid: u16) -> Vec<u8> {
-            let _result = SubtensorModule::get_subnet_hyperparams(netuid);
-            if _result.is_some() {
-                let result = _result.expect("Could not get SubnetHyperparams");
-                result.encode()
-            } else {
-                vec![]
-            }
+            SubtensorModule::get_subnet_hyperparams(netuid).map(|r| r.encode()).unwrap_or_default()
```
ppolewicz:
```suggestion
            let _result = SubtensorModule::get_subnet_hyperparams(netuid);
            if _result.is_some() {
                let result = _result.expect("Could not get SubnetHyperparams");
                result.encode()
            } else {
                vec![]
            }
```

ppolewicz (commented):
Suggest revert refactor in an area of the code that doesn't have tests and we really, _really_ want to get this PR merged, so we'll come back to that refactor in a separate PR

PR #522, opened 2024-06-12
reset neuron data on registration

andreea-popescu-reef:
## Description

On neuron registration the trust, emission, consensus, incentive, dividends values as well as axon info associated with the assigned neuron_uid are inherited from the previous neuron. Thus for the first few blocks after registration these values will be wrong, then reset to zero and then get populated with the correct values for the newly registered neuron. This change resets these values on neuron registration.

## Related Issue(s)
N/A

## Type of Change

- [x] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] Documentation update
- [ ] Other (please describe):

## Breaking Change
no

## Checklist
- [x] I have performed a self-review of my own code
- [x] I have commented my code, particularly in hard-to-understand areas
- [x] I have run `cargo fmt` and `cargo clippy` to ensure my code is formatted and linted correctly
- [ ] I have made corresponding changes to the documentation
- [x] My changes generate no new warnings
- [x] I have added tests that prove my fix is effective or that my feature works
- [x] New and existing unit tests pass locally with my changes
- [ ] Any dependent changes have been merged and published in downstream modules

## Screenshots (if applicable)

N/A

## Additional Notes

N/A

PR #755, opened 2024-08-27
reset neuron data on dereg

andreea-popescu-reef:
reviving: https://github.com/opentensor/subtensor/pull/522

## Description

On neuron registration the trust, emission, consensus, incentive, dividends values as well as axon info associated with the assigned neuron_uid are inherited from the previous neuron. Thus for the first few blocks after registration these values will be wrong, then reset to zero and then get populated with the correct values for the newly registered neuron. This change resets these values on neuron registration.

### DIFF
--- a/pallets/subtensor/src/subnets/uids.rs
+++ b/pallets/subtensor/src/subnets/uids.rs
@@ -8,6 +8,23 @@ impl<T: Config> Pallet<T> {
         SubnetworkN::<T>::get(netuid)
     }
 
+    /// Sets value for the element at the given position if it exists.
+    pub fn set_element_at<N>(vec: &mut [N], position: usize, value: N) {
+        if let Some(element) = vec.get_mut(position) {
+            *element = value;
+        }
+    }
+
+    /// Resets the trust, emission, consensus, incentive, dividends of the neuron to default
+    pub fn clear_neuron(netuid: u16, neuron_uid: u16) {
+        let neuron_index: usize = neuron_uid.into();
+        Emission::<T>::mutate(netuid, |v| Self::set_element_at(v, neuron_index, 0));
+        Trust::<T>::mutate(netuid, |v| Self::set_element_at(v, neuron_index, 0));
+        Consensus::<T>::mutate(netuid, |v| Self::set_element_at(v, neuron_index, 0));
+        Incentive::<T>::mutate(netuid, |v| Self::set_element_at(v, neuron_index, 0));
+        Dividends::<T>::mutate(netuid, |v| Self::set_element_at(v, neuron_index, 0));
+    }
+
     /// Replace the neuron under this uid.
     pub fn replace_neuron(
         netuid: u16,
@@ -40,6 +57,12 @@ impl<T: Config> Pallet<T> {
 
         // 4. Clear neuron certificates
         NeuronCertificates::<T>::remove(netuid, old_hotkey.clone());
+
+        // 5. Reset new neuron's values.
+        Self::clear_neuron(netuid, uid_to_replace);
+
+        // 5a. reset axon info for the new uid.
+        Axons::<T>::remove(netuid, old_hotkey);
     }
 
     /// Appends the uid to the network.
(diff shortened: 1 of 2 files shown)


pallets/subtensor/src/uids.rs
```diff
@@ -9,6 +9,28 @@ impl<T: Config> Pallet<T> {
         SubnetworkN::<T>::get(netuid)
     }
 
+    /// Returns a callback that sets the element at the given position to zero, doing nothing if the
+    /// position is out of bounds
+    fn clear_element_at<N>(position: u16) -> impl Fn(&mut Vec<N>)
```
distributedstatemachine: What is the advantage of using this over `take` or `remove` ?

distributedstatemachine:
@andreea-popescu-reef also changed the base against `devnet-ready` , so you need to resolve conflicts , or fork off devnet-ready

pallets/subtensor/src/uids.rs
andreea-popescu-reef: the vector is indexed by uid so we don't want to remove an element from it - just set it to zero (and check for bounds)

pallets/subtensor/src/subnets/uids.rs
```diff
@@ -9,6 +9,22 @@ impl<T: Config> Pallet<T> {
...
 
+    /// Sets value for the element at the given position if it exists.
+    pub fn set_element_at<N>(vec: &mut [N], position: usize, value: N) {
+        if let Some(element) = vec.get_mut(position) {
+            *element = value;
+        }
+    }
+
+    /// Resets the trust, emission, consensus, incentive, dividends of the neuron to default
+    pub fn clear_neuron(netuid: u16, neuron_uid: u16) {
```
distributedstatemachine, line 19: Can we optimise this by reducing the number of storage / writes , and number of function calls ? 

```rust
pub fn clear_neuron(netuid: u16, neuron_uid: u16) {
    let neuron_index: usize = neuron_uid.into();
    let default_value = 0;

    for storage in &mut [
        &mut Emission::<T>::get(netuid),
        &mut Trust::<T>::get(netuid),
        &mut Consensus::<T>::get(netuid),
        &mut Incentive::<T>::get(netuid),
        &mut Dividends::<T>::get(netuid),
    ] {
        set_element_at(storage, neuron_index, default_value);
    }
}
```

andreea-popescu-reef: that does not quite work as the vectors have different int types

andreea-popescu-reef: it's the same number of calls to the set_element_at. is there much difference between get vs mutate calls? aren't they compiled to similar things?

andreea-popescu-reef: anyway made the change

andreea-popescu-reef: reverted - that does not work

andreea-popescu-reef:
rebased this pr

pallets/subtensor/src/subnets/uids.rs
camfairchild: Can we not just clear the netuid prefix? Or is that the same runtime?

andreea-popescu-reef: it does not modify the vector - &mut Emission::<T>::**get**(netuid) just gives back the value & StorageMap does not have a way to get a mut ref to the value

runtime/src/lib.rs
```diff
@@ -160,7 +160,7 @@ pub const VERSION: RuntimeVersion = RuntimeVersion {
     //   `spec_version`, and `authoring_version` are the same between Wasm and native.
     // This value is set to 100 to notify Polkadot-JS App (https://polkadot.js.org/apps) to use
     //   the compatible custom types.
-    spec_version: 208,
+    spec_version: 209,
```
camfairchild:
```suggestion
    spec_version: 211,
```
Might need to bump here

camfairchild (dismissed):

basfroman (dismissed):

andreea-popescu-reef:
rebased again

camfairchild (approved):

camfairchild (approved):

PR #759, opened 2024-08-27
fix benchmarks + CI

sam0x17:
Fixes broken benchmarks and sets cargo test CI job so that it checks all features, removing the need for a separate benchmarks CI job that was somehow passing anyway


sam0x17:
edit: turns out there just happened to be a force push that broke some benchmarks

keithtensor (approved):

JohnReedV (approved):

unconst (approved):

PR #771, opened 2024-08-29
Add token argument to the publish script

keithtensor:
This PR adds a token argument to the publish script so that we can pass in the correct crates.io token when publishing our crates.


sam0x17:
don't merge, this needs to be on the other repo

distributedstatemachine (approved):

unconst (approved):

PR #772, opened 2024-08-29
Feat/evm devnet ready

### LINKED ISSUE #862: add `skip-cargo-audit` label
Should allow skipping when cargo audit fails, use case is when a cargo audit failure is being fixed in a separate PR but something needs to be hotfixed that is unrelated

gztensor:
## Description
Add EVM functionality to the chain

## Related Issue(s)
n/a

## Type of Change
- [ ] Documentation update

## Breaking Change
n/a

## Checklist

n/a

## Additional Notes

Testing can be done with instructions here:
https://github.com/gztensor/evm-demo

fixes #862


node/src/ethereum.rs
```diff
@@ -0,0 +1,347 @@
...
+}
+
+pub fn new_frontier_partial(
+    config: &EthConfiguration,
+) -> Result<FrontierPartialComponents, ServiceError> {
+    Ok(FrontierPartialComponents {
+        filter_pool: Some(Arc::new(Mutex::new(BTreeMap::new()))),
+        fee_history_cache: Arc::new(Mutex::new(BTreeMap::new())),
+        fee_history_cache_limit: config.fee_history_limit,
+    })
```
sam0x17, line 126: Would be good do discuss the performance implications of having a `BTreeMap` behind an arc+mutex. Perhaps something like a dashmap would be more appropriate if key ordering doesn't matter? Ideally we use something with more fine-grained locking I would think, though I'm not at all familiar with the access pattern here so this might be reasonable. What is the access pattern?

node/src/ethereum.rs
```diff
@@ -0,0 +1,347 @@
...
+        "frontier-fee-history",
+        Some("frontier"),
+        EthTask::fee_history_task(
+            client,
+            storage_override,
+            fee_history_cache,
+            fee_history_cache_limit,
+        ),
+    );
+}
```
sam0x17: this looks good, but we should probably think about adding a mechanism for handling task failures... if any of these tasks fail, it could leave the node in an inconsistent state. Logging and retry mechanisms could be useful here. Basically would feel safer if it was more self-healing

node/src/ethereum.rs
```diff
@@ -0,0 +1,347 @@
...
+            block_data_cache,
+        )
+        .into_rpc(),
+    )?;
+
+    #[cfg(feature = "txpool")]
+    io.merge(TxPool::new(client, graph).into_rpc())?;
+
+    Ok(io)
+}
```
sam0x17: I recommend breaking this up into helper functions, especially if we think this will grow to be even longer

runtime/src/precompiles/mod.rs
```diff
@@ -0,0 +1,116 @@
...
+/// which consumes all gas
+///
+pub fn bytes_to_account_id(account_id_bytes: &[u8]) -> Result<AccountId32, PrecompileFailure> {
+    AccountId32::from_slice(account_id_bytes).map_err(|_| {
+        log::info!("Error parsing account id bytes {:?}", account_id_bytes);
+        PrecompileFailure::Error {
+            exit_status: ExitError::InvalidRange,
+        }
+    })
+}
```
sam0x17: This should be a `TryFrom`

sam0x17 (commented):
looks good, a few areas I think could be improved / explored

node/src/ethereum.rs
gztensor: Didn't think about it, tbh. This comes from a Polkadot-SDK repo:

https://github.com/polkadot-evm/frontier/blob/80c768f688d86bfb350753f3e2aebc9d89037727/template/node/src/eth.rs#L101

sam0x17: gotcha. The litmus test would be lots of contention => dashmap

runtime/src/precompiles/mod.rs
gztensor: Done, please review @sam0x17

node/src/ethereum.rs
gztensor: Done, please review @sam0x17

runtime/Cargo.toml
```diff
@@ -93,6 +93,23 @@ pallet-registry = { default-features = false, path = "../pallets/registry" }
 # Metadata commitment pallet
 pallet-commitments = { default-features = false, path = "../pallets/commitments" }
 
+# Frontier
+fp-evm = { workspace = true }
```
sam0x17, line 98: btw you can:
```toml
fp-evm.workspace = true
```

sam0x17 (dismissed):
looks good so far!

runtime/src/lib.rs
```diff
@@ -1050,6 +1045,218 @@ impl pallet_admin_utils::Config for Runtime {
     type WeightInfo = pallet_admin_utils::weights::SubstrateWeight<Runtime>;
 }
 
+// Define the ChainId
+parameter_types! {
+    pub const SubtensorChainId: u64 = 0x03B1; // Unicode for lowercase alpha
+    // pub const SubtensorChainId: u64 = 0x03C4; // Unicode for lowercase tau
+}
```
orriin, line 1052: @gregzaitsev is this ok?

unconst (approved):

PR #786, opened 2024-09-06
change weight commit logic

soring323:
## Description

- This should only check if the weights_rate_limit time has passed since the last committed weights, without imposing the interval-based restriction.
- The chain should now store multiple weight commits in the WeightCommits storage, so we will modify the storage to retain all commits older than the defined commit_reveal_interval * 2
- Validators can reveal weights if the difference between the current_block and committed_block exceeds the commit_reveal_interval.

## Related Issue(s)

- Closes #[issue number]

## Type of Change

- [ ] Documentation update

## Breaking Change

## Checklist

## Additional Notes

### DIFF
--- a/pallets/subtensor/src/subnets/weights.rs
+++ b/pallets/subtensor/src/subnets/weights.rs
@@ -43,14 +43,10 @@ impl<T: Config> Pallet<T> {
 
         // Insert the new commit with the current block number.
         let current_block = Self::get_current_block_as_u64();
-        WeightCommits::<T>::insert(
-            netuid,
-            &who,
-            (commit_hash, current_block),
-        );
+        WeightCommits::<T>::insert(netuid, &who, (commit_hash, current_block));
 
         // Retain only commits where committed_block < current_block - commit_reveal_interval * 2
-        let interval = Self::get_commit_reveal_weights_interval(netuid) * 2;
+        let interval = Self::get_commit_reveal_weights_interval(netuid).saturating_mul(2);
         let threshold_block = current_block.saturating_sub(interval);
 
         // Iterate through the commits and remove old entries
--- a/pallets/subtensor/src/subnets/weights.rs
+++ b/pallets/subtensor/src/subnets/weights.rs
@@ -38,24 +38,33 @@ impl<T: Config> Pallet<T> {
         // Check that the validator is not committing too frequently (weights_rate_limit check)
         ensure!(
             Self::check_rate_limit_for_commit(netuid, &who),
+            
             Error::<T>::WeightsCommitNotAllowed
         );
 
-        // Insert the new commit with the current block number.
+        // Insert the new commit with the current block number
         let current_block = Self::get_current_block_as_u64();
         WeightCommits::<T>::insert(netuid, &who, (commit_hash, current_block));
 
-        // Retain only commits where committed_block < current_block - commit_reveal_interval * 2
+        // Calculate the threshold block beyond which old commits should be removed
         let interval = Self::get_commit_reveal_weights_interval(netuid).saturating_mul(2);
         let threshold_block = current_block.saturating_sub(interval);
 
-        // Iterate through the commits and remove old entries
-        WeightCommits::<T>::drain_prefix(netuid).for_each(|(who, (_hash, commit_block))| {
-            if commit_block < threshold_block {
-                WeightCommits::<T>::remove(netuid, who);
-            }
-        });
+        // Retrieve all commits for the given network as a vector, sorted by block number
+        let mut commits: Vec<(T::AccountId, (H256, u64))> = WeightCommits::<T>::iter_prefix(netuid).collect();
 
+        // Use binary search to find the first valid commit that is above the threshold
+        let first_valid_index = commits
+            .binary_search_by(|&(_, (_, commit_block))| commit_block.cmp(&threshold_block))
+            .unwrap_or_else(|index| index); // If not found, return the position of the first valid commit
+
+        // If there are old commits, remove them in bulk
+        if first_valid_index > 0 {
+            // Remove old commits that are below the threshold
+            commits.drain(0..first_valid_index).for_each(|(account, _)| {
+                WeightCommits::<T>::remove(netuid, &account);
+            });
+        }
         Ok(())
     }
 
@@ -467,6 +476,7 @@ impl<T: Config> Pallet<T> {
     /// This function checks if the `weights_rate_limit` time has passed since the last
     /// committed weight. If it has, the validator can commit new weights.
     ///
+    /// 
     pub fn check_rate_limit_for_commit(netuid: u16, who: &T::AccountId) -> bool {
         if let Some((_hash, last_commit_block)) = WeightCommits::<T>::get(netuid, who) {
             let current_block: u64 = Self::get_current_block_as_u64();
--- a/pallets/subtensor/src/subnets/weights.rs
+++ b/pallets/subtensor/src/subnets/weights.rs
@@ -59,14 +59,27 @@ impl<T: Config> Pallet<T> {
             .unwrap_or_else(|index| index); // If not found, return the position of the first valid commit
 
         // If there are old commits, remove them in bulk
+        // Optimize the removal of old commits
         if first_valid_index > 0 {
-            // Remove old commits that are below the threshold
-            commits
-                .drain(0..first_valid_index)
-                .for_each(|(account, _)| {
-                    WeightCommits::<T>::remove(netuid, &account);
-                });
+            // Get the range of valid commits
+            let valid_commits = commits.split_off(first_valid_index);
+
+            // Clear all commits for this netuid
+            let _ = WeightCommits::<T>::clear_prefix(netuid, u32::MAX, None);
+
+            // Re-insert only the valid commits
+            for (account, (commit_hash, block_number)) in valid_commits {
+                WeightCommits::<T>::insert(netuid, &account, (commit_hash, block_number));
+            }
         }
+        // if first_valid_index > 0 {
+        //     // Remove old commits that are below the threshold
+        //     commits
+        //         .drain(0..first_valid_index)
+        //         .for_each(|(account, _)| {
+        //             WeightCommits::<T>::remove(netuid, &account);
+        //         });
+        // }
         Ok(())
     }
 
--- a/pallets/subtensor/src/subnets/weights.rs
+++ b/pallets/subtensor/src/subnets/weights.rs
@@ -477,40 +477,6 @@ impl<T: Config> Pallet<T> {
         uids.len() <= subnetwork_n as usize
     }
 
-    // ///
-    // /// This function checks if the `weights_rate_limit` time has passed since the last
-    // /// committed weight. If it has, the validator can commit new weights.
-    // ///
-    // ///
-    // pub fn check_rate_limit_for_commit(netuid: u16, who: &T::AccountId) -> bool {
-    //     if let Some((_hash, last_commit_block)) = WeightCommits::<T>::get(netuid, who) {
-    //         let current_block: u64 = Self::get_current_block_as_u64();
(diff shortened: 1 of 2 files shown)


pallets/subtensor/src/subnets/weights.rs
```diff
@@ -35,16 +35,27 @@ impl<T: Config> Pallet<T> {
             Error::<T>::CommitRevealDisabled
         );
 
+        // Check that the validator is not committing too frequently (weights_rate_limit check)
         ensure!(
-            Self::can_commit(netuid, &who),
+            Self::check_rate_limit_for_commit(netuid, &who),
```
distributedstatemachine, line 40: If you are using rate_limits , please follow the new pattern in utils/rate_limiting

distributedstatemachine (changes requested):
Also add tests to verify the functionality of the changes

pallets/subtensor/src/subnets/weights.rs
soring323: I guess it looks difference concept each other

pallets/subtensor/src/subnets/weights.rs
```diff
@@ -452,50 +463,29 @@ impl<T: Config> Pallet<T> {
...
-
-        // Allow reveal if the current block is within the interval following the commit's interval
-        if current_block >= reveal_interval_start
-            && current_block < reveal_interval_start.saturating_add(interval)
-        {
-            return true;
-        }
-
-        false
+    pub fn is_reveal_valid(commit_block: u64, netuid: u16) -> bool {
```
surcyf123: please add a comment explaining this function

pallets/subtensor/src/subnets/weights.rs
distributedstatemachine: Please look at these 

- rate limit declaration : https://github.com/DedicatedDev/subtensor/blob/datura/feat/update-weight-commit-logic/pallets/subtensor/src/utils/rate_limiting.rs#L28
- how is it applied : https://github.com/DedicatedDev/subtensor/blob/datura/feat/update-weight-commit-logic/pallets/subtensor/src/staking/set_children.rs#L249-L277

Rapiiidooo:
Looks good to me. Only added value to commit reveal.

pallets/subtensor/src/subnets/weights.rs
ppolewicz: getting an error like that is the most confusing thing ever. Please return a very detailed error, saying "too early" or "too late" and detailing the exact extrinsic ids and block numbers involved so that people who need to figure out why a validator failed to set weights have something to go on

pallets/subtensor/src/subnets/weights.rs
```diff
@@ -102,7 +113,7 @@ impl<T: Config> Pallet<T> {
                 .ok_or(Error::<T>::NoWeightsCommitFound)?;
 
             ensure!(
-                Self::is_reveal_block_range(netuid, *commit_block),
+                Self::is_reveal_valid(*commit_block, netuid),
```
ppolewicz: getting an error like that ("it's a bad time") is the most confusing thing ever. Please return a very detailed error, saying "too early" or "too late" and detailing the exact extrinsic ids and block numbers involved so that people who need to figure out why a validator failed to set weights have something to go on

pallets/subtensor/src/subnets/weights.rs
```diff
@@ -35,16 +35,27 @@ impl<T: Config> Pallet<T> {
...
+        WeightCommits::<T>::insert(netuid, &who, (commit_hash, current_block));
+
+        // Retain only commits where committed_block < current_block - commit_reveal_interval * 2
+        let interval = Self::get_commit_reveal_weights_interval(netuid).saturating_mul(2);
+        let threshold_block = current_block.saturating_sub(interval);
+
+        // Iterate through the commits and remove old entries
+        WeightCommits::<T>::drain_prefix(netuid).for_each(|(who, (_hash, commit_block))| {
+            if commit_block < threshold_block {
+                WeightCommits::<T>::remove(netuid, who);
```
ppolewicz: I might be wrong because maybe it's a LinkedList already (which would be another problem), but if it isn't:

Maybe it doesn't matter much because there shouldn't be many of those, but if someone sets the threshold high on a fake subnet, commits a ton of weights from all 64 validators as fast as they can (every 50 blocks I think) and then, say, changes the hyperparameter, it might cause the chain to remove entries from this map inefficiently as soon as someone tries adding something here.

Perhaps it should be a list sorted by block_id and we should bisect in search for the first item on that list to find the first item that is not to be filtered out, and then we'll copy off everything from that point on till the end, making it the new list, causing the entire operation to be log(n) lookups + one memcopy rather than n**2 (looking at each item and removing the (first?) element from an array which requires a rewrite).

pallets/subtensor/src/subnets/weights.rs
```diff
@@ -452,50 +463,29 @@ impl<T: Config> Pallet<T> {
...
-            let interval: u64 = Self::get_commit_reveal_weights_interval(netuid);
-            if interval == 0 {
-                return true; //prevent division by 0
-            }
-
+    ///
+    /// This function checks if the `weights_rate_limit` time has passed since the last
+    /// committed weight. If it has, the validator can commit new weights.
+    ///
+    pub fn check_rate_limit_for_commit(netuid: u16, who: &T::AccountId) -> bool {
```
ppolewicz, line 515: there should be one function which works for both set_weights and commit_weights. If there are two, then sooner or later they will start doing different things.

pallets/subtensor/src/subnets/weights.rs
```diff
@@ -452,50 +463,29 @@ impl<T: Config> Pallet<T> {
...
-            && current_block < reveal_interval_start.saturating_add(interval)
-        {
-            return true;
-        }
-
-        false
+    pub fn is_reveal_valid(commit_block: u64, netuid: u16) -> bool {
+        let interval = Self::get_commit_reveal_weights_interval(netuid);
+        let current_block = Self::get_current_block_as_u64();
+        current_block.saturating_sub(commit_block) > interval
```
ppolewicz: This is almost an exact duplicate of `check_rate_limit_for_commit()` - surely you can pull it off to a private helper such as `is_rate_limit_satisfied(rate_limit_interval)` which gets the current block, checks one against the other, has its own tests and we can read the code in English
```rust
return Self::is_rate_limit_satisfied(Self::get_commit_reveal_weights_interval(netuid))
```
and then you can stick a nice log there so that if it ever malfunctions, you can see what it's doing in debug logs

pallets/subtensor/src/subnets/weights.rs
soring323: added

pallets/subtensor/src/subnets/weights.rs
soring323: @distributedstatemachine  I don't think so. 
Please check this. https://github.com/opentensor/subtensor/blob/58c1f51cc9f0716545fbf99c10c8d761d19dd472/pallets/subtensor/src/subnets/weights.rs#L321
It's similar things. not global feature

soring323: @ppolewicz  "too early" | "too late' it has meaning? 
This is just checking weight set_rate_limit, 
If they did not make a quick update, that's possible to make a commit or not, impossible. 
So current error message enough for this action

pallets/subtensor/src/subnets/weights.rs
soring323: updated

pallets/subtensor/src/subnets/weights.rs
soring323: updated

pallets/subtensor/src/subnets/weights.rs
soring323: added debug logs so I don't think so we need to add additional logs

pallets/subtensor/src/subnets/weights.rs
soring323: It's related to the commit part, do_set_weight function is difference one so it doesn't make sense for me to add this to do_set_weight.

pallets/subtensor/src/subnets/weights.rs
ppolewicz: No it's not. It should clearly say:
you've made a commit at block X
the hyperparameter is at Y
the correct time to reveal is from A to B and you are too early / too late

pallets/subtensor/src/subnets/weights.rs
ppolewicz: I don't think the debug logs are being sent to the client which has to debug the error in their code. Can you return a readable error message all the way to the client, please?

pallets/subtensor/src/subnets/weights.rs
ppolewicz: wrong: it doesn't make sense to write two nearly-identical functions for the same thing

pallets/subtensor/src/subnets/weights.rs
```diff
@@ -41,21 +41,32 @@ impl<T: Config> Pallet<T> {
...
+            .unwrap_or_else(|index| index); // If not found, return the position of the first valid commit
+
+        // If there are old commits, remove them in bulk
+        if first_valid_index > 0 {
+            // Remove old commits that are below the threshold
+            commits
+                .drain(0..first_valid_index)
+                .for_each(|(account, _)| {
+                    WeightCommits::<T>::remove(netuid, &account);
+                });
```
ppolewicz: what's `account` in this context?

Doesn't this call `WeightCommits::<T>::remove();` repeatedely, where each of those operations rewrites the entire list in memory O(N*M), where it could get a range from `first_valid_index..end` and replace the substructure with it in one memcopy O(M)?

soring323: the new solution optimizes the process by reducing the complexity to O(N). Instead of making multiple calls to remove each entry one by one, it uses a bulk operation—clear_prefix()—to clear all weight commits for the specific netuid in one step

pallets/subtensor/src/subnets/weights.rs
soring323: where it's repeated? I guess it's already done as one function.

pallets/subtensor/src/subnets/weights.rs
soring323: it's just the implementation of error parts, so treating it in other PR might be better. 
In macros error implementation, there are only simple default error implementations, I don't want to change this implementation in this PR so we can consider this in a separate PR. maybe you can open your PR for this. 
Thanks

soring323: @distributedstatemachine  added unit test. Thank you

pallets/subtensor/src/subnets/weights.rs
```diff
@@ -59,14 +59,27 @@ impl<T: Config> Pallet<T> {
...
+
+            // Clear all commits for this netuid
+            let _ = WeightCommits::<T>::clear_prefix(netuid, u32::MAX, None);
+
+            // Re-insert only the valid commits
+            for (account, (commit_hash, block_number)) in valid_commits {
+                WeightCommits::<T>::insert(netuid, &account, (commit_hash, block_number));
+            }
         }
+        // if first_valid_index > 0 {
```
distributedstatemachine, line 75: Remove commented code

distributedstatemachine (changes requested):

pallets/subtensor/src/subnets/weights.rs
distributedstatemachine: Sorry i missed this , @DedicatedDev . Can you read the whole rate limiting file , we have per subnet // global limits to choose from . In `set_children` , we use a subnet limit , but the global on is there too.

PR #788, opened 2024-09-08
Indexed Preexecution Errors

JohnReedV:
## Description
This PR modifies the preexecution error codes to ensure that the code matches its index in the metadata. 

## Related Issue(s)

- Closes #[issue number]

## Type of Change

- [x] Documentation update

## Breaking Change
- [x] SDK PR https://github.com/opentensor/bittensor/pull/2305
- [ ] CLI PR ...


pallets/subtensor/src/macros/errors.rs
```diff
@@ -184,5 +184,32 @@ mod errors {
...
+            match e {
+                Error::<T>::ColdkeyInSwapSchedule => 89,
+                Error::<T>::CommitWeightsBelowMinStake => 90,
+                Error::<T>::RevealWeightsBelowMinStake => 91,
+                Error::<T>::SetWeightsBelowMinStake => 92,
+                Error::<T>::SetRootWeightsBelowMinStake => 93,
+                Error::<T>::MaxIntervalRegistrationsReached => 94,
+                _ => u8::MAX,
+            }
+        }
```
sam0x17: so yeah would be great if you add a test with a match statement covering all the error variants and asserting the number for each one that would be sufficient because if a new variant is added or changed the test will either fail to compile or just fail

sam0x17 (approved):

pallets/subtensor/tests/root.rs
```diff
@@ -1052,3 +1052,33 @@ fn test_user_add_network_with_identity_fields_ok() {
...
+                Error::<Test>::SetRootWeightsBelowMinStake => 93,
+                Error::<Test>::MaxIntervalRegistrationsReached => 94,
+                _ => u8::MAX,
+            };
+
+            let value: u8 = error.into();
+            assert_eq!(value, expected_value);
+        }
+    });
+}
```
sam0x17: actually the whole point is you want to do a match on the enum itself so you get a compile error if something is missing / added, so this wouldn't be sufficient

sam0x17 (changes requested):

sam0x17 (approved):

PR #821, opened 2024-09-18
require explicit pallet indexes

### LINKED ISSUE #787: lint check to disallow implicit pallet indexing
Pallet indexing in `construct_runtime!` has been a footgun recently. It would be desirable if we could just disable not using explicit pallet indexing. This can be done with a custom lint.

This is a large item because it requires fully parsing `construct_runtime!` invocations from within our linting environment

sam0x17:
fixes #787

also fixes CI bug that was causing custom lint check to not catch lint failures

Adds a custom lint that completely disallows invocations of `construct_runtime!` that do not use explicit indexes for all mentioned pallets. 

This was extremely tricky to get right. After many attempts to re-implement `construct_runtime!` parsing from scratch, I decided to fork `frame-support-procedural` from upstream and modify it so that instead of a proc-macro crate, it exports all parsing logic so it can be used from other crates. This can be fully updated to whatever tag we want in `polkadot-sdk` using the provided `update.sh` script by modifying this line:

```bash
POLKADOT_SDK_TAG="v1.10.0-rc3"
```

If polkadot-sdk ever adds a public way of accessing the parsing logic for frame internals, we can remove this fork, however that is unlikely to happen any time soon. Maintaining this fork should be super easy since it can be updated with the script whenever we want, and the structure of that crate is likely never going to change (I used to maintain it).

## Notes for Reviewers / TLDR:
* the contents of `procedural-fork` OTHER THAN the `lib.rs`, `update.sh` and `Cargo.toml` are copied verbatim from `polkadot-sdk` 1.10., so those 3 files are all you need to review from that directory
* otherwise all the relevant code changes are in `support/linting/pallet_index.rs` and `build.rs`


sam0x17:
fixing one edge case, do not merge

sam0x17:
should be good to review / merge now!

orriin (dismissed):

Cargo.toml
```diff
@@ -32,7 +32,7 @@ members = [
     "runtime",
     "support/tools",
     "support/macros",
-    "support/linting",
+    "support/linting", "support/procedural-fork",
```
keithtensor:
```suggestion
    "support/linting",
    "support/procedural-fork",
```

support/linting/Cargo.toml
```diff
@@ -7,6 +7,7 @@ edition = "2021"
 syn.workspace = true
 quote.workspace = true
 proc-macro2.workspace = true
+procedural-fork = { version = "1.10.0-rc3", path = "../procedural-fork" }
```
keithtensor, line 10: Shouldn't just the path attribute work alone, without the version?

support/linting/src/pallet_index.rs
```diff
@@ -0,0 +1,301 @@
...
+
+    fn check_pallets_for_index(
+        &mut self,
+        pallets: &[procedural_fork::exports::construct_runtime::parse::Pallet],
+    ) {
+        for pallet in pallets {
+            // Check for explicit index and detect missing indices
+            if !self
+                .original_tokens
+                .contains(format!("={},", pallet.index).as_str())
```
keithtensor, line 90: Technically speaking, we really just need to look for the `=` sign, since it MUST also have been followed by a number in order for it to compile, but this here also checks whether the pallet index is what we expect so it's fine?

support/linting/Cargo.toml
sam0x17: not if we publish to crates.io eventually, need both

support/linting/src/pallet_index.rs
sam0x17: problem is things like `=1` vs `=10`

sam0x17: what we're verifying here is that the index _isn't_ an auto-calculated one, so we are verifying that it appears somewhere in the code basically because FRAME's parsing code just stores it as a `u8` for these two cases and doesn't differentiate between auto-calculated and explicitly specified.

JohnReedV (dismissed):

orriin (dismissed):

support/linting/src/pallet_index.rs
gztensor, line 90: It would be great to have a freeze on construct_runtime so that pallet indexes are preserved, but that should be a separate PR.

gztensor (dismissed):

support/linting/src/pallet_index.rs
sam0x17: yes will be doing this in a separate PR thanks for the suggestion!!

sam0x17:
had to turn off `cargo test` for `procedural-fork` because the tests rely on a bunch of janky features to pass and no need for that, if it compiles we are good / we just care about parsing

gztensor (approved):

orriin (approved):

unconst (approved):

PR #832, opened 2024-09-27
Feat/runtime apis return types

camfairchild:
## Description

Note: depends on https://github.com/opentensor/btcli/pull/188

This PR is prepping for the ability to pull metadata straight from the nodes for use in decoding runtime API results. 
The return types are only available in the metadata if they are set in the runtime API declarations. 

To keep the RPCs, we will still be using the `Vec<u8>` encoding/decoding process *in just the RPC call* but the RuntimeAPIs will use the expected types as the client will now be expected to adjust when decoding based on the block metadata.

This change breaks the RuntimeAPIs, but *not* the RPCs. However, it guarantees that future RuntimeAPI changes will not break the client.

## Related
- #831: A RAO branch companion-PR

## Type of Change

- [ ] Documentation update

## Breaking Change

## Checklist

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -9883,6 +9883,7 @@ dependencies = [
  "serde",
  "sp-api",
  "sp-blockchain",
+ "sp-core",
  "sp-rpc",
  "sp-runtime",
  "subtensor-custom-rpc-runtime-api",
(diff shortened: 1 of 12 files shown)


distributedstatemachine (dismissed):

camfairchild:
Replaced by #1227

PR #833, opened 2024-09-28
Updated release occurrences with production to match the new --profile flag

nolonek:
Hello, made some edits to the script.

1. Replaced 'release' occurrences with 'production' to match the new --profile=production flag.
2. Switched fast blocks off. Caused issue when using script to build.


open-junius:
1. release is much faster than production for compilation. we can use it in local env for testing.
2. can you specify what's kind of issue caused by fast block, it is important feature to make test fast.

PR #848, opened 2024-10-09
code coverage suite

sam0x17:
adds code coverage suite to the repo

- [x] find and parse all pallets dynamically and in parallel
- [x] code to import missing sections of pallets defined with `#[import_section(..)]` / `#[pallet_section]`
- [x] collect metadata on all extrinsics/calls in all pallets, in parallel
- [ ] collect metadata on free functions defined within pallets (non-calls)
- [x] find all tests in the repo in parallel
- [x] find all benchmark tests (v2 syntax) in the repo in parallel
- [x] upgrade existing benchmarking v1 tests to v2 (only one file, quick)
- [x] add a quick little lint disallowing benchmarking v1 syntax and enable this lint in `build.rs`
- [ ] collect metadata and coverage info for hooks, might be impractical, happy path would be if people commonly enter hooks manually within tests, then you can detect tests that do this and expect one covering each hook state, possibly for each pallet
- [x] analyze tests and generate coverage statistics based on what functions are/are not mentioned across all tests
- [x] display code coverage results in the terminal from the build script (using sam's `build-print` crate)
- [ ] by default don't run code coverage unless `RUN_COVERAGE` env var is present (from within `build.rs`)

future work:
* take function complexity into account (number of distinct terminal points = needs more coverage)
* use conventional code coverage tools for everything that _isn't_ a pallet and then roll this information into the final report
* once 100% coverage is achieved, have CI fail if coverage drops below 100% and flag the specific functions/extrinsics/whatever that are untested
* take function call graph into account -- if function or extrinsic A calls function or extrinsic B, then coverage on A at least partially covers B


sam0x17:
closed in favor of #1008

Issue #887, opened 2024-10-30
Separate types for TAO and Alpha balances (instead of u64)


gztensor:

ales-otf:
That would make the compiler catch errors like [this](https://github.com/opentensor/subtensor/issues/879) for us.

PR #888, opened 2024-10-30
Use alpha for stake in neuron info

### LINKED ISSUE #879: Inconsistent stake numbers across RPCs
NeuronInfo RPC currently reports stake values like follows:

```
NeuronInfoLite(hotkey='5DfZR9qRhv2C7JT2fxJ2dLUX1QPZYQwwA5D3NGGUBW3RLsXW', coldkey='5CvaAT12X8nJBczNDm6QoqFgDrsWhtzkr4F5wdnD8yYz326C', uid=0, netuid=2, active=True, stake=τ 34.8542, stake_dict={'5CvaAT12X8nJBczNDm6QoqFgDrsWhtzkr4F5wdnD8yYz326C': τ 34.8542}, total_stake=τ 34.8542, rank=0.0, emission=28.210070956, incentive=0.0, consensus=0.0, trust=0.0, validator_trust=0.0, dividends=0.0, last_update=181, validator_permit=True, prometheus_info=PrometheusInfo(block=0, version=0, ip='0.0.0.0', port=0, ip_type=0), axon_info=AxonInfo(version=0, ip='0.0.0.0', port=0, ip_type=0, hotkey='5DfZR9qRhv2C7JT2fxJ2dLUX1QPZYQwwA5D3NGGUBW3RLsXW', coldkey='5CvaAT12X8nJBczNDm6QoqFgDrsWhtzkr4F5wdnD8yYz326C', protocol=0, placeholder1=0, placeholder2=0), pruning_score=65535, is_null=False)
```

which is inconsistent with StakeInfo RPC. This can be reproduced with btcli by running:

```bash
btcli st list
btcli w balance
btcli w inspect
```

Image
Image
Image

ales-otf:
## Description

This PR fixes inconsistency issues related to staking values returned by RPC's `NeuronInfoRuntimeApi` and `StakeInfoRuntimeApi`.

Getters for the `stake` values are added to `NeuronInfo`, `NeuronInfoLite` and `StakeInfo`, because there is no other way to test this values in the integration tests.

## Related Issue(s)

- Closes #879

## Type of Change

- [ ] Documentation update

## Checklist

## Additional Notes

It would be better to introduce different types for TAO and Alpha balances, so these kinds of errors would be discoverable at the compilation time (#887).

Also, to prevent possible inconsistencies between `NeuronInfo` and `NeuronInfoLite` it would be better to make `NeuronInfoLite` a part of `NeuronInfo`, instead of duplicating information between them.

Allowing unit/module tests would help with private API testing. In case the module tests are not possible, another option could be a cargo feature, which would make API dedicated to tests being compiled for tests only (`#[cfg(test)]` doesn't work for the integration tests).

### DIFF
--- a/pallets/subtensor/src/rpc_info/neuron_info.rs
+++ b/pallets/subtensor/src/rpc_info/neuron_info.rs
@@ -1,6 +1,5 @@
 use super::*;
 use frame_support::pallet_prelude::{Decode, Encode};
-use frame_support::storage::IterableStorageDoubleMap;
 extern crate alloc;
 use codec::Compact;
 
@@ -117,8 +116,12 @@ impl<T: Config> Pallet<T> {
                 }
             })
             .collect::<Vec<(Compact<u16>, Compact<u16>)>>();
-        let stake_weight: u64 = Self::get_stake_weight(netuid, uid) as u64;
-        let stake: Vec<(T::AccountId, Compact<u64>)> = vec![(coldkey.clone(), stake_weight.into())];
+
+        let stake: Vec<(T::AccountId, Compact<u64>)> = vec![(
+            coldkey.clone(),
+            Alpha::<T>::get((&hotkey, coldkey.clone(), netuid)).into(),
+        )];
+
         let neuron = NeuronInfo {
             hotkey: hotkey.clone(),
             coldkey: coldkey.clone(),
@@ -177,12 +180,10 @@ impl<T: Config> Pallet<T> {
         let last_update = Self::get_last_update_for_uid(netuid, uid);
         let validator_permit = Self::get_validator_permit_for_uid(netuid, uid);
 
-        let stake: Vec<(T::AccountId, Compact<u64>)> =
-            <Stake<T> as IterableStorageDoubleMap<T::AccountId, T::AccountId, u64>>::iter_prefix(
-                hotkey.clone(),
-            )
-            .map(|(coldkey, stake)| (coldkey, stake.into()))
-            .collect();
+        let stake: Vec<(T::AccountId, Compact<u64>)> = vec![(
+            coldkey.clone(),
+            Alpha::<T>::get((&hotkey, coldkey.clone(), netuid)).into(),
+        )];
 
         let neuron = NeuronInfoLite {
             hotkey: hotkey.clone(),
--- a/pallets/subtensor/src/rpc_info/neuron_info.rs
+++ b/pallets/subtensor/src/rpc_info/neuron_info.rs
@@ -28,6 +28,12 @@ pub struct NeuronInfo<T: Config> {
     pruning_score: Compact<u16>,
 }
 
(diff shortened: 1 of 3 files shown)


camfairchild:
The reason we use `stake_weight` is because this is what miners need to prioritize Validator requests appropriately. 
This value directly impacts a miner's emission, versus the alpha entry, which is not meaningful.

ales-otf:
> The reason we use `stake_weight` is because this is what miners need to prioritize Validator requests appropriately. This value directly impacts a miner's emission, versus the alpha entry, which is not meaningful.

But in `NeuronInfoLite` the value was from `Stake` storage, while in `NeuronInfo` from `StakeWeight`, and in `StakeInfo` it's `Alpha`. Which one is true then?

Or is it `StakeWeight` only in `NeuronInfo`?

ales-otf:
> The reason we use `stake_weight` is because this is what miners need to prioritize Validator requests appropriately. This value directly impacts a miner's emission, versus the alpha entry, which is not meaningful.

Can we add additional field `stake_weight` to `NeuronInfo` and `NeuronInfoLite`? Also, if we can, could we make `NeuronInfoLite` being a part of `NeuronInfo` to omit the duplication and inconsistencies?

camfairchild:
> > The reason we use `stake_weight` is because this is what miners need to prioritize Validator requests appropriately. This value directly impacts a miner's emission, versus the alpha entry, which is not meaningful.
> 
> But in `NeuronInfoLite` the value was from `Stake` storage, while in `NeuronInfo` from `StakeWeight`, and in `StakeInfo` it's `Alpha`. Which one is true then?
> 
> Or is it `StakeWeight` only in `NeuronInfo`?

I believe this change is attempting to keep the same struct type-wise, while changing the value to be as meaningful to the client as it was.

When we didn't have the `Alpha` map, the only value that mattered in this context was the `Stake` map. But now we use a calculated `stake_weight` to mean the same thing. So this is what should be returned

camfairchild:
> Can we add additional field `stake_weight` to `NeuronInfo` and `NeuronInfoLite`? 

We could, but the big issue with this is we don't have proper versioning set up for the RuntimeAPIs and RPCs. So this will break any clients that rely on this API (many).

gztensor:
@camfairchild , we have `get_global_for_hotkey_and_coldkey` and `get_global_for_hotkey_and_coldey_on_subnet` methods. Can we just use them in `StakeInfo` and `NeuronInfo` respectively?

ales-otf:
Closing, as `NeuronInfo::stake` should be `StakeWeight` and the difference between `StakeInfo::stake` and `NeuronInfo::stake` is expected and should be handled on the client's side.

To fix the inconsistency between `NeuronInfo::stake` and `NeuronInfoLite::stake` a separate [PR](https://github.com/opentensor/subtensor/pull/896) is open.

PR #924, opened 2024-11-06
Duplicate hotfixes to devnet-ready

### LINKED ISSUE #917: Check for childkey stake when setting weights
The `set_weights` extrinsic, as well as `commit_weights` are checking if the weight setting hotkey has enough stake.

- `commit_weights` does it here: https://github.com/opentensor/subtensor/blob/dfbed6bb24cef95fced901597b4910de85705c2d/pallets/subtensor/src/lib.rs#L1446
- `set_weights` does it here: https://github.com/opentensor/subtensor/blob/aa521f46fab9ba4702c752792ba4f5eb38df39fd/pallets/subtensor/src/subnets/weights.rs#L231

but both check using `check_weights_min_stake`, which uses `TotalColdkeyStake`. We need to modify `check_weights_min_stake` so that it considers parent/child stake.

Acceptance criteria:
- [ ] Update `check_weights_min_stake`
- [ ] Unit test or e2e test for each path that sets weights

keithtensor:
Refs #938.

Contains the following PRs to merge into devnet-ready:

* skip netuid 0 for pending emission accum:
https://github.com/opentensor/subtensor/pull/933
* handle stake deltas better
https://github.com/opentensor/subtensor/pull/934
* respect chk for checking stake & weight filtering:
https://github.com/opentensor/subtensor/pull/935
* use chk for neuron info lite stake:
https://github.com/opentensor/subtensor/pull/936

### DIFF
--- a/pallets/subtensor/src/lib.rs
+++ b/pallets/subtensor/src/lib.rs
@@ -1299,7 +1299,7 @@ pub mod pallet {
         /// Returns the transaction priority for setting weights.
         pub fn get_priority_set_weights(hotkey: &T::AccountId, netuid: u16) -> u64 {
             if let Ok(uid) = Self::get_uid_for_net_and_hotkey(netuid, hotkey) {
-                let _stake = Self::get_total_stake_for_hotkey(hotkey);
+                let _stake = Self::get_stake_for_hotkey_on_subnet(hotkey, netuid);
                 let current_block_number: u64 = Self::get_current_block_as_u64();
                 let default_priority: u64 =
                     current_block_number.saturating_sub(Self::get_last_update_for_uid(netuid, uid));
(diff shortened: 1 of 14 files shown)


pallets/subtensor/src/lib.rs
```diff
@@ -1309,9 +1309,17 @@ pub mod pallet {
         }
 
         /// Is the caller allowed to set weights
-        pub fn check_weights_min_stake(hotkey: &T::AccountId) -> bool {
+        pub fn check_weights_min_stake(hotkey: &T::AccountId, netuid: u16) -> bool {
```
gztensor, line 1312: We should also consider stake that this hotkey gives to children, not only the one it receives from parents. There's a `get_stake_for_hotkey_on_subnet` function in `run_epoch.rs` that implements that. I think we should use it all over instead of `get_total_stake_for_hotkey` for this feature.

gztensor (changes requested):

gztensor (approved):

camfairchild (approved):

unconst (approved):

PR #927, opened 2024-11-06
Lint `ForbidKeysRemoveCall`

### LINKED ISSUE #915: Check in static analysis for removals from Keys map
Check in static analysis that we never remove from `Keys` unless we have matching insert at the same function that inserts with the same map key like in `replace_neuron`:

```rust
// Remove uid_to_replace
Keys::<T>::remove(netuid, uid_to_replace);
...
// Insert with the same key: uid_to_replace
Keys::<T>::insert(netuid, uid_to_replace, new_hotkey.clone());
```

If we remove from `Keys`, epoch function is going to break because it relies on neuron IDs being consecutive. 

Note: Currently we don't have such removals anywhere but `replace_neuron` function, so we can limit such replacements that use `Keys::<T>::remove` to only be present in `replace_neuron`.

eagr:
## Description

Add a custom lint to forbid `Keys::<T>::remove()`

## Related Issue(s)

- Closes #915 

## Type of Change

- [ ] Documentation update
- [x] Other (please describe): lint

### DIFF
--- a/support/linting/src/forbid_as_primitive.rs
+++ b/support/linting/src/forbid_as_primitive.rs
@@ -45,10 +45,11 @@ fn is_as_primitive(ident: &Ident) -> bool {
 #[cfg(test)]
 mod tests {
     use super::*;
+    use quote::quote;
 
-    fn lint(input: &str) -> Result {
-        let expr: ExprMethodCall = syn::parse_str(input).expect("should only use on a method call");
+    fn lint(input: proc_macro2::TokenStream) -> Result {
         let mut visitor = AsPrimitiveVisitor::default();
+        let expr: ExprMethodCall = syn::parse2(input).expect("should be a valid method call");
         visitor.visit_expr_method_call(&expr);
         if !visitor.errors.is_empty() {
             return Err(visitor.errors);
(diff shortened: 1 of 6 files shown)


eagr:
first, let's see if the lint complains about the usage in `replace_neuron`, then I'll disable the lint in it

sam0x17:
looks like this one is failing @eagr: https://github.com/opentensor/subtensor/actions/runs/11709455612/job/32816680567?pr=927

eagr:
yeah as expected, kinda wanted to watch it fail first. I'll disable the lint on that line then it should pass

eagr:
I think `#[allow(unknown_lints)]` is meant for situations like this (to prevent the compiler from failing when using custom lints) right?

sam0x17:
looks good to me except I would prefer in the tests you use `quote!` directly so we don't have to do any string parsing. Makes the tests a little bit faster. See the tests for the other lints I did for inspiration! @eagr

eagr:
done

support/linting/src/forbid_as_primitive.rs
```diff
@@ -45,10 +45,11 @@ fn is_as_primitive(ident: &Ident) -> bool {
 #[cfg(test)]
 mod tests {
     use super::*;
+    use quote::quote;
 
-    fn lint(input: &str) -> Result {
-        let expr: ExprMethodCall = syn::parse_str(input).expect("should only use on a method call");
+    fn lint(input: proc_macro2::TokenStream) -> Result {
         let mut visitor = AsPrimitiveVisitor::default();
+        let expr: ExprMethodCall = syn::parse2(input).expect("should be a valid method call");
```
sam0x17, line 52: should just `?` here instead of expecting

eagr: I think it should panic here. otherwise, invalid syntax in a test case expecting an error could be a false positive. regardless, invalid syntax should be considered a bug in the test no?

sam0x17: ahhh you are right

sam0x17 (approved):

PR #930, opened 2024-11-06
fix: host functions (main)

orriin:
Re-implement https://github.com/opentensor/subtensor/pull/801


gztensor (approved):

unconst (approved):

jp-blockwise:
What about dockerised setup? It still fails.

PR #985, opened 2024-11-15
Add RPC API for getting rate limits information

### LINKED ISSUE #892: Add RuntimeAPI for Rate Limit Checks

Currently, submitting weights, registering networks, and staking/unstaking (among others) are rate-limited. It would be useful to be able to check this rate limit via an API without having to send off a transaction and receive an error. 

Add a Runtime API that returns whether a specific key (if applicable) is within the rate limit for a next extrinsic.

- The alternative is to pull the rate limit value and the current usage from storage and duplicate the check for the rate limit on the client-side. But this would be difficult if any changes are made on the chain-side later-on and would involve a level of coupling.

### Additional context

_No response_

ales-otf:
## Description

This PR adds `RateLimitInfoRuntimeApi`, which provides RPC methods to fetch information about rate-limited transactions. The intention behind it is to allow clients to build logic not based on request errors, but by actual values.

## Related Issue(s)

- Closes https://github.com/opentensor/subtensor/issues/892

## Type of Change

- [ ] Documentation update

## Breaking Change

## Checklist

### DIFF
--- a/pallets/subtensor/src/rpc_info/rate_limit_info.rs
+++ b/pallets/subtensor/src/rpc_info/rate_limit_info.rs
@@ -1,14 +1,14 @@
-//! API for getting transaction rate limits associated with coldkeys and hotkeys.
-
+//! API for getting rate-limited transactions info.
 use codec::Compact;
 use frame_support::pallet_prelude::{Decode, Encode};
 
 use crate::{
-    utils::rate_limiting::TransactionType, Config, Pallet, TargetStakesPerInterval, TxRateLimit,
+    freeze_struct, utils::rate_limiting::TransactionType, Config, Pallet, TargetStakesPerInterval,
+    TxRateLimit,
 };
 
 /// Transaction rate limits.
-// #[freeze_struct("fe79d58173da662a")]
+#[freeze_struct("e3734bd0690f2da8")]
 #[derive(Decode, Encode, Clone, Debug)]
 pub struct RateLimits {
     transaction: Compact<u64>,
@@ -18,7 +18,7 @@ pub struct RateLimits {
 }
 
 /// Contains last blocks, when rate-limited transactions was evaluated.
-// #[freeze_struct("fe79d58173da662a")]
+#[freeze_struct("9154106cd720ce08")]
 #[derive(Decode, Encode, Clone, Debug)]
 pub struct HotkeyLimitedTxInfo<AccountId> {
     hotkey: AccountId,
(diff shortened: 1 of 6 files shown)


pallets/subtensor/src/rpc_info/rate_limit_info.rs
```diff
@@ -0,0 +1,56 @@
...
+#[freeze_struct("e3734bd0690f2da8")]
+#[derive(Decode, Encode, Clone, Debug)]
+pub struct RateLimits {
+    transaction: Compact<u64>,
+    set_children: Compact<u64>,
+    set_childkey_take: Compact<u64>,
+    stakes: Compact<u64>,
+}
+
+/// Contains last blocks, when rate-limited transactions was evaluated.
```
gztensor, line 20: This design doesn't account for all of requirements:

1. As the issue described, we need to provide the rate limit information for: "submitting weights, registering networks, and staking/unstaking (among others)", not all of these are exposed

2. The following concern from the issue is not addressed either: "But this would be difficult if any changes are made on the chain-side later-on and would involve a level of coupling." We need to provide a consistent rate limit information across all rate-limited items. This is just a suggestion made for example, but it could be designed as follows:

- Requests remaining this interval
- Total requests allowed per interval
- Next interval start

Then for interval based limits like subnet registration it just fits the design and for cool-off type of limiting it can work as "remaining requests = 0" and "Next interval start = block when cool off period ends".

gztensor (changes requested):

pallets/subtensor/src/rpc_info/rate_limit_info.rs
gztensor: First of all, the client software needs a way to know:
1. If it is allowed to make a request now
2. If not, then how long does it need to wait

ales-otf: I've found that I forgot to add `NetworkRateLimit`, BTW.

As I saw it, the problem is that when you deal with limits you get the error and you need to figure out whether you are allowed to perform the request without getting error - without performing an actual transaction.

"submitting weights":

There is already API for getting weight limits: `SubnetInfoRuntimeApi::get_subnet_hyperparams`:

- `min_allowed_weights`
- `max_weights_limit`
- `weights_rate_limit`

"registering networks":

Regarding subnet registrations per block: `SubnetInfoRuntimeApi::get_subnet_hyperparams::max_regs_per_block`.

Couldn't find anything more related to network registration rate limits.

Staking/unstaking covered by two methods: `RateLimitInfoRuntimeApi::get_rate_limits` and
`RateLimitInfoRuntimeApi::get_stakes_this_interval`.

If the clients want to check whether they exceeded the quota, they check the number of stakes allowed during interval `stakes` and compare it to the stakes made during interval. 

We can't cover all the cases and logic, but given this data clients can build the logic, which prevent from executing transactions.

sam0x17:
bit late to this but idea: what if we just provide rate limiting information in the responses to all rate-limited requests. Like tell them how close to the quota they are at all times. Then we won't have extra traffic of them constantly checking these rpcs

PR #1002, opened 2024-11-19
Fix chainspec update again

sam0x17:
Have to exclude source branches that are immutable


sam0x17:
note to reviewers: makes sense the chainspec upgrade ran again, this is the first PR to have it since it was merged, so there were some previous PRs that got merged that didn't have it so it's bringing in those changes

gztensor (approved):

JohnReedV (approved):

PR #1051, opened 2024-12-03
Fix archive sync panic

orriin:
- Fixes the archive sync getting stuck at block `2585477` due to panic by adding a `codeSubstitute` that patches the buggy code with one that doesn't panic.
- Updates `node/src/chain_spec/finney.rs` so Finney chainspec always builds with the `codeSubstitues` field.


camfairchild (dismissed):

scripts/build_all_chainspecs.sh
```diff
@@ -55,8 +69,8 @@ update_spec() {
 
   echo "*** Restoring genesis in '$chain'..."
 
-  update_genesis "$raw_genesis_temp" "$raw_spec_temp" "$raw_path"
-  update_genesis "$plain_genesis_temp" "$plain_spec_temp" "$plain_path"
+  update_genesis_and_code_substitutes "$raw_genesis_temp" "$raw_code_substitutes_temp" "$raw_spec_temp" "$raw_path"
+  update_genesis_and_code_substitutes "$plain_genesis_temp" "$plain_code_substitutes_temp" "$plain_spec_temp" "$plain_path"
 
   # cleanup
   rm -f "$raw_genesis_temp" "$plain_genesis_temp" "$raw_spec_temp" \
```
ales-otf: Makes sense adding `raw_code_substitutes_temp` and `plain_code_substitutes_temp` here as well.

ales-otf (commented):
It would be better to have the substitute in the `ChainSpec` builder or in the function producing the chain spec instead. It's more reliable and doesn't require patching. If it's possible.

ales-otf:
> It would be better to have the substitute in the `ChainSpec` builder or in the function producing the chain spec instead. It's more reliable and doesn't require patching. If it's possible.

Probably this could help: https://paritytech.github.io/polkadot-sdk/master/sc_chain_spec/fn.set_code_substitute_in_json_chain_spec.html

orriin:
> It would be better to have the substitute in the ChainSpec builder or in the function producing the chain spec instead. It's more reliable and doesn't require patching. If it's possible.

@ales-otf  I tried this but couldn't figure it out. It seems the `ChainSpec` Builder doesn't support setting `code_substitutes`, also the fn you linked seems to operate on a JSON object, whereas we have a `ChainSpec`. Let me know if you have any ideas tho! Would be great to get it working that way.

ales-otf:
> > It would be better to have the substitute in the ChainSpec builder or in the function producing the chain spec instead. It's more reliable and doesn't require patching. If it's possible.
> 
> @ales-otf I tried this but couldn't figure it out. It seems the `ChainSpec` Builder doesn't support setting `code_substitutes`, also the fn you linked seems to operate on a JSON object, whereas we have a `ChainSpec`. Let me know if you have any ideas tho! Would be great to get it working that way.

~~I believe you can modify this [JSON](https://docs.rs/serde_json/1.0.132/serde_json/value/enum.Value.html): https://github.com/opentensor/subtensor/blob/devnet-ready/node/src/chain_spec/finney.rs#L198~~

~~Oh, it's just genesis. Then yeah. Looks like you have to patch the json the way you do.~~ As an option, you can get the json from the `ChainSpec` (https://paritytech.github.io/polkadot-sdk/master/sc_chain_spec/struct.GenericChainSpec.html#method.as_json) modify it using the function and build the `ChainSpec` from the modified JSON. It's ugly, but it's more safe doing it in rust, than in a shell script.

BTW, it looks like we can omit using the script and just hardcode everything, as there is [with_code](https://paritytech.github.io/polkadot-sdk/master/sc_chain_spec/struct.ChainSpecBuilder.html#method.with_code) method.

orriin:
@ales-otf that is the genesis JSON not the chain spec JSON

ales-otf:
> @ales-otf that is the genesis JSON not the chain spec JSON

yeah, I've updated my comment

orriin:
> As an option, you can get the json from the ChainSpec (https://paritytech.github.io/polkadot-sdk/master/sc_chain_spec/struct.GenericChainSpec.html#method.as_json) modify it using the function and build the ChainSpec from the modified JSON. It's ugly, but it's more safe doing it in rust, than in a shell script.

Good spot! I will have a go at this.

> BTW, it looks like we can omit using the script and just hardcode everything, as there is [with_code](https://paritytech.github.io/polkadot-sdk/master/sc_chain_spec/struct.ChainSpecBuilder.html#method.with_code) method.

I don't think `with_code` is what we want, I believe it sets the genesis code rather than a code substitute for a specific block number

ales-otf:
> I don't think `with_code` is what we want, I believe it sets the genesis code rather than a code substitute for a specific block number

Yeah, that's exactly what I meant. We now need to use this script to keep the genesis mostly because of the genesis code. The rest comes frome that `generate_genesis` function. So if we put the code value into `ChainSpec` builder and if we also add substitute there, we can not only omit the shell script, but we can even connect to the running mainnet using `cargo run -- --chain finney` or something like this - without requiring to have the chainspec file at all. Still, it would be good to track changes in the chainspecs. Some warning system at least. Also, it would be better to extract the code value into a separate file and include it into the source from that file. It would be easier to edit the file and the git diffs would be smaller.

Opened an [issue](https://github.com/opentensor/subtensor/issues/1057) for this.

ales-otf:
> > I don't think `with_code` is what we want, I believe it sets the genesis code rather than a code substitute for a specific block number
> 
> Yeah, that's exactly what I meant. We now need to use this script to keep the genesis mostly because of the genesis code. The rest comes frome that `generate_genesis` function. So if we put the code value into `ChainSpec` builder and if we also add substitute there, we can not only omit the shell script, but we can even connect to the running mainnet using `cargo run -- --chain finney` or something like this - without requiring to have the chainspec file at all. Still, it would be good to track changes in the chainspecs. Some warning system at least. Also, it would be better to extract the code value into a separate file and include it into the source from that file. It would be easier to edit the file and the git diffs would be smaller.

This is definitely out of scope of this PR. I've just spotted the method and this idea came to mind.

orriin:
A bit hacky but figured out a solution for setting `codeSubstitutes` at the `node/src/chain_spec/finney.rs` level, thanks @ales-otf !

node/src/chain_spec/finney.rs
```diff
@@ -181,7 +181,27 @@ pub fn finney_mainnet_config() -> Result<ChainSpec, String> {
...
-    .build())
+    .build();
+
+    // Load and set the code substitute to avoid archive node sync panic
+    // See <https://github.com/opentensor/subtensor/pull/1051>
+    //
+    // Need to do it in this hacky way because the ChainSpec builder doesn't support setting it
+    let manifest_dir = env!("CARGO_MANIFEST_DIR");
+    let blob_path =
+        std::path::Path::new(manifest_dir).join("src/chain_spec/code_substitute_2585476.txt");
```
ales-otf: This requires having `code_substitute_2585476.txt` at the specified path at the runtime. You need to embed it into binary using `include_str!`or `include_bytes!` macro.

ales-otf (changes requested):
Please, look the comment - this change is important. 

And just an additional not much important styling thing, which would improve readability and debugging — to encapsulate this whole part related to setting the substitute into a separate function.

node/src/chain_spec/finney.rs
ales-otf: Also, I believe in case of `include_*` macros you can use a relative path.

orriin: I'm not sure what you mean, `code_substitute_2585476.txt` is checked into Git at the specified path so it should always resolve correctly.

Why does it need to be embedded into the binary?

ales-tsurko: Because the node is a binary, you can install it in a different location with different directory structure. If you build your node and move the binary from the build directory to another path then run it with `--chain finney` it will panic, because it won't find `code_substitute_2585476.txt`. Moreover, `CARGO_MANIFEST_DIR` won't be available at that time. For the node to work correctly, you need either provide this file and `CARGO_MANIFEST_DIR`, change the path to relative to binary and also provide the file, or simply embed it into the binary.

ales-otf: Just imagine you installed the node at `/foo/bar/node-subtensor` and what will happen then if you run it like `/foo/bar/node-subtensor --chain finney`. Not taking `CARGO_MANIFEST_DIR` availability into account, even if there were a relative to binary path, the binary would expect `code_substitute_2585476.txt` at `/foo/bar/src/chain_spec/code_substitute_2585476.txt`.

orriin: I do see your point there is a benefit, but the chain spec is almost 1mb, it would bloat the binary a lot if we added it. 

I can't think of when we would want to generate the chain spec outside of the repo using the bare binary, so I think the costs outweigh the benefits to embed it. Let me know if you think I'm missing something though.

ales-otf: Yeah, unfortunately. I mentioned this in the issue, where I suggested embedding the genesis blob into the binary. For validator's and standard nodes it's fine, given the size of the blockchain and that you have the chainspec file behind the binary anyway. But for the lightnodes (if we will have one), it's definitely not good. It's a subject for a discussion though, as we have to implement it separately from a normal node using something like https://github.com/smol-dot/smoldot, don't we?
 
If we go the way when we should solve the path at the runtime, then instead of panicing, it would be better to extend the `build-spec` CLI command with a required argument that accepted the genesis file. At least that would persist the portability. Otherwise, using the script as you initially did is a better solution, than relying on a file that may not exist at the runtime.

orriin: > But for the lightnodes (if we will have one), it's definitely not good

we don't support lightnodes

> If we go the way when we should solve the path at the runtime, then instead of panicing, it would be better to extend the build-spec CLI command with a required argument that accepted the genesis file.

it might be a bit difficult to extend the build-spec cli. I checked the release binary size and it's already 82mb, so I think adding another 1mb is fine. I will just do that for now  :)

orriin: Hey @ales-otf, please take another look

ales-otf (approved):
Thanks!

sam0x17 (approved):

PR #1093, opened 2024-12-12
Fix staking counter

### LINKED ISSUE #1091: Fix Stake rate limit for target>1
### Describe the bug

Currently, we rate limit staking transactions by checking how many staking transactions were done within a set interval, and limiting if it exceeds the set target.

The way we do this is flawed. In the `get_stakes_this_interval_for_coldkey_hotkey` function, we take the value stored in `TotalHotkeyColdkeyStakesThisInterval` of `(num_stakes, last_stake_block)`.

We reset `num_stakes` to `0` if it has been more than `stake_interval` blocks since `last_stake_block`.
Otherwise we return `num_stakes` with no reset.

This passes the rate limit, and then we do the transaction. We then increment `num_stakes` and set `last_stake_block` to the current block.

This is fine for `stake_interval == 1`, but for anything greater, we run into an issue.
Example with `stake_interval = n` where `n > 1`:
1. stake `n - 1` times at block `X`
2. wait `stake_interval - 1` blocks
3. stake `1` time at block `X + stake_interval - 1`
4. wait for next block;
5. stake again ==> failed, rate limited

Why does this fail? 
In 4., we *should* expect the rate limit count to reset to `1` as we have only staked once in the last `stake_interval` blocks. However, because of the faulty logic, we actually just increase the count and rate limit for another `stake_interval` blocks.

ales-otf:
## Description

- Fixes the issue, when the block of the last stake/unstake was reset to the current block.

- Removes `get_stakes_this_interval_for_coldkey_hotkey` and `set_stakes_this_interval_for_coldkey_hotkey` by 
  replacing it with `try_increase_staking_counter`.

  For a getter, `get_stakes_this_interval_for_coldkey_hotkey` has an unwanted side effect of mutating the storage.

  `set_stakes_this_interval_for_coldkey_hotkey` is just a storage setter.

  Both methods used only in `do_add_stake` and `do_remove_stake`.

- Replaces `StakeRateLimitExceeded` and `UnstakeRateLimitExceeded` error variants with `StakingRateLimitExceeded` as we don't separate rate limit logic and value between staking/unstaking.

## Related Issue(s)

- Closes #1091

## Type of Change

- [ ] Documentation update

## Checklist

- [X] I have run `cargo fmt` and `cargo clippy` to ensure my code is formatted and linted correctly

### DIFF
--- a/pallets/subtensor/src/utils/misc.rs
+++ b/pallets/subtensor/src/utils/misc.rs
@@ -154,18 +154,36 @@ impl<T: Config> Pallet<T> {
             target_stakes_per_interval,
         ));
     }
-    pub fn set_stakes_this_interval_for_coldkey_hotkey(
+
+    // Counts staking events within the [`StakeInterval`]. It increases the counter by 1 in case no
+    // limit exceeded, otherwise returns an error.
+    pub(crate) fn try_increase_staking_counter(
         coldkey: &T::AccountId,
         hotkey: &T::AccountId,
-        stakes_this_interval: u64,
-        last_staked_block_number: u64,
-    ) {
-        TotalHotkeyColdkeyStakesThisInterval::<T>::insert(
-            coldkey,
-            hotkey,
-            (stakes_this_interval, last_staked_block_number),
+    ) -> DispatchResult {
+        let current_block = Self::get_current_block_as_u64();
+        let stake_interval = StakeInterval::<T>::get();
+        let stakes_limit = TargetStakesPerInterval::<T>::get();
+        let (stakes_count, last_staked_at) =
+            TotalHotkeyColdkeyStakesThisInterval::<T>::get(coldkey, hotkey);
+
+        if stakes_count == 0 || last_staked_at.saturating_add(stake_interval) <= current_block {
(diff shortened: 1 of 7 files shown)


pallets/subtensor/src/utils/misc.rs
```diff
@@ -154,18 +154,36 @@ impl<T: Config> Pallet<T> {
...
-            hotkey,
-            (stakes_this_interval, last_staked_block_number),
+    ) -> DispatchResult {
+        let current_block = Self::get_current_block_as_u64();
+        let stake_interval = StakeInterval::<T>::get();
+        let stakes_limit = TargetStakesPerInterval::<T>::get();
+        let (stakes_count, last_staked_at) =
+            TotalHotkeyColdkeyStakesThisInterval::<T>::get(coldkey, hotkey);
+
+        if stakes_count == 0 || last_staked_at.saturating_add(stake_interval) <= current_block {
```
camfairchild, line 171:
```suggestion
        // Reset staking counter if it's been stake_interval blocks since the first staking action of the series.
        if stakes_count == 0 || last_staked_at.saturating_add(stake_interval) <= current_block {
```

camfairchild (changes requested):

camfairchild (approved):

sam0x17 (approved):

PR #1103, opened 2024-12-18
Override childkey take to zero if sharing coldkey with parent

gztensor:
## Description
If parent and child are sharing the same coldkey, the childkey take is ignored and overridden with zero.

## Related Issue(s)

n/a

## Type of Change
- [ ] Documentation update

## Breaking Change
n/a

## Checklist

### DIFF
--- a/pallets/subtensor/src/tests/children.rs
+++ b/pallets/subtensor/src/tests/children.rs
@@ -3861,7 +3861,7 @@ fn test_childkey_take_same_coldkey() {
         step_block((hotkey_tempo * 2) as u16);
 
         // Verify how emission is split between keys
-        //   - Child stake is not increased
+        //   - Child stake is not increased (child & parent have same owner => no child-take)
         //   - Parent stake increased by 50% of total emission
         //   - Nominator stake increased by 50% of total emission
         let child_emission = crate::Stake::<Test>::get(child, coldkey);
--- a/pallets/subtensor/src/tests/children.rs
+++ b/pallets/subtensor/src/tests/children.rs
@@ -3300,7 +3300,8 @@ fn test_set_weights_no_parent() {
 #[test]
 fn test_childkey_take_drain() {
     new_test_ext(1).execute_with(|| {
-        let coldkey = U256::from(1);
+        let coldkey_parent = U256::from(1);
+        let coldkey_child = U256::from(5);
         let parent = U256::from(2);
         let child = U256::from(3);
         let nominator = U256::from(4);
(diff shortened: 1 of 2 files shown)


pallets/subtensor/src/tests/children.rs
```diff
@@ -3747,3 +3749,136 @@ fn test_do_set_child_cooldown_period() {
...
+
+        // Prevent further subnet epochs
+        crate::Tempo::<Test>::set(netuid, u16::MAX);
+        crate::Tempo::<Test>::set(root_id, u16::MAX);
+
+        // Run run_coinbase until PendingHotkeyEmission is drained for both child and parent
+        step_block((hotkey_tempo * 2) as u16);
+
+        // Verify how emission is split between keys
+        //   - Child stake is not increased
```
camfairchild:
```suggestion
        //   - Child stake is not increased (child & parent have same owner => no child-take)
```

camfairchild (dismissed):
little nit for a comment, but looks good

camfairchild (approved):

pallets/subtensor/src/tests/children.rs
```diff
@@ -3747,3 +3749,136 @@ fn test_do_set_child_cooldown_period() {
...
+        SubtensorModule::set_max_childkey_take(max_take);
+        assert_ok!(SubtensorModule::set_childkey_take(
+            RuntimeOrigin::signed(coldkey),
+            child,
+            netuid,
+            max_take
+        ));
+
+        // Set zero hotkey take for childkey
+        SubtensorModule::set_min_delegate_take(0);
```
ppolewicz, line 3802: In this test child take and childkey take should be non-zero and it should be observed that the parent didn't get a deduction from either and has extraced full dividend that was due. The current implementation doesn't have this bug, but the test should not assume that - one day someone might be refactoring the code and might make a mistake that this test would then not detect. Maybe just use 0.01, 0.02, 0.05 so that we are sure they have never been confused with each other.

pallets/subtensor/src/tests/children.rs
```diff
@@ -3437,7 +3438,8 @@ fn test_childkey_take_drain() {
 #[test]
 fn test_childkey_take_drain_validator_take() {
```
ppolewicz, line 3439: This test should be refactored into a helper function that would then be used in two tests:
- test_childkey_take_same_coldkey
- test_childkey_take_drain_validator_take

The helper function in question wouild accept parameters for `coldkey_parent` and `coldkey_child`, which would be the same in the first test and different in the other test. The other parameters of the helper would be:
- `expected_child_emission`
- `expected_parent_emission`
- `expected_nominator_emission`

This should save ~120 LOC and it will make maintenance of the test suite less costly as adjustments to the internal subtensor interface woudln't have to be replicated in two places, but only in one.

PR #1141, opened 2025-01-09
Fix/charge-for-swap-before-schedule

camfairchild:
## Description

Right now, we only try to charge the key swap cost *after* a coldkey is already scheduled to be swapped. This means the scheduled swap will *fail* if there is not enough balance in the coldkey for the cost.  

However, we still let them schedule the swap.

This PR checks for the cost upfront so we don't need to worry about this anymore. Also adds a test for preventing the schedule call if the cost can't be paid.

This PR *also* provides  the swap cost via the `swap_coldkey` call, so this can break any *existing* scheduled calls.  
This is meant to preserve the cost at the time of scheduling, while still charging if the swap_coldkey call is called directly.

## Related Issue(s)

- Closes #[issue number]

## Type of Change

- [ ] Documentation update

## Breaking Change

## Checklist


sam0x17:
clippy + test

sam0x17 (dismissed):

PR #1153, opened 2025-01-13
Feat/evm rao staking

### LINKED ISSUE #1104: [RAO] Update staking precompile to handle staking to subnets

gztensor:
## Description

Implement usage of netuid parameter in staking precompile in `add_stake` and `remove_stake`. 

## Related Issue(s)

- Closes #1104 

## Type of Change
- [ ] Documentation update

## Breaking Change

none

## Checklist

### DIFF
--- a/runtime/src/precompiles/staking.rs
+++ b/runtime/src/precompiles/staking.rs
@@ -119,7 +119,7 @@ impl StakingPrecompile {
     }
 
     fn parse_netuid(data: &[u8], offset: usize) -> Result<u16, PrecompileFailure> {
-        if data.len() < (offset + 2) as usize {
+        if data.len() < offset + 2 {
             return Err(PrecompileFailure::Error {
                 exit_status: ExitError::InvalidRange,
             });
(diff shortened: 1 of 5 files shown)


runtime/src/precompiles/staking.rs
```diff
@@ -68,9 +68,7 @@ impl StakingPrecompile {
     fn add_stake(handle: &mut impl PrecompileHandle, data: &[u8]) -> PrecompileResult {
         let hotkey = Self::parse_hotkey(data)?.into();
         let amount: U256 = handle.context().apparent_value;
-
-        // TODO: Use netuid method parameter here
-        let netuid: u16 = 0;
+        let netuid: u16 = Self::parse_netuid(data, 0x3E)?;
```
ales-otf: You don't have to clarify type parameter here as `parse_netuid` returns `u16`. But is it ok to parse `u16` here, while having `u256` in contract?

ales-otf (dismissed):

sam0x17 (dismissed):

runtime/src/precompiles/staking.rs
gztensor: - Removed type from netuid declaration.
- Yes, when parsing u16, I only use two bytes of passed value, so it wouldn't panic. The RLP packs uint16 into 32 bytes, so to avoid ambiguity in the future, I made netuid parameter a uint256 in solidity.

ales-otf (dismissed):

sam0x17 (approved):

PR #1162, opened 2025-01-15
Feat/no more weird runtime api encoding

camfairchild:
## Description

This PR changes the runtime APIs to only use the actual types they are expecting, instead of encoding/decoding to `Vec<u8>`.
This means we can decode runtime API results using just the chain metadata, so we can more easily support future changes to the APIs.
This breaks old clients, but the timing with RAO makes it worthwhile, as old clients will break anyway.

## Type of Change

- [ ] Documentation update

## Breaking Change

Clients upgrading should be grabbing their type information from the V15 Metadata on the chain. 
They should also be able to see these runtime APIs listed there, with their expected input and output types, and the encoding for said types.

## Checklist

### DIFF
--- a/pallets/subtensor/src/lib.rs
+++ b/pallets/subtensor/src/lib.rs
@@ -234,6 +234,11 @@ pub mod pallet {
     }
     #[pallet::type_value]
     /// Default value for zero.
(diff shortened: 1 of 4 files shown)


camfairchild:
We'll hold on merging this until bittensor/btcli publish the impl

PR #1165, opened 2025-01-16
add fix_rust.sh script

sam0x17:
Adds a script that checks all failing CI lints and makes a commit with auto fixes for each one. Also makes a commit for uncommitted Cargo.lock changes if applicable.

### DIFF
--- a/scripts/fix_rust.sh
+++ b/scripts/fix_rust.sh
@@ -6,7 +6,7 @@ set -e  # Exit immediately if a command exits with a non-zero status.
 commit_if_changes() {
     if [ -n "$(git status --porcelain)" ]; then
         echo "changes detected, committing..."
-        git commit --allow-empty -am "$1"
+        git commit -am "$1"
         echo "commit created."
     fi
 }
--- a/scripts/fix_rust.sh
+++ b/scripts/fix_rust.sh
@@ -1,0 +1,28 @@
+#!/bin/sh
+
+set -e  # Exit immediately if a command exits with a non-zero status.
+
+# Function to check for git changes and commit if necessary.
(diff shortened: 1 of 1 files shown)


scripts/fix_rust.sh
```diff
@@ -0,0 +1,28 @@
+#!/bin/sh
+
+set -e  # Exit immediately if a command exits with a non-zero status.
+
+# Function to check for git changes and commit if necessary.
+commit_if_changes() {
+    if [ -n "$(git status --porcelain)" ]; then
+        echo "changes detected, committing..."
+        git commit --allow-empty -am "$1"
```
ales-otf: Doesn't `allow-empty` requires you to force push? If that is the case, you can force push some changes, which are not supposed to be pushed, like, rewriting the history on remote branch in case multiple people working on it - like we had recently on `feat/rao-*`. If it requires a force push, it would be better to cancel the script and tell to manually commit the changes (and then probably it makes sense trying to push them to detect changes on the remote branch). But given that commit happens only in the case there are changes, I suppose you can omit using `--allow-empty`.

ales-otf (dismissed):

scripts/fix_rust.sh
ales-otf: I mean force push in case of an empty commit.

sam0x17: --allow-empty just lets you make an empty commit, but you are right it isn't necessary here, since we know the commit will be non-empty, in fact that is the point. I just have muscle memory for doing `git commit --allow-empty -m` haha. Fixing

ales-otf (approved):

PR #1166, opened 2025-01-17
Fix TotalIssuance on migration and TotalStake check

ales-otf:
## Description

Fixes the issue we have with try-runtime. 

It splits the state checker into two parts:

- checking `TotalIssuance`
- checking `TotalStake`

First, there is a bug in "migrate rao". It was increasing the currency issuance, because it returned a part of lock to coldkey. But `TotalIssuance` storage hadn't been updated. We have `frame_config::Config::SingleBlockMigrations`, where `TotalIssuance` is recalculated, but it happens before pallet's migrations are triggered. Probably, we should refactor migrations to use `frame_config::Config::SingleBlockMigrations` instead to prevent such issues in the future.

Another issue was related to removing of `Stake` storage. Because of that, it's not possible to sum all the stakes to check `TotalStake`. From the chat:

> Regarding the try-runtime failure. I've fixed it for devnet. The 100 TAO difference was indeed caused by the root network, which wasn't initialized with the initial supply (judging by the genesis builder - let me know, if there are other places where it could be set). 
> 
> But with or without it, the things are more complicated on testnet and finney.
> 
> This is how pool_initial_tao is initialized (from do_register_network implementation on devnet-ready):
> 
> ```rust
>         let mut pool_initial_tao = POOL_INITIAL_TAO;
>         if pool_initial_tao > actual_tao_lock_amount {
>             pool_initial_tao = actual_tao_lock_amount;
>         }
>         if pool_initial_tao < 1 {
>             pool_initial_tao = 1;
>         }
> ```
> 
> As you can see, it can also be actual_tao_lock_amount and 1 (is 1 ok btw?). The exact value is not written into a separate store, so we can't determine what was the value for a particular subnet. I can be wrong in all of this, but if the only source of staked TAO per subnet is SubnetTAO store, we can't calculate the exact value of TotalStake to verify it. If all of this true, the only "fix" is to remove this check.

## Related Issue(s)

- Closes #[issue number]

## Type of Change

- [ ] Documentation update

## Checklist

## Additional Notes

### DIFF
--- a/pallets/subtensor/src/subnets/subnet.rs
+++ b/pallets/subtensor/src/subnets/subnet.rs
@@ -2,6 +2,8 @@ use super::*;
 use frame_support::IterableStorageMap;
 use sp_core::Get;
 
+pub(crate) const POOL_INITIAL_TAO: u64 = 100_000_000_000;
+
 impl<T: Config> Pallet<T> {
     /// Retrieves the unique identifier (UID) for the root network.
     ///
@@ -234,7 +236,7 @@ impl<T: Config> Pallet<T> {
         ); // Set subnet token symbol.
 
         // Put 100 TAO from lock into subnet TAO and produce numerically equal amount of Alpha
-        let mut pool_initial_tao = 100_000_000_000;
+        let mut pool_initial_tao = POOL_INITIAL_TAO;
         if pool_initial_tao > actual_tao_lock_amount {
             pool_initial_tao = actual_tao_lock_amount;
         }
--- a/pallets/subtensor/src/subnets/subnet.rs
+++ b/pallets/subtensor/src/subnets/subnet.rs
@@ -3,6 +3,7 @@ use frame_support::IterableStorageMap;
 use sp_core::Get;
 
 pub(crate) const POOL_INITIAL_TAO: u64 = 100_000_000_000;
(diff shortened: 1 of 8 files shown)


pallets/subtensor/src/subnets/subnet.rs
```diff
@@ -234,12 +237,12 @@ impl<T: Config> Pallet<T> {
...
         // Put 100 TAO from lock into subnet TAO and produce numerically equal amount of Alpha
-        let mut pool_initial_tao = 100_000_000_000;
+        let mut pool_initial_tao = POOL_INITIAL_TAO;
         if pool_initial_tao > actual_tao_lock_amount {
             pool_initial_tao = actual_tao_lock_amount;
         }
-        if pool_initial_tao < 1 {
-            pool_initial_tao = 1;
+        if pool_initial_tao == 0 {
+            pool_initial_tao = POOL_INITIAL_TAO_FALLBACK;
```
ales-otf: Comment from @gregzaitsev:

1 very unlikely to happen, but it is definitely too low, it breaks the liquidity pool. It's my code, but I maybe we can use a higher number just to stay safe.

ales-otf: Fixed on the base branch.

sam0x17 (approved):

JohnReedV (approved):

PR #1168, opened 2025-01-17
make docker image builds trigger on key branches + ability to manually trigger

sam0x17:
This makes it so the current state of `devnet-ready`, `devnet`, and `testnet` is continuously deployed to docker hub using the branch names as the tag name.

It also makes it so we can manually trigger branches or tags to deploy to docker hub based on a workflow dispatch. Currently it _does not_ allow for deploying a branch/tag with a different branch/tag name in docker hub, however this could be added if it is desired later.

### DIFF
--- a/.github/workflows/docker.yml
+++ b/.github/workflows/docker.yml
@@ -4,6 +4,12 @@ on:
   release:
     types: [published]
   workflow_dispatch:
+    inputs:
+      branch-or-tag:
+        description: "Branch or tag to use for the Docker image tag (optional)"
+        required: false
+        default: ""
+
   push:
     branches:
       - devnet-ready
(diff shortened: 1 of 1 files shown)


.github/workflows/docker.yml
```diff
@@ -4,6 +4,17 @@ on:
...
+      branch-or-tag:
+        description: "Branch or tag to use for the Docker image tag and ref to checkout (optional)"
+        required: false
+        default: ""
+
+  push:
+    branches:
+      - devnet-ready
+      - devnet
+      - testnet
```
keithtensor, line 17: not testing on main?

sam0x17: nope, for main people should use the releases, which also get deployed by this automatically

unconst (approved):

PR #1170, opened 2025-01-18
transfer stake (untested)

bdmason:
# Transfer stake

Adds a new `transfer_stake` extrinsic. Needs testing.


pallets/subtensor/src/staking/transfer_stake.rs
```diff
@@ -0,0 +1,89 @@
...
+
+        // --- 5. Unstake the amount of alpha from the origin coldkey
+        Self::decrease_stake_for_hotkey_and_coldkey_on_subnet(
+            &hotkey.clone(),
+            &coldkey.clone(),
+            netuid,
+            alpha_amount,
+        );
+
+        // --- 6. Unstake the amount of alpha for the destination coldkey
```
camfairchild:
```suggestion
        // --- 6. Stake the amount of alpha for the destination coldkey
```

pallets/subtensor/src/staking/transfer_stake.rs
```diff
@@ -0,0 +1,89 @@
...
+        ensure!(
+            Self::hotkey_account_exists(&hotkey),
+            Error::<T>::HotKeyAccountNotExists
+        );
+
+        // --- 4. Get the current alpha stake for the origin hotkey-coldkey pair in the origin subnet
+        let origin_alpha =
+            Self::get_stake_for_hotkey_and_coldkey_on_subnet(&hotkey, &coldkey, netuid);
+        ensure!(
+            alpha_amount <= origin_alpha,
```
camfairchild:
```suggestion          
        ensure!(
   Self::has_enough_stake_on_subnet(&hotkey, &coldkey, netuid, alpha_amount)
```

pallets/subtensor/src/staking/transfer_stake.rs
```diff
@@ -0,0 +1,89 @@
...
+            Error::<T>::NotEnoughStakeToWithdraw
+        );
+
+        let origin_tao: u64 = Self::swap_alpha_for_tao(netuid, alpha_amount);
```
camfairchild:
```suggestion
```

If we swap here, we also have to swap back, otherwise we mess up the accounting.

Also might be fine to not swap at all.

pallets/subtensor/src/staking/transfer_stake.rs
```diff
@@ -0,0 +1,89 @@
...
+        );
+
+        // --- 7. Log the event.
+        log::info!(
+            "StakeTransferred( coldkey:{:?}, origin_hotkey:{:?}, origin_netuid:{:?}, destination_hotkey:{:?}, destination_netuid:{:?} )",
+            coldkey.clone(),
+            destination_coldkey.clone(),
+            hotkey,
+            netuid.clone(),
+            origin_tao
```
camfairchild:
```suggestion
            "StakeTransferred( coldkey:{:?}, destination_coldkey:{:?}, hotkey:{:?}, netuid:{:?}, alpha:{:?} )",
            coldkey.clone(),
            destination_coldkey.clone(),
            hotkey,
            netuid.clone(),
            alpha_amount
```

pallets/subtensor/src/staking/transfer_stake.rs
```diff
@@ -0,0 +1,89 @@
...
+        );
+        Self::deposit_event(Event::StakeTransferred(
+            coldkey,
+            destination_coldkey,
+            hotkey,
+            netuid,
+            origin_tao,
```
camfairchild:
```suggestion
            alpha_amount,
```

camfairchild (changes requested):

pallets/subtensor/src/staking/transfer_stake.rs
bdmason: I prefer this, think I was just copying another event's arguments

bdmason: I removed the swap entirely

axionyu:
I have a suggestion here. I think the function can modified to be something like this for more flexibility when transferring stake. 

```
pub fn do_transfer_stake(
    origin: T::RuntimeOrigin,
    source_hotkey: T::AccountId,
    dest_hotkey: T::AccountId
    destination_coldkey: T::AccountId,
    netuid: u16,
    alpha_amount: u64,
) 
```

The `decrease_stake_for_hotkey_and_coldkey_on_subnet` and `increase_stake_for_hotkey_and_coldkey_on_subnet` call can be modified to look something like this.

```
        // --- 5. Unstake the amount of alpha from the origin coldkey
        Self::decrease_stake_for_hotkey_and_coldkey_on_subnet(
            &source_hotkey.clone(),
            &coldkey.clone(),
            netuid,
            alpha_amount,
        );

        // --- 6. Stake the amount of alpha for the destination coldkey
            &dest_hotkey.clone(),
            &destination_coldkey.clone(),
            netuid,
            alpha_amount,
        );
```

So this would support the following use cases:
1. User can transfer the alpha stake from their current coldkey to another coldkey for a given hotkey on a subnet (current use case)
2. User can transfer the alpha stake from a source hotkey to a destination hotkey. The stake will still be on the current coldkey (new use case) - For example, I want to transfer the stake from one validator to another validator in a single transaction.
3. User can transfer the alpha stake from a source hotkey from the current coldkey to a destination hotkey on another coldkey. (new use case) - For example, I want to transfer the stake to a validator from my current wallet to another validator in a new wallet.

camfairchild:
> I have a suggestion here. I think the function can modified to be something like this for more flexibility when transferring stake.

> So this would support the following use cases:
> 

>   2. User can transfer the alpha stake from a source hotkey to a destination hotkey. The stake will still be on the current coldkey (new use case) - For example, I want to transfer the stake from one validator to another validator in a single transaction.  

We already have a function for this `move_stake`.

>   3. User can transfer the alpha stake from a source hotkey from the current coldkey to a destination hotkey on another coldkey. (new use case) - For example, I want to transfer the stake to a validator from my current wallet to another validator in a new wallet.

I think we want to avoid this because we want to be able to enforce rate limits, etc.  


It's possible that we replace `move_stake` with this function, as you're suggesting, but, as with 3., we have to be careful of how we handle the staking rate limits. And we have to charge the tx fee potentially differently, otherwise people will be paying for `move_stake` when they currently do not. 

It's not a bad idea though as it makes things simpler from a user-perspective.

PR #1173, opened 2025-01-20
[RAO] put owner cut in a pending

camfairchild:
PR puts an Pending Owner cut on the epoch. previously we only gave one block of owner cut where we should've done much more.


sam0x17 (dismissed):

sam0x17:
but needs spec version bump

sam0x17 (approved):

PR #1191, opened 2025-01-25
Resolved an issue where the `get_metagraph` function was returning None instead of metagraph data when owner key identities were not set.

basfroman:
### Description
Resolved an issue where the `get_metagraph` function was returning None instead of metagraph data when owner key identities were not set.

## Type of Change
- [x] Bug fix


basfroman:
fixed here https://github.com/opentensor/subtensor/pull/1199

PR #1192, opened 2025-01-27
[RAO] add alpha transfer toggle

camfairchild:
Adds functionality to enable and a check for Alpha-transferability to be set by subnet owner *only*.  

Currently, it defaults to `false` and only allows enabling by SubnetOwner, i.e., not by Root.


unconst (dismissed):

camfairchild:
Closed for #1248

PR #1255, opened 2025-02-07
[RAO] add migration and set init min lock cost 1 tao

camfairchild:
This PR sets the min lock cost for networks to `1 TAO`, which will lower the initial network pools to `1 TAO` as well.   

The alternative is we set a new storage value just for the initial pool lock.


camfairchild:
replaced by #1259

PR #1349, opened 2025-02-27
fix docker not publishing "latest" tag

sam0x17:


sam0x17:
disclaimer: no way to test this really but in production, so yolo

ales-otf (approved):

PR #1386, opened 2025-03-10
Fix/staking fees

gztensor:
## Description
 
Introduce flexible staking fees. Fees will be equal to the hotkey staking rewards worth of one epoch for the amount being staked/unstaked.

## Type of Change
- [ ] Documentation update
- [x] Other (please describe):

## Checklist

### DIFF
--- a/pallets/subtensor/src/staking/add_stake.rs
+++ b/pallets/subtensor/src/staking/add_stake.rs
@@ -1,6 +1,4 @@
 use super::*;
-use safe_math::SafeDiv;
-use sp_core::Get;
 use substrate_fixed::types::I96F32;
 
 impl<T: Config> Pallet<T> {
--- a/pallets/subtensor/src/staking/add_stake.rs
+++ b/pallets/subtensor/src/staking/add_stake.rs
@@ -60,17 +60,16 @@ impl<T: Config> Pallet<T> {
         // 3. Ensure the remove operation from the coldkey is a success.
         let tao_staked: I96F32 =
             Self::remove_balance_from_coldkey_account(&coldkey, stake_to_be_added)?.into();
-        let alpha_estimate = tao_staked.saturating_mul(Self::get_alpha_price(netuid));
 
         // 4. Swap the stake into alpha on the subnet and increase counters.
         // Emit the staking event.
-
+        let fee = DefaultStakingFee::<T>::get();
         Self::stake_into_subnet(
             &hotkey,
             &coldkey,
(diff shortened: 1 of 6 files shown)


sam0x17 (approved):

camfairchild (approved):

pallets/subtensor/src/staking/add_stake.rs
```diff
@@ -58,13 +58,19 @@ impl<T: Config> Pallet<T> {
         )?;
 
         // 3. Ensure the remove operation from the coldkey is a success.
-        let tao_staked: u64 =
-            Self::remove_balance_from_coldkey_account(&coldkey, stake_to_be_added)?;
+        let tao_staked: I96F32 =
```
ales-otf, line 61: we don't need this conversion, if we don't use the fee calculation from that function.

pallets/subtensor/src/staking/add_stake.rs
```diff
@@ -148,12 +154,19 @@ impl<T: Config> Pallet<T> {
         }
 
         // 5. Ensure the remove operation from the coldkey is a success.
-        let tao_staked: u64 = Self::remove_balance_from_coldkey_account(&coldkey, possible_stake)?;
+        let tao_staked: I96F32 =
```
ales-otf, line 157: same thing about conversion here.

pallets/subtensor/src/staking/stake_utils.rs
```diff
@@ -1066,6 +1066,28 @@ impl<T: Config> Pallet<T> {
 
         Ok(())
     }
+
+    pub(crate) fn calculate_staking_fee(
+        netuid: u16,
+        hotkey: &T::AccountId,
+        alpha_estimate: I96F32,
+    ) -> u64 {
+        if (netuid == Self::get_root_netuid()) || (SubnetMechanism::<T>::get(netuid)) == 0 {
```
ales-otf, line 1075: parenthesis are unnecessary here. surprised it passed clippy.

pallets/subtensor/src/staking/stake_utils.rs
```diff
@@ -1066,6 +1066,28 @@ impl<T: Config> Pallet<T> {
...
+            DefaultStakingFee::<T>::get()
```
ales-otf, line 1076: just a personal preference, but it would be more readable with `return` statement here (guard style check), so you could avoid introducing `else` block.

PR #1508, opened 2025-04-04
Fix docker image overwrites by adding platform to image tag

shamil-gadelshin:
## Description
This PR alters the `docker.yml` by introducing additional tags related to the target platform when building docker images and pushing them to the repository.  The previous changes to the workflow split the multiplatform into two separate jobs; however, that caused image overwrites since the image tags were the same. The fix changes that and produces these results:

https://github.com/opentensor/subtensor/pkgs/container/subtensor/388517732?tag=ci-platform-fix-amd64
https://github.com/opentensor/subtensor/pkgs/container/subtensor/388656587?tag=ci-platform-fix-arm64

## Type of Change

- [ ] Documentation update
- [x] Other (CI workflow fix):

### DIFF
--- a/.github/workflows/docker.yml
+++ b/.github/workflows/docker.yml
@@ -66,8 +66,8 @@ jobs:
           push: true
           platforms: linux/amd64
           tags: |
-            ghcr.io/${{ github.repository }}:${{ env.tag }}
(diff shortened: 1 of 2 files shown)


sam0x17 (dismissed):
💯

sam0x17 (approved):

Issue #1558, opened 2025-04-21
Reimplement StakeDelta


camfairchild:
We need to delay the movement of stake. 
Currently, we only delay the movement of childkey stake, and by ~24h (7200 blocks).

We should also delay normal stakes using a similar mechanism.
Ideally, we modify the CHK mechanism to use the same implementation.

This would be possibly similar to the old `StakeDelta` mechanism, where we re-add stake to hotkeys it was removed from, and remove stake from hotkeys it was added to, with a delay to clear the *delta*.
See: #807

surcyf123:
why?

camfairchild:
> why?

Prevents consensus issues with large stake movements between hotkeys

PR #1600, opened 2025-05-06
Reimplement StakeDelta

shamil-gadelshin:
## Description
This PR reimplements "stake delta" feature that was removed previously. It saves stake movement into the separate storage by hotkey, coldkey, and netuid. During the `epoch()` function execution stake delta is used as a "damper" to counter large stake movement effects on the `epoch`.

## Related Issue(s)

- Closes https://github.com/opentensor/subtensor/issues/1558

## Type of Change

- [ ] Documentation update

## Breaking Change

The PR changes the `epoch` calculations.

## Checklist

## Additional Notes

The original PR comes as a draft to gather feedback.


pallets/subtensor/src/coinbase/run_coinbase.rs
```diff
@@ -511,7 +511,16 @@ impl<T: Config> Pallet<T> {
...
                 Self::get_stake_for_hotkey_on_subnet(hotkey, Self::get_root_netuid());
-            stake_map.insert(hotkey.clone(), (alpha_stake, root_stake));
+
+            let alpha_stake_delta = Self::get_stake_delta_for_hotkey_on_subnet(hotkey, netuid);
+            let root_stake_delta =
+                Self::get_stake_delta_for_hotkey_on_subnet(hotkey, Self::get_root_netuid());
+
+            let final_alpha_stake = alpha_stake.saturating_sub(alpha_stake_delta);
+
+            let final_toot_stake = root_stake.saturating_sub(root_stake_delta);
```
camfairchild, line 521:
```suggestion
            let final_root_stake = root_stake.saturating_sub(root_stake_delta);
```

pallets/subtensor/src/coinbase/run_coinbase.rs
```diff
@@ -511,7 +511,16 @@ impl<T: Config> Pallet<T> {
...
+
+            stake_map.insert(hotkey.clone(), (final_alpha_stake, final_toot_stake));
```
camfairchild:
```suggestion
            let final_root_stake = root_stake.saturating_sub(root_stake_delta);

            stake_map.insert(hotkey.clone(), (final_alpha_stake, final_root_stake));
```

pallets/subtensor/src/lib.rs
```diff
@@ -2646,3 +2667,40 @@ pub enum RateLimitKey {
...
+
+impl<AccountId: Ord + Clone> StakeDeltaMap<AccountId> {
+    pub fn new() -> Self {
+        StakeDeltaMap {
+            coldkey_stake_deltas: BTreeMap::new(),
+        }
+    }
+
+    pub fn total(&self) -> i128 {
+        self.coldkey_stake_deltas.values().sum()
```
camfairchild: We should consider if it helps the runtime to update this on every add/remove that way calling `total()` is faster.

Might be worse runtime depending on where we need to update it.

pallets/subtensor/src/staking/stake_utils.rs
```diff
@@ -243,17 +247,32 @@ impl<T: Config> Pallet<T> {
...
     /// * `u64`: The total inherited alpha for the hotkey on the subnet after considering the
     ///   stakes allocated to children and inherited from parents.
     ///
     /// # Note
     /// This function uses saturating arithmetic to prevent overflows.
-    pub fn get_tao_inherited_for_hotkey_on_subnet(hotkey: &T::AccountId, netuid: u16) -> u64 {
-        let initial_tao: U96F32 = U96F32::saturating_from_num(
-            Self::get_stake_for_hotkey_on_subnet(hotkey, Self::get_root_netuid()),
-        );
+    pub fn get_tao_inherited_for_hotkey_on_subnet(
```
camfairchild, line 258: Need to use `stake_delta` here: 
https://github.com/opentensor/subtensor/blob/ff5b3429788985e6e55e06b851743388d0b57243/pallets/subtensor/src/staking/stake_utils.rs#L317-L323

pallets/subtensor/src/staking/stake_utils.rs
```diff
@@ -344,6 +363,24 @@ impl<T: Config> Pallet<T> {
         finalized_tao.saturating_to_num::<u64>()
     }
 
+    pub fn get_stake_delta_for_hotkey_on_subnet(hotkey: &T::AccountId, netuid: u16) -> u64 {
+        let total_stake = StakeDeltaSinceLastEmissionDrain::<T>::get(netuid, hotkey).total();
+
+        total_stake.abs().saturated_into()
```
camfairchild: How do we know if the sum is positive or negative here?

I think we use it like `total_stake.max(0)`, so we discard the negatives. But `abs` isn't the same.

camfairchild (changes requested):

pallets/subtensor/src/staking/stake_utils.rs
shamil-gadelshin: If we remove the stake this epoch we will get a negative value. But we should mitigate the stake movement without taking the movement direction into account (add or remove), shouldn't we? We only dampen the movement. Please, correct me if I'm wrong.

pallets/subtensor/src/lib.rs
shamil-gadelshin: The data structure is not finalized (probably, we'll use StorageNMap). We'll reconsider the `total` operation after that.

pallets/subtensor/src/staking/stake_utils.rs
camfairchild: Here we return only an unsigned int so we can't tell if the delta was positive or negative.

We only truly care about positive deltas anyway (at least right now).

shamil-gadelshin: > Here we return only an unsigned int so we can't tell if the delta was positive or negative.

Yes, because I was going to reduce any movement effect ("punish any stake movement for the epoch"). I chose "abs" because otherwise "remove_stake" will create a negative value and we'll increase the emissions, which doesn't seem to be a desired outcome. 

> We only truly care about positive deltas anyway (at least right now).

Ok. I will go with `total_stake.max(0)` then. Thanks!

shamil-gadelshin:
The last changes address PR comments, introduce additional places for stake delta effects, add tests, and simplify stake delta storage map by removing the coldkey from the storage.

shamil-gadelshin:
We chose a different approach: https://github.com/opentensor/subtensor/pull/1731

PR #1606, opened 2025-05-07
Disallow subnet owners to set AdjustmentAlpha greater than 0.97

keithtensor:
## Related Issue(s)

- Closes #[issue number]

## Type of Change

- [ ] Documentation update

## Breaking Change

## Checklist

## Additional Notes


pallets/subtensor/src/migrations/migrate_reset_adjustment_alpha.rs
```diff
@@ -0,0 +1,59 @@
...
+
+    // ------------------------------
+    // Step 1: Reset all subnet's adjustment alpha to 0.5 if the value exceeds 0.5
+    // ------------------------------
+
+    let mut reset_entries_count = 0u64;
+
+    for netuid in AdjustmentAlpha::<T>::iter_keys() {
+        AdjustmentAlpha::<T>::mutate(netuid, |adjustment| {
+            *adjustment = (*adjustment).min(32768); // 0.5
```
surcyf123: This is .5 of u16, but adjustment alpha is a u64

surcyf123 (commented):
This is a big change... Most subnet owners have changed their adjustment alpha. Why are we limiting it to root only? Is this being abused?

pallets/admin-utils/src/lib.rs
```diff
@@ -374,7 +374,7 @@ pub mod pallet {
             netuid: u16,
             adjustment_alpha: u64,
         ) -> DispatchResult {
-            pallet_subtensor::Pallet::<T>::ensure_subnet_owner_or_root(origin, netuid)?;
+            ensure_root(origin)?;
```
distributedstatemachine: I dont believe this is what we discusssed on the call. 

It was to have a set the max to 0.5 and not only adjustabe by root, not have the entire extrinsic to be root only cc @ppolewicz

distributedstatemachine (changes requested):

evm-tests/test/subnet.precompile.hyperparameter.test.ts
```diff
@@ -163,26 +163,27 @@ describe("Test the Subnet precompile contract", () => {
     //     assert.equal(valueFromContract, onchainValue);
     // })
 
-    it("Can set adjustmentAlpha parameter", async () => {
+    // need sudo as origin now
```
distributedstatemachine: This diff becomes more nuanced

ts-paulswaim:
These types of changes should have their source and purpose documented in the origination comment. There's nothing stated as to the reason for the change, and it is contrary to the public discussion where this likely came from. Care should be taken when making params root only.

ppolewicz:
The change was supposed to:
- make it so that the sudo key can still set it to whatever it wants, 0 to 1
- make is so that the owner can only set it between 0 and 0.5
- in the migration reduce the current setting down to 0.5 if it's higher than 0.5

and the reason for the change is that multiple subnet owners are exploiting the alpha adjustment parameter to prevent the registration price from settling in at its market value, for example setting it to 1 when the price is high, freezing the price in place, circumventing the removal of the minimum price setting by the subnet owner.

distributedstatemachine (dismissed):

surcyf123:
@ppolewicz @keithtensor This doesn't make any sense to me. Adjustment alpha of .5 is worse at finding the market rate than some value like .8-.97. 

An adjustment alpha value of .5 will have 50% increases in price on full intervals. And this is what you want the max to be? This parameter was designed to help subnets find a market price, and a value of .97 was recommended. 

Even with a value of .8, the cost will increase by 20% in one interval if someone registers 3 keys and then drop 10% in the next when nobody registers. 

With constant 20% movements in price, it becomes a reg bot game to register at the lower 20%, and then nobody registers at the higher 20%, when the "market" price should be somewhere in the middle 

If you want to enforce a cap, I suggest the cap be changed to .97. There is a lot of value in increasing it above .5 to help the subnet find the market price if that's your goal. This PR does the opposite.

The only valid concern that you mentioned is if people lock the registration price at a value of 1 or a value very close to 1 to try to get the price not to move. A cap at .97 would fix that.

surcyf123:
Also, I second that we need more descriptive PR descriptions in the future. A change like this is a very big change to subnet owners and miners. Miners will need to start working on optimizing their reg bot scripts if this is merged. And no motivation behind it or the benefits of the PR were mentioned at all.

ppolewicz:
@surcyf123 aboslutely - we knew that it needs to be limited but we didn't know what the proper top value should be. Lets make it 0.97 as you say. There was a big discussion about it on opendev and it wasn't properly serialized into the ticket because the recording got corrupted, sorry about that :(

pallets/subtensor/src/migrations/migrate_reset_adjustment_alpha.rs
```diff
@@ -0,0 +1,59 @@
...
+
+    // ------------------------------
+    // Step 1: Reset all subnet's adjustment alpha to 0.5 if the value exceeds 0.5
+    // ------------------------------
+
+    let mut reset_entries_count = 0u64;
+
+    for netuid in AdjustmentAlpha::<T>::iter_keys() {
+        AdjustmentAlpha::<T>::mutate(netuid, |adjustment| {
+            *adjustment = (*adjustment).min(u64::MAX.saturating_div(2)); // 0.5
```
ppolewicz: please change to 0.97

pallets/subtensor/src/migrations/migrate_reset_adjustment_alpha.rs
```diff
@@ -0,0 +1,59 @@
...
+        return weight;
+    }
+
+    log::info!(
+        "Running migration '{}'",
+        String::from_utf8_lossy(&migration_name)
+    );
+
+    // ------------------------------
+    // Step 1: Reset all subnet's adjustment alpha to 0.5 if the value exceeds 0.5
```
ppolewicz: please change to 0.97

pallets/admin-utils/src/lib.rs
```diff
@@ -374,7 +376,14 @@ pub mod pallet {
             netuid: u16,
             adjustment_alpha: u64,
         ) -> DispatchResult {
-            pallet_subtensor::Pallet::<T>::ensure_subnet_owner_or_root(origin, netuid)?;
+            pallet_subtensor::Pallet::<T>::ensure_subnet_owner_or_root(origin.clone(), netuid)?;
+
+            if pallet_subtensor::Pallet::<T>::ensure_subnet_owner(origin, netuid).is_ok() {
+                ensure!(
+                    adjustment_alpha <= (u64::MAX.saturating_div(2)),
```
ppolewicz: please change to 0.97

ppolewicz:
@surcyf123 changed to 0.97 - the reason behind it is that we want to prevent subnet owners from hiking the price and setting alpha adjustment to 1 to bypass the min_burn setting restriction and make the price reflect its market value

If further refined this might end up consuming the entire incentive of all subnets with strong uid pressure where its miners will race to the bottom. It's not yet clear if we'll go on that road, but freezing the reg price is unacceptable.

open-junius (approved):

PR #1614, opened 2025-05-08
RPC to get validators

### LINKED ISSUE #1589: Get validators efficiently
I propose that we add a convenience function to return subnet validators, so that all miners of all service-based subnets can efficiently get the information they need about the validators in one go, without getting data for the other miners (nearly a thousand on sn1) unnecessarily.

I suggest to add it as a feature of the selective metagraph fetcher, which already is quite nice, flexible and efficient.

The new boolean flag `only_validators` would:
- check how many validators a subnet can have (it's a hyperparameter currently set to 64 on all subnets, but that might change in the future)
- check how much effective stake is required for a validator to be able to set weights (currently it's 1k alpha on mainnet, 10 alpha on testnet, but this can change)
- get a list of uids which meet the minimum stake requirement and have vpermit set to `true`
- sort it by stake
- leave only as many as the hyperparameter specifies that this subnet supports
- return the sorted result

Bittensor can survive without this, but clients will need to do this work on their side and either they won't do it properly and we'll have bugs or they will do it well and we'll have extra load on subtensors that can be avoided.

open-junius:
## Related Issue(s)

- Closes #1589 

## Type of Change

- [ ] Documentation update

## Breaking Change

## Checklist

## Additional Notes


pallets/subtensor/src/rpc_info/metagraph.rs
```diff
@@ -1356,13 +1369,52 @@ impl<T: Config> Pallet<T> {
...
-            None => SelectiveMetagraph {
+            _ => SelectiveMetagraph {
                 // Subnet index
                 netuid: netuid.into(),
                 ..Default::default()
             },
         }
     }
+
+    fn get_valiators(netuid: u16) -> SelectiveMetagraph<T::AccountId> {
```
camfairchild:
```suggestion
    fn get_validators(netuid: u16) -> SelectiveMetagraph<T::AccountId> {
```

pallets/subtensor/src/rpc_info/metagraph.rs
```diff
@@ -791,10 +800,14 @@ impl<T: Config> Pallet<T> {
     pub fn get_selective_metagraph(
         netuid: u16,
         metagraph_indexes: Vec<u16>,
+        validator_only: bool,
     ) -> Option<SelectiveMetagraph<T::AccountId>> {
         if !Self::if_subnet_exist(netuid) {
             None
         } else {
+            if validator_only {
+                return Some(Self::get_valiators(netuid));
```
camfairchild:
```suggestion
                return Some(Self::get_validators(netuid));
```

camfairchild (changes requested):

pallets/subtensor/src/rpc_info/metagraph.rs
open-junius: will update it.

pallets/subtensor/src/rpc_info/metagraph.rs
open-junius: will update it.

pallets/subtensor/src/rpc_info/metagraph.rs
```diff
@@ -1356,13 +1369,52 @@ impl<T: Config> Pallet<T> {
                     ..Default::default()
                 }
             }
-            None => SelectiveMetagraph {
+            _ => SelectiveMetagraph {
```
keithtensor: This means if the `SelectiveMetagraphIndex` is `Validators`, we'd still be going to this match arm. Is this correct?

open-junius: Yes, it is correct. get validators is a separate branch. that's the validator_only parameter means.

keithtensor: But is there a case where the `validators_only` parameter is false, and yet the `SelectedMetagraphIndex` is `Validators`?

open-junius: if validators_only is false, we don't return validators in the result. The index is wield, let me check how to change it. thanks

open-junius: removed the validators index.

basfroman:
Am I right in understanding that if we pass the argument `validator_only=True`, we get the same `SelectiveMetagraph` but only with the filled field `validators=(1, 0)` (instead of validators=None`) which means that if index `validators[0]=1` then the key with index `0` belongs to the `validator`, otherwise it does not (like `validators[1]=0`)?

If so, then why don't we just implement this behavior by default, when the `SelectiveMetagraph.validators` field always returns key indices with values ​​1 or 0 meaning whether the key is a validator or not?

pallets/subtensor/src/rpc_info/metagraph.rs
camfairchild: should we not use the index of `ValidatorsOnly` to do this and not add a boolean to the api?

e.g. remove the `only_validators` param. Then, if `ValidatorsOnly` is present just keys that are validators. 

This requires we add a `uids` field to the struct also

basfroman: I'd say this impl can be like this:
- we add `SelectedMetagraphIndex::Validators`. Then it gonna handled the same way as other `SelectedMetagraphIndex` indices.
- we add `validators` field into `Metagraph` struct.

Result:
- `get_metagraph_info` always has filed `Metagraph::validators` field
- `SelectiveMetagraph` returns object with filed `validators` as Vec with related HK indexes if `SelectedMetagraphIndex::Validators` passed in params as item in `field_indices`
- SDK `MetagraphInfo` class will get a new field `validators`

This is gonnabe clearest way for this and SDK implementation as far as I see. 
Pleas feel free to correct me if I'm wrong.

open-junius: Agreed, we can handle the validators the same as other items in metagraph. user can get any data according to passed index list.

keithtensor (dismissed):

basfroman (dismissed):
LGTM for SDK
Tested with PR https://github.com/opentensor/bittensor/pull/2887 locally

sam0x17 (approved):

shamil-gadelshin (approved):

PR #1629, opened 2025-05-13
Ensure stake threshold is respected when handing out vpermits

### LINKED ISSUE #1058: Validator Permit should only be given if you also have the min_stake

Currently, the vpermit process does not account for the min_stake needed to set_weights. Meaning, a peer can both have a vpermit and not have enough stake to act as a validator, making the metagraph inconsistent. 

The solution is to check for min_stake when we set the vpermits.

_No response_

### Additional context

_No response_

keithtensor:
Fixes #1058.

### DIFF
--- a/pallets/subtensor/src/epoch/math.rs
+++ b/pallets/subtensor/src/epoch/math.rs
@@ -246,6 +246,22 @@ pub fn is_topk(vector: &[I32F32], k: usize) -> Vec<bool> {
     result
 }
 
+// Returns a bool vector where an item is true if the vector item is in topk values and is non-zero.
+#[allow(dead_code, clippy::indexing_slicing)]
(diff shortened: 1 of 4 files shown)


shamil-gadelshin (dismissed):

keithtensor:
Looks like the fix is working _too_ well, to the point where if the stake of a validator is 0, it gets automatically pruned out to not be a validator, which makes testing a bit annoying. We'll have to set the stake of the validator to at least be 1 first.

camfairchild (approved):

camfairchild:
Kinda makes #1614 redundant

sam0x17 (approved):

basfroman:
SDK e2e test fixed https://github.com/opentensor/bittensor/pull/2884

Issue #1567, opened 2025-04-24
Clear small stake values.


gztensor:
Roughly 50% of the entries in Alpha map describe stake that is below 10 rao (in alpha units). Agenda for this issue:

1. Write a migration that clears these entries (possibly it may take multiple blocks to execute).
2. Find out why we have entries below the minimum stake and fix it

PR #1635, opened 2025-05-14
Clean small values from Alpha map.

shamil-gadelshin:
## Description
This PR adds a process that iterates through small number (10) of Alpha map entries each block and cleans values less than 10 RAO.

## Related Issue(s)

- Closes https://github.com/opentensor/subtensor/issues/1567

## Type of Change

- [ ] Documentation update

## Breaking Change

This PR clears alpha stake values. We need to be sure that we chose the constants correctly.

## Checklist

### DIFF
--- a/pallets/subtensor/src/staking/helpers.rs
+++ b/pallets/subtensor/src/staking/helpers.rs
@@ -196,6 +196,54 @@ impl<T: Config> Pallet<T> {
         }
     }
 
+    /// The function clears Alpha map in batches. Each run will check ALPHA_MAP_CLEAN_BATCH_SIZE
+    /// alphas. It keeps the alpha value stored when it's >= than MIN_ALPHA.
+    /// The function uses AlphaMapCleanLastKey as a storage for key iterator between runs.
+    pub fn clear_small_nominations_batch() {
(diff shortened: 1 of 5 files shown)


pallets/subtensor/src/staking/helpers.rs
```diff
@@ -196,6 +196,54 @@ impl<T: Config> Pallet<T> {
...
+            }
+
+            let mut new_starting_key = None;
+            let new_iteration = keys.len() < ALPHA_MAP_CLEAN_BATCH_SIZE;
+
+            // Check and remove alphas if necessary
+            for key in keys {
+                let alpha = Alpha::<T>::get(&key);
+                if alpha < U64F64::saturating_from_num(MIN_ALPHA) {
+                    Alpha::<T>::remove(&key);
```
gztensor, line 233: In the ideal case when there are no rounding errors, the sum of all alpha for one hotkey will match `TotalHotkeyShares`, so `TotalHotkeyShares` should be also decreased here, but in practice we need to check (offline, not in the chain code) if the sum really matches.

gztensor (changes requested):

PR #1660, opened 2025-05-20
fix broken subtensor pallet on_runtime_upgrade()

sam0x17:
basically 2 problems:
this: https://github.com/opentensor/subtensor/actions/runs/15144171875/job/42575641270?pr=1634
and this: https://github.com/opentensor/subtensor/blob/devnet-ready/pallets/subtensor/src/macros/hooks.rs#L107


l0r1s:
For the try-runtime, there is something weird happening because the number of scrapped keys is 64963, which is 10x less than what was scrapped in previous try-runtime jobs

Issue #1509, opened 2025-04-04
Anti-MEV aggregation


gztensor:
### Is your feature request related to a problem? Please describe.

We are observing MEV type of activity when users execute some typical sandwich attacks that enclose a normal stake or unstake operation between two malicious transactions that elevate/drop price within the same block before the victim realizes it.

### Describe the solution you'd like

TL;DR: The solution is to aggregate all transactions within a block (or a short interval of blocks if needed) and reduce them to only one swap.

### Problem statement

1. See the proposed solution and research if a similar solution was ever implemented by anyone else before.
2. Implement (using 3rd party implementation is preferable if their license allows).

### Proposed solution

As user `add_stake` and `remove_stake` transactions are coming, instead of executing the swap right away (calling `swap_tao_for_alpha` and `swap_alpha_for_tao`), we should record all such orders until _settlement_.

Settlement occurs either in `on_finalize`, or in `on_initialize` in the next block (or one of the next blocks), or at any other time as requirements clarify over time. During the settlement: 

- Add all TAO and all Alpha.
- If current_price * {sum of alpha} > {sum of tao}, we have excess Alpha.
- If current_price * {sum of alpha} < {sum of tao}, we have excess TAO.
- If current_price * {sum of alpha} == {sum of tao}, there is no excess.
- Execute a single swap to exchange the excess add received value to {sum of tao} (in case of alpha excess) or to {sum of alpha} (in case of tao excess).
- Distribute {sum of alpha} to buyers proportionally to their contribution.
- Distribute {sum of tao} to sellers proportionally to their contribution.

Above was is a simplified version of algorithm that doesn't account for limit orders, but we have limit orders (`add_stake_limit` and `remove_stake_limit`) that allow (1) setting the limit price, and (2) specifying fill or kill flag. In order to account these, sort all orders by limit price from best to worst (with non-limit orders coming at the end of the list). Best price for buy order is lower, best price for sell order is higher. Then, if final price is beyond the limit price of the order, modify the order to either reduce amount (if it is limit order) or remove it completely if it is fill-or-kill type. Then recalculate the final price and continue moving down the list of orders until the final price is better or equal than the limit price.

PR #1666, opened 2025-05-21
Introduce stake lock after add_stake operation.

shamil-gadelshin:
## Description
This PR addresses the stake manipulation problem. It introduces a stake lock after each stake deposit operation (like `add_stake` or `add_stake_limit`). The lock is the block number after which stake removal is allowed. It calculates as "current_block + subnet tempo".

It relates to both MEV and "stake delta" problems. 

The PR removes the commented code related to https://github.com/opentensor/subtensor/pull/1641 in the first commit. 

## Related Issue(s)

- https://github.com/opentensor/subtensor/issues/1509
- https://github.com/opentensor/subtensor/issues/1558

## Type of Change

- [ ] Documentation update

## Breaking Change

Code that removes the stake immediately will stop working.

## Checklist


shamil-gadelshin:
cc @ppolewicz

pallets/subtensor/src/staking/stake_utils.rs
```diff
@@ -890,6 +894,28 @@ impl<T: Config> Pallet<T> {
...
+
+        StakeLocks::<T>::insert(hotkey, netuid, stake_lock);
+    }
+
+    pub fn ensure_stake_is_unlocked(hotkey: &T::AccountId, netuid: u16) -> Result<(), Error<T>> {
+        if !StakeLocks::<T>::contains_key(hotkey, netuid) {
+            return Ok(());
+        }
+
+        let stake_lock = StakeLocks::<T>::get(hotkey, netuid);
```
gztensor, line 911: I suggest that we remove the key here if lock is expired.

pallets/subtensor/src/lib.rs
```diff
@@ -1176,6 +1094,19 @@ pub mod pallet {
     pub type TokenSymbol<T: Config> =
         StorageMap<_, Identity, u16, Vec<u8>, ValueQuery, DefaultUnicodeVecU8<T>>;
 
+    #[pallet::storage]
+    /// DMAP ( hot, netuid ) --> lock until block number | Returns the block number of the stake lock
+    pub type StakeLocks<T: Config> = StorageDoubleMap<
```
gztensor, line 1099: The stake locks should be by hotkey, coldkey, and netuid. Otherwise, we will be locking all nominators (all coldkeys) as soon as one coldkey stakes to this hotkey and nobody will be able to unstake until the next epoch. 

Also, what happens if a coldkey has a 1000 TAO stake with expired lock and adds 1 TAO more? Should we keep them from unlocking only 1 TAO or the whole 1001 TAO amount?

gztensor (changes requested):

pallets/subtensor/src/lib.rs
shamil-gadelshin: > The stake locks should be by hotkey, coldkey, and netuid. Otherwise, we will be locking all nominators (all coldkeys) as soon as one coldkey stakes to this hotkey and nobody will be able to unstake until the next epoch.

Indeed, this is a mistake. FYI: this was the issue that _alpha_ mentioned to _const_. 

> Also, what happens if a coldkey has a 1000 TAO stake with expired lock and adds 1 TAO more? Should we keep them from unlocking only 1 TAO or the whole 1001 TAO amount?

I thought about that - IMO introducing complex lock math leads to a lot of confusion: add 100 TAO, add 10 TAO - remove 90 TAO. Can I remove my 20 remaining TAO? We need to introduce a special interface for locked TAO schedule etc. The current approach could be investigated using a simple RPC call for the current lock by (coldkey, hotkey, subnet) tuple.

shamil-gadelshin:
Update version is here: https://github.com/opentensor/subtensor/pull/1731

PR #1685, opened 2025-05-26
dTao-Compatible Subnet Deregistration

### LINKED ISSUE #1651: dTao Compatible Subnet Deregistration
# dTao-Compatible Subnet Deregistration

## Background

Since the dTao launch, our existing subnet deregistration (`dissolve_network` / `remove_network`) is no longer working correctly.

## Subnet Pruning
- Triggered when `SubnetLimit` is reached
- **Step 1:** Exclude subnets still within `NetworkImmunityPeriod`  
- **Step 2:** Among the rest, find the subnet with the lowest price EMA  
- **Step 3:** If multiple share the same emission, pick the one with the earliest registration timestamp  

## Network Dissolution 

- In `dissolve_network` / `remove_network`, perform full dTao cleanup:
  - Unwind liquidity providers   
  - Destroy all α-in and α-out stakes  
  - Distribute remaining Tao to α-out stakers pro-rata
  - **Adjust the owner’s returned lock cost** by subtracting the portion of total emissions the owner actually received (`owner_received_emission = E * get_float_subnet_owner_cut()`), so the final refund is `max(0, lock_cost - owner_received_emission)`.
- Maintain **root-only** access to direct calls for now

## Explicit Subnet Limit
Add new sudo hyperparameter `SubnetLimit` starting at `128`.
Immunity period changed to `864_000` blocks (4 months) from `1_296_000` (6 months)
`NetworkRateLimit` 2 -> 4 days, `28_800` blocks.
`LockCost` will begin at 2,000 Tao
Registrations can begin to occur 7 days after the deploy.

## High-Level Flow

```text
New Registration → check slot cap?
  ├─ No → register network, grant immunity
  └─ Yes → prune one subnet → deregister → register new
Deregistration (manual or pruning) → unwind LPs → destroy α-in/out → distribute Tao → remove network
```

JohnReedV:
## Description
dTao-Compatible Subnet Deregistration

## Related Issue(s)

- Closes #1651

## Type of Change

- [ ] Documentation update


mcjkula:
Currently the immunity period of the subnet's neurons is being called instead of the network immunity period, which you should be able to get with `Self::get_network_immunity_period();` (I think), as well you don't need to cast it as u64, because the method then already returns u64. @JohnReedV

pallets/subtensor/src/lib.rs
```diff
@@ -854,6 +854,12 @@ pub mod pallet {
         50400
     }
 
+    #[pallet::type_value]
+    /// Default value for subnet limit.
+    pub fn DefaultSubnetLimit<T: Config>() -> u16 {
+        256
```
creativeessence: 256 or 128?

creativeessence (changes requested):

pallets/subtensor/src/lib.rs
```diff
@@ -872,7 +872,7 @@ pub mod pallet {
     #[pallet::type_value]
     /// Default value for subnet limit.
     pub fn DefaultSubnetLimit<T: Config>() -> u16 {
-        128
+        148
```
mcjkula: Why increment by 20, and not something "cleaner" like 16 or 32 (that's just a preference though hahah)

mcjkula: Also what's the idea on the 20, is it arbitrarily picked or is there some idea behind it like anticipating a registration every X days, so with 20 slots to be filled it would give enough time until more subnets are out of immunity?

JohnReedV: It was arbitrarily picked and will likely change again. Also the `SubnetLimit` will be increased after the feature is released without a runtime upgrade as it is a sudo only hyperparameter. I suspect that the main bottleneck will still be the network registration rate limit which is currently set to 2 days.

sam0x17 (dismissed):

pallets/swap/src/pallet/impls.rs
```diff
@@ -1212,6 +1212,104 @@ impl<T: Config> Pallet<T> {
...
+            }
+
+            let protocol_account = Self::protocol_account_id();
+
+            // Non‑protocol first
+            to_close
+                .sort_by(|a, b| (a.owner == protocol_account).cmp(&(b.owner == protocol_account)));
+
+            for CloseItem { owner, pos_id } in to_close.into_iter() {
+                let rm = Self::do_remove_liquidity(netuid, &owner, pos_id)?;
```
gztensor: Would it make sense to ignore errors here since we’re in the loop? If reserves are ideal, errors will not happen, but we had a few small discrepancies in the state so liquidity removals may fail for some (larger) accounts while they will succeed for smaller ones.

pallets/swap/src/pallet/impls.rs
```diff
@@ -1212,6 +1212,104 @@ impl<T: Config> Pallet<T> {
...
+
+            return Ok(());
+        }
+
+        // V2 / non‑V3: ensure V3 residues are cleared (safe no‑ops).
+        let _ = Positions::<T>::clear_prefix((netuid,), u32::MAX, None);
+        let active_ticks: sp_std::vec::Vec<TickIndex> =
+            Ticks::<T>::iter_prefix(netuid).map(|(ti, _)| ti).collect();
+        for ti in active_ticks {
+            ActiveTickIndexManager::<T>::remove(netuid, ti);
```
gztensor, line 1291: I assume this is just a safety measure because liquidity removal also removes ticks.

gztensor (changes requested):

sam0x17 (approved):

pallets/subtensor/src/migrations/migrate_subnet_locked.rs
```diff
@@ -0,0 +1,118 @@
...
+        log::info!(
+            target: "runtime",
+            "Migration '{}' already run - skipping.",
+            String::from_utf8_lossy(&migration_name)
+        );
+        return weight;
+    }
+
+    // (netuid, locked_rao) pairs taken from the historical snapshot (block #4_828_623).
+    const SUBNET_LOCKED: &[(u16, u64)] = &[
```
mcjkula, line 23: Maybe I misunderstand the migration, but why again are we re-inserting lock costs into pre-DTAO subnets?

Their locked should be 0 because we have given them their lock cost already, so nothing to refund

PR #1720, opened 2025-06-05
feat: make `add_stake_limit` payable

### LINKED ISSUE #1719: `add_stake_limit` should be a `payable` function in the precompile
Currently, [`add_stake_limit`](https://github.com/opentensor/subtensor/blob/dc7eb20e7ffb6e7ced9af33c5f7a9e31936152e0/precompiles/src/staking.rs#L281-L282) is not a `payable` function in the  precompile, (unlike i.e. [`add_stake`](https://github.com/opentensor/subtensor/blob/dc7eb20e7ffb6e7ced9af33c5f7a9e31936152e0/precompiles/src/staking.rs#L83-L85)). This goes against convention. `add_stake_limit` should be a `payable` function.

servantofGodtheOne:
regarding #1719

### DIFF
--- a/precompiles/src/staking.rs
+++ b/precompiles/src/staking.rs
@@ -279,6 +279,7 @@ where
     }
 
     #[precompile::public("addStakeLimit(bytes32,uint256,uint256,bool,uint256)")]
+	#[precompile::payable]
     fn add_stake_limit(
         handle: &mut impl PrecompileHandle,
         address: H256,


open-junius:
Thanks for the fix. Could you add a test case in evm-test

open-junius:
As we discussed, the token transfer is not in evm layer. the token amount is the parameter in transaction, not in the msg.value. I think we can close it.

sy3am:
I'm going to go ahead and close this and open another PR for a newer version of the staking precompile to address #1719

Issue #1135, opened 2025-01-08
Locked stake, fixed epochs & fair validating rewards


bdmason:
### Validator rewards don't reflect validator performance

In the current runtime the movement of stake to and from a validator (or one of their children) inflates the rewards. In the dTao runtime the movement of stake has the opposite effect and deflates the rewards.

### Current Runtime

When a nominator adds stake their stake it is immediately available to the validator to earn rewards in subnet epochs (which are approx hourly), however it is considered non-viable for the nominator until the next hotkey epoch (daily). So for up to a whole day the stake of a new nominator will be earning rewards for a validator that will be shared between everyboth but themselves at the next hotkey drain. This causes nom/kt to be spiked as has been explained before when there was a very real case of a nom/kt of 3091 TAO for a small validator one day when a whale moved in.

Removing stake has the same effect too.

### dTao Runtime

As hotkey epochs have been removed, the concept of non viable stake has gone with it. Staking rate limits have also been removed, perhaps because slippage is now considered a big enough deterrent to frequent stake movements.

The effect of completely free movement of stake and it being viable for rewards at each subnet epoch is the opposite of the current runtime. Take this trivialised example:

1. A subnet epoch starts and a validator has a total of 10 Alpha staked in the subnet.
2. The epoch runs and the validator earns rewards based on a combination its performance & the 10 Alpha staked.
3. On the penultimate block of the subnet epoch a new staker adds 10 Alpha to the subnet.
4. The subnet epoch ends and the rewards that were based on the original 10 Alpha stake are now shared amongst the validator based on the final stake value of 20 Alpha.

In this example the new staker has suppressed the rewards of the validator by 50% for the subnet epoch. This is the opposite of the current runtime where the rewards are inflated by the new stake.

It may be the case that slippage makes this economically unviable, but it offers a way for people to intentionally supress a validators rewards on chain to make them less appealing to stakers. Much like the current runtime offers a way to boost a validators rewards by sacrificing your own stake.

### Solution

In the past having a lock up period to ensure fair rewards by fixing stake to a full epoch was dismissed as the freedom of movement was considered more important. But now with dTao the maximum epoch is roughly an hour, very little freedom of movement needs to be sacrificed to ensure fair rewards.

When `add_stake` and `remove_stake` are called add the stake into a pending state until the subnet epoch runs.

### Additional context

[...]

PR #1731, opened 2025-06-10
Add stake locks

shamil-gadelshin:
## Description

This PR introduces stake locks. 

This PR addresses the stake manipulation problem by introducing a stake lock after each stake deposit operation (like add_stake or add_stake_limit). The lock is the block number after which stake removal is allowed. It calculates as "current_block + subnet tempo".

This PR relates to both MEV and "stake delta" problems.

Affected extrinsics:
- add_stake
- add_stake_limit
- remove_stake
- remove_stake_limit
- unstake_all
- unstake_all_alpha
- move_stake
- swap_stake
- swap_stake_limit
- transfer_stake

All extrinsics that add stake to the destination netuid lock tuple (hotkey, coldkey, netuid) for the tempo period of the subnet. All extrinsics that remove stake check the stake lock first and fail with the `StakeLocked` error if the lock is active. The only exception is `transfer_stake` when the destination coldkey is different from the origin (to prevent putting locks on others).

## Related Issue(s)

- https://github.com/opentensor/subtensor/issues/1509
- https://github.com/opentensor/subtensor/issues/1558

## Type of Change

- [ ] Documentation update

## Breaking Change

The code that removes the stake immediately will stop working.

## Checklist

## Additional Notes

The additional weights fix commit will be likely required and added after the initial code review.

### DIFF
--- a/pallets/subtensor/src/lib.rs
+++ b/pallets/subtensor/src/lib.rs
@@ -825,6 +825,12 @@ pub mod pallet {
         50400
     }
 
+    #[pallet::type_value]
+    /// Default value for block number
+    pub fn DefaultBlockNumber<T: Config>() -> BlockNumberFor<T> {
+        0u32.into()
+    }
+
     #[pallet::storage]
     pub type MinActivityCutoff<T: Config> =
         StorageValue<_, u16, ValueQuery, DefaultMinActivityCutoff<T>>;
@@ -1101,6 +1107,20 @@ pub mod pallet {
     #[pallet::storage] // --- MAP ( netuid ) --> token_symbol | Returns the token symbol for a subnet.
     pub type TokenSymbol<T: Config> =
         StorageMap<_, Identity, u16, Vec<u8>, ValueQuery, DefaultUnicodeVecU8<T>>;
+
+    #[pallet::storage]
+    /// DMAP ( hot, netuid ) --> lock until block number | Returns the block number of the stake lock
+    pub type StakeLocks<T: Config> = StorageNMap<
+        _,
+        (
+            NMapKey<Blake2_128Concat, T::AccountId>, // hot
+            NMapKey<Blake2_128Concat, T::AccountId>, // cold
+            NMapKey<Identity, u16>,                  // subnet
+        ),
+        BlockNumberFor<T>,
+        ValueQuery,
+        DefaultBlockNumber<T>,
+    >;
 
     /// ============================
     /// ==== Global Parameters =====
--- a/pallets/subtensor/src/staking/move_stake.rs
+++ b/pallets/subtensor/src/staking/move_stake.rs
@@ -49,6 +49,7 @@ impl<T: Config> Pallet<T> {
             None,
             None,
             false,
+            true,
         )?;
 
         // Log the event.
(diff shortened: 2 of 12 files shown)


pallets/subtensor/src/lib.rs
```diff
@@ -825,6 +825,12 @@ pub mod pallet {
         50400
     }
 
+    #[pallet::type_value]
+    /// Default value for block number
+    pub fn DefaultBlockNumber<T: Config>() -> BlockNumberFor<T> {
+        0u32.into()
```
gztensor, line 831: To avoid confusion, please change to u64 because block numbers are u64 in subtensor.

pallets/subtensor/src/staking/move_stake.rs
```diff
@@ -119,6 +120,9 @@ impl<T: Config> Pallet<T> {
         // Ensure the extrinsic is signed by the origin_coldkey.
         let coldkey = ensure_signed(origin)?;
 
+        // Prevent DoS attacks
+        let set_lock = coldkey == destination_coldkey;
```
gztensor, line 124: Noo, the transfer stake functionality main use case is to transfer from one coldkey to another :)

gztensor (changes requested):

bdmason:
If we're going to lock stake up for a whole epoch, then it seems fair to remove the dynamic staking fee when unstaking. Currently on non root subnets the fee is equivalent to 1 epoch of staking rewards for the amount being unstaked. This should go back to the fixed 50k RAO as you can no longer manipulate an epoch of rewards by staking on the block before the step then leaving the block after.

gztensor:
> If we're going to lock stake up for a whole epoch, then it seems fair to remove the dynamic staking fee when unstaking. Currently on non root subnets the fee is equivalent to 1 epoch of staking rewards for the amount being unstaked. This should go back to the fixed 50k RAO as you can no longer manipulate an epoch of rewards by staking on the block before the step then leaving the block after.

@bdmason , true, but with swap v3 the dynamic fees will go away and will be replaced with flat rate swap fee (because there will be users providing liquidity who receive these fees).

RD4Fun:
> All extrinsics that remove stake check the stake lock first and fail with the StakeLocked error if the lock is active. The only exception is transfer_stake when the destination coldkey is different from the origin (to prevent putting locks on others).

What about a MEV bot running with two cold keys?
It now bypasses the stake lock by transferring in and transferring out.
imo `transfer_stake` needs a lock on the destination coldkey also.

bdmason:
I've been thinking about this a bit more... whilst I've been [wanting stake locks for a long time](https://github.com/opentensor/subtensor/issues/1135), Bittensor has moved on and become a marketplace. Have we considered the effect stake locks could have on integrations with exchanges e.g. Binance?

The functionality to stop sandwich attacks already exists in `add_stake_limit` and `remove_stake_limit`, you just need to understand how to set a very tight limit correctly, something we can help educate people to do.

Also, on the topic of sandwich attacks, the recent update to mainnet `add_stake_limit_aggregate` just adds a new type of sandwich attack, you monitor the pool for a submission of `add_stake_limit_aggregate`, you front run them with your own `add_stake_limit` call, you are 100% guaranteed to get in before them as they are pushed to the following block, then in the next block you call `unstake_all` and you are guaranteed to be executed after them as the aggregates are run at the start of the block.

Aren't we just complicating the network to fix a problem that was fixed a long time ago with limit prices?

shamil-gadelshin:
https://github.com/opentensor/subtensor/pull/1766

PR #1766, opened 2025-06-17
Stake operation limits

shamil-gadelshin:
This PR introduces staking operation rate limits. All operations that 'add stake' including moving stake operations set a "this block limit". It prevents any 'remove stake' operations including moving stake operations until the next block. The exception is transfer_stake that doesn't set the destination limit. All limits will be removed on each block. We're limiting the triplet (hot, cold, subnet).


pallets/subtensor/src/lib.rs
```diff
@@ -899,6 +899,12 @@ pub mod pallet {
     /// Ensures unique IDs for StakeJobs storage map
     pub type NextStakeJobId<T> = StorageValue<_, u64, ValueQuery, DefaultZeroU64<T>>;
 
+    #[pallet::type_value]
+    /// Default value for block number
+    pub fn DefaultBlockNumber<T: Config>() -> BlockNumberFor<T> {
```
bdmason: It doesn't functionally change anything, but you wouldn't need this if `StakingOperationRateLimiter` had a boolean as the value.

pallets/subtensor/src/macros/errors.rs
```diff
@@ -216,5 +216,7 @@ mod errors {
         ZeroMaxStakeAmount,
         /// Invalid netuid duplication
         SameNetuid,
+        /// Too frequent staking operations
+        StakingOperationRateExceeded,
```
bdmason: Perhaps call this `StakingOperationRateLimitExceeded`

bdmason (approved):
Looks good, comments are about superficial things.

pallets/subtensor/src/lib.rs
```diff
@@ -1671,6 +1671,20 @@ pub mod pallet {
         OptionQuery,
     >;
 
+    #[pallet::storage]
+    /// DMAP ( hot, cold, netuid ) --> rate limits for staking operations
+    /// Returns the block number of the last staking operation
```
camfairchild: Is this comment accurate? Looks like it returns a bool

pallets/subtensor/src/staking/remove_stake.rs
```diff
@@ -282,6 +282,7 @@ impl<T: Config> Pallet<T> {
             NetUid::ROOT,
             total_tao_unstaked,
             0, // no fee for restaking
+            true,
```
camfairchild: Can be false as it's into root

```suggestion
            false,
```

pallets/subtensor/src/macros/errors.rs
```diff
@@ -216,6 +216,8 @@ mod errors {
         ZeroMaxStakeAmount,
         /// Invalid netuid duplication
         SameNetuid,
+        /// Too frequent staking operations
+        StakingOperationRateLimitExceeded,
```
lorismoulin: It will shift the index of the error and mess with decoding. The error should be added at the end of the enum.

pallets/subtensor/src/macros/hooks.rs
```diff
@@ -44,7 +44,9 @@ mod hooks {
         // 	* 'n': (BlockNumberFor<T>):
         // 		- The number of the block we are finalizing.
         fn on_finalize(_block_number: BlockNumberFor<T>) {
-            // Self::do_on_finalize(block_number);
+            for _ in StakingOperationRateLimiter::<T>::drain() {
+                // Clear all entries each block
+            }
```
lorismoulin, line 49: Why do we have an empty loop here?

l0r1s (commented):
small changes needed

pallets/subtensor/src/macros/hooks.rs
shamil-gadelshin: `drain()` returns an iterator. It must be consumed.

l0r1s (dismissed):

sam0x17 (approved):

sirouk:
I think this same-block limitation is also applying to stake that is moved under the same triplet from the origin of spearate hotkeys and towards a validator hotkey for the purposes of staking.

I will test with `transfer_stake`

PR #1789, opened 2025-06-26
Commit-Reveal Weight Inheritance

### LINKED ISSUE #1787: Last Update  for Validators with CRV3 nulifies the deregistration check in the epoch
## Summary

A vulnerability exists in the commit-reveal weight mechanism that allows attackers to inherit weights intended for other neurons through strategic registration spam. Additionally, the commit-reveal implementation incorrectly updates the `last_update` timestamp during commit rather than reveal, which can bypass deregistration protection mechanisms.

### The Attack Vector

When using commit-reveal for weight updates with a short tempo (e.g., 4 blocks), an attacker can:

1. Monitor when validators commit weights for existing miners
2. Spam registrations to force deregistration of targeted miners (those with low pruning scores)
3. Acquire the deregistered miner's UID
4. Inherit the weights when validators reveal their commits

### Timeline Example

```
Block 1000: Validator commits weights for neurons [A, B, C] with UIDs [0, 1, 2]
Block 1001: Attacker begins registration spam
Block 1002: Neuron B (UID 1) is pruned due to low pruning score
Block 1003: Attacker registers and receives UID 1
Block 1004: Validator reveals weights - attacker's neuron inherits weight meant for B
```

### Premature `last_update` Setting

Both `do_commit_weights` and `do_commit_crv3_weights` incorrectly set the validator's `last_update` timestamp at commit time:

```rust
// In do_commit_weights (line 106) and do_commit_crv3_weights (line 296)
Self::set_last_update_for_uid(netuid, neuron_uid, commit_block);
```

### Current Protection Mechanisms

The epoch calculation in `run_epoch.rs` includes weight validation:

```rust
// Remove weights referring to deregistered neurons.
weights = vec_mask_sparse_matrix(
    &weights,
    &last_update,           // When validator last set weights
    &block_at_registration, // When each neuron was registered
    &|updated, registered| updated <= registered,
);
```

This removes weights where the validator's last update was before the target neuron's registration. However, this protection has limitations:

1. **Timing Gap**: With commit-reveal, `last_update` is set during commit, but weights aren't applied until reveal
2. **UID Reuse**: When neurons are deregistered, their UIDs are immediately reused by new registrations
3. **Window of Vulnerability**: Between reveal and the next epoch, incorrect weights may influence consensus

### Example Scenario Demonstrating the Issue

```
Block 1000: Validator commits weights (last_update = 1000)
Block 1002: Target neuron deregistered
Block 1003: Attacker registers, gets same UID (block_at_registration = 1003)
Block 1004: Validator reveals weights

In run_epoch:
- Validator's last_update = 1000
- Attacker's block_at_registration = 1003
- Check: 1000 <= 1003 is true, so weight is REMOVED (protection works)

BUT the validator appears "active" since block 1000, even if they never reveal!
```

### Root Cause

The fundamental issues are:

1. The reveal process (`do_reveal_weights`) doesn't validate that target UIDs still belong to the same neurons that existed during commit
2. The `replace_neuron` function preserves UIDs, allowing new neurons to inherit positions in weight matrices
3. `last_update` is set prematurely during commit instead of when weights are actually applied

## Impact

- **Consensus Manipulation**: Attackers can gain undeserved weights from validators
- **Activity Tracking Failure**: Validators can appear active without actually setting weights

### Solution 1: Fix `last_update` Timing (Critical)

Move the `last_update` setting from commit to reveal:

```rust
// In do_commit_weights and do_commit_crv3_weights
// REMOVE this line:
// Self::set_last_update_for_uid(netuid, neuron_uid, commit_block);

// In do_set_weights (called from reveal)
// Always update when weights are actually applied:
Self::set_last_update_for_uid(netuid, neuron_uid, current_block);
```

### Solution 2: Weight Invalidation on UID Reassignment

Track registration blocks during commit and validate during reveal:

```rust
// Storage for tracking target registrations at commit time
#[pallet::storage]
pub type CommitTargetRegistrations<T: Config> = StorageDoubleMap<
    _,
    Blake2_128Concat,
    u16, // netuid
    Blake2_128Concat,
    T::AccountId, // committer
    Vec<(u16, u64)>, // Vec of (uid, block_at_registration)
    ValueQuery,
>;

// During reveal, validate UIDs haven't been reassigned
for (uid, stored_block) in stored_registrations {
    let current_block = Self::get_neuron_block_at_registration(netuid, uid);
    ensure!(
        current_block == stored_block,
        Error::<T>::InvalidRevealTargetUIDReassigned
    );
}
```

### Solution 3: Include Target Hotkeys in Commit Hash

Modify the commit hash to include target hotkeys:

```rust
let provided_hash: H256 = BlakeTwo256::hash_of(&(
    who.clone(),
    netuid,
    uids.clone(),
    values.clone(),
    salt.clone(),
    version_key,
    target_hotkeys.clone(), // NEW: Include target identities
));
```

### Solution 4: Enhanced Neuron Replacement

Update `replace_neuron` to clear pending reveals and bonds:

```rust
pub fn replace_neuron(...) {
    // ... existing logic ...
    
    // Clear weights set BY this neuron
    Weights::<T>::remove(netuid, uid_to_replace);
    
    // Clear bonds TO this neuron from all validators
    for (i, bonds_vec) in Bonds::<T>::iter_prefix(netuid) {
        let filtered_bonds: Vec<(u16, u16)> = bonds_vec
            .into_iter()
            .filter(|(j, _)| *j != uid_to_replace)
            .collect();
        
        if filtered_bonds.is_empty() {
            Bonds::<T>::remove(netuid, i);
        } else {
            Bonds::<T>::insert(netuid, i, filtered_bonds);
        }
    }
    
    // Invalidate pending reveals targeting this UID
    Self::invalidate_pending_reveals_for_uid(netuid, uid_to_replace);
}
```

JohnReedV:
## Description
This upgrade makes the epoch ignore weights that point to a miner which **registered** during the **current** commit-reveal period when commit-reveal is enabled. The result is that *only* miners that were registered before the last reveal can receive incentive.

Also bounds `reveal_period` to [1, 100] to prevent owners from indefinitely excluding new miners via excessive values.

## Motivation  
Bad actors could **snipe a freshly-vacated UID** because unrevealed weights still target that UID, instantly collecting rank & rewards for work they never did.  Masking weights that reference de-registered miners closes this exploit and aligns rewards with actual effort.

## Related Issue(s)

- Closes #1787

## Type of Change

- [ ] Documentation update


pallets/subtensor/src/epoch/run_epoch.rs
```diff
@@ -577,6 +577,32 @@ impl<T: Config> Pallet<T> {
...
+            let current_epoch = Self::get_epoch_index(netuid, current_block);
+            let reveal_period = Self::get_reveal_period(netuid);
+
+            for (uid_i, _) in hotkeys.iter().enumerate() {
+                // Epoch in which this UID most recently registered.
+                let reg_epoch = Self::get_epoch_index(netuid, block_at_registration[uid_i]);
+
+                // Zero weights for miners during their registration period &
+                // the one after in case of unrevealed commits from deregistered
+                if current_epoch < reg_epoch.saturating_add(reveal_period.saturating_mul(2)) {
```
camfairchild: Would this allow subnet owners to exclude miners by changing the C/R period?

JohnReedV: Yes good point. Owners could set `reveal_peroid` to `U64::MAX` and exclude new miners forever. 

Given that currently the highest `reveal_period` is set to `20` on `Netuid: 39` a cap of `25-50` could be introduced.

RD4Fun: What about the edge case of the reveal period reducing in size? eg from 15 to 5

sam0x17 (approved):

pallets/subtensor/src/epoch/run_epoch.rs
```diff
@@ -577,6 +577,32 @@ impl<T: Config> Pallet<T> {
...
+        if Self::get_commit_reveal_weights_enabled(netuid) {
+            // Precompute safe blocks for all UIDs
+            let safe_blocks: Vec<u64> = block_at_registration
+                .iter()
+                .map(|reg_block| {
+                    let reg_epoch = Self::get_epoch_index(netuid, *reg_block);
+                    let safe_epoch =
+                        reg_epoch.saturating_add(Self::get_reveal_period(netuid).saturating_mul(2));
+
+                    Self::get_first_block_of_epoch(netuid, safe_epoch)
```
camfairchild, line 589: If the reveal period is not tied to the epoch timing, then the check here should be to get the next epoch from the block after the reveal period ends.

camfairchild: So maybe last block of epoch +1, or next epoch, with just the reveal period end block

camfairchild (changes requested):

pallets/subtensor/src/epoch/run_epoch.rs
JohnReedV: > the check here should be to get the next epoch from the block after the reveal period ends.

This is what the current implementation already does

* `safe_epoch = reg_epoch + (reveal_period * 2)` -> first block of that epoch is the block after the 2-period window. 
* The extra period shields against unrevealed commits made during the deregistration period that still target the vacated UID. Dropping it would remove that protection.
* `reveal_period` is in epochs.

PR #1814, opened 2025-07-03
Make liquidity events consistent

axionyu:
## Description
The goal of this fix is to make the liquidity events more consistent - `liquidity`, `tick_low` and `tick_high` will be emitted for LiquidityAdded (Mint), LiquidityRemoved (Burn) and LiquidityModified. This is to ensure that the event emission logic is similar to Uniswap V3 (https://docs.uniswap.org/contracts/v3/reference/core/interfaces/pool/IUniswapV3PoolEvents#burn)

Referencing the Burn and Mint event there.

Note: I have also moved the `UpdateLiquidityResult` from the `swap-interface` pallet to the `swap` pallet. The reason is that the `TickIndex` is in the `swap` pallet. If I add the `swap` pallet as a dependency, it would result in a circular dependency so thus I decided to move it.

## Related Issue(s)

- Closes #[issue number]

## Type of Change

- [x ] New feature (non-breaking change which adds functionality)
- [ ] Documentation update

## Breaking Change

## Checklist

- [x ] I have run `cargo fmt` and `cargo clippy` to ensure my code is formatted and linted correctly
- [x ] New and existing unit tests pass locally with my changes

## Additional Notes

### DIFF
--- a/pallets/swap-interface/src/lib.rs
+++ b/pallets/swap-interface/src/lib.rs
@@ -41,12 +41,3 @@ pub struct SwapResult {
     pub tao_reserve_delta: i64,
     pub alpha_reserve_delta: i64,
 }
-
-#[derive(Debug, PartialEq)]
-pub struct UpdateLiquidityResult {
-    pub tao: u64,
-    pub alpha: u64,
-    pub fee_tao: u64,
-    pub fee_alpha: u64,
-    pub removed: bool,
-}
--- a/pallets/swap/src/pallet/impls.rs
+++ b/pallets/swap/src/pallet/impls.rs
@@ -8,7 +8,7 @@ use sp_arithmetic::helpers_128bit;
 use sp_runtime::traits::AccountIdConversion;
 use substrate_fixed::types::{I64F64, U64F64, U96F32};
 use subtensor_runtime_common::{BalanceOps, NetUid, SubnetInfo};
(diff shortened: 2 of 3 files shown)


pallets/swap/src/tick.rs
```diff
@@ -174,6 +181,13 @@ impl Deref for TickIndex {
     }
 }
 
+impl TypeInfo for TickIndex {
```
ales-otf: why did you do this? we made this trick for `NetUid` to make it backward-compatible with the clients, who depended on `u16` for NetUid. but there is no reason to do this for `TickIndex`. also, there is no reason to introduce CompactAs implementation and especially with the target type `i32` until it's really necessary.

ales-otf (changes requested):

pallets/swap/src/tick.rs
axionyu: Without adding the TypeInfo realised this will cause the subxt rust client to return TickIndex instead of i32

Yes CompactAs is redundant so will remove it

ales-otf: yes, and this is expected, because it's `TickIndex` - not `i32`. `TypeInfo` should be derived as it was before this change

ales-otf (dismissed):

ales-otf (approved):

sam0x17 (approved):

PR #1832, opened 2025-07-08
fix: scaling values of alpha precompiles

kenobijon:
## Description
Alpha precompile was converting the U96F32 price directly to u64 without proper scaling, while the runtime API applies a scaling factor of 1_000_000_000.
Introduced fix to apply proper scaling to getAlphaPrice and getMovingAlphaPrice and pull raw storage value of tao weight. 

## Type of Change
- [ ] Documentation update

## Breaking Change

No breaking change.

## Checklist

### DIFF
--- a/precompiles/src/alpha.rs
+++ b/precompiles/src/alpha.rs
@@ -32,14 +32,16 @@ where
             <pallet_subtensor_swap::Pallet<R> as SwapHandler<R::AccountId>>::current_alpha_price(
                 netuid.into(),
             );
-        Ok(U256::from(price.saturating_to_num::<u64>()))
+        let price_scaled = price.saturating_mul(U96F32::saturating_from_num(1_000_000_000));
+        Ok(U256::from(price_scaled.saturating_to_num::<u64>()))
     }
 
     #[precompile::public("getMovingAlphaPrice(uint16)")]
     #[precompile::view]
     fn get_moving_alpha_price(_handle: &mut impl PrecompileHandle, netuid: u16) -> EvmResult<U256> {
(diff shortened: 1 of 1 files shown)


JohnReedV (commented):
The `Validate-Benchmarks` check will work after merging in the latest `devnet-ready` branch.

open-junius (approved):

precompiles/src/alpha.rs
```diff
@@ -32,14 +32,16 @@ where
             <pallet_subtensor_swap::Pallet<R> as SwapHandler<R::AccountId>>::current_alpha_price(
                 netuid.into(),
             );
-        Ok(U256::from(price.saturating_to_num::<u64>()))
+        let price_scaled = price.saturating_mul(U96F32::saturating_from_num(1_000_000_000));
+        Ok(U256::from(price_scaled.saturating_to_num::<u64>()))
```
lorismoulin, line 36: Could be nice to have the 1_000_000_000 set as the EVM_TO_SUBSTRATE_DECIMALS and set this constant in the runtime common instead of the main runtime file to be able to reuse it.

l0r1s (changes requested):
one change so we don't end up with magic constants across the codebase

camfairchild (approved):

precompiles/src/alpha.rs
gztensor: In this case it is not EVM to Substrate conversion, but rather price being U96F32 type internally and u64 in per-bil format externally.

gztensor (approved):

sam0x17 (approved):

precompiles/src/alpha.rs
kenobijon: what do u recommend I call the constant

PR #1859, opened 2025-07-16
turn on automatic const fns clippy lint

sam0x17:
Makes it so any function that _could_ be a const fn, clippy will try to make it into a const fn. Should be a good performance optimization in certain places.


gztensor (approved):

l0r1s (approved):

sam0x17:
never mind, this lint can't handle substrate

PR #1875, opened 2025-07-22
rework for pure proxy evm

### LINKED ISSUE #1714: [EVM] Expose Pure Proxy to EVM

We should have a precompile to creat Pure Proxies and some functions to use them, particularly for staking and transfers.

- `createPure()`
   - takes proxy deposit, returns new address 
    - Creates a sub account now owned by the contract (contract can make calls on behalf of the sub account)
- `proxy(real, call)` -> takes "real" address of the pure account (sub account) and a call to run as that pure-account
    - generalized for calling any call via the pure proxy
- [ ] staking precompile proxy calls 
    - E.g. `addStakeWithProxy(real, netuid, amount)
- [ ] transfer proxy calls

_No response_

### Additional context

_No response_

open-junius:
## Related Issue(s)

- Closes #1714 

## Type of Change

- [ ] Documentation update

## Breaking Change

## Checklist

## Additional Notes


sam0x17:
merge conflict

evm-tests/src/contracts/pureProxy.ts
```diff
@@ -0,0 +1,55 @@
...
+        "outputs": [
+            {
+                "internalType": "bytes32[]",
+                "name": "",
+                "type": "bytes32[]"
+            }
+        ],
+        "stateMutability": "view",
+        "type": "function"
+    },
```
camfairchild:
```suggestion
```
If the pure proxy is transferred then this command won't be useful

evm-tests/src/contracts/pureProxy.ts
```diff
@@ -0,0 +1,55 @@
...
+                "name": "call",
+                "type": "uint8[]"
+            }
+        ],
+        "name": "pureProxyCall",
+        "outputs": [],
+        "stateMutability": "nonpayable",
+        "type": "function"
+    }
+];
```
camfairchild, line 35: we also need calls for regular proxies. can we add a proxy precompile as well. it will need to use `proxy.proxy`

camfairchild, line 35: Another call to add here is `proxy.proxy(proxy.addProxy())` as a function, can just call it `callAddProxy` with args `real: pubkey, delegate: pubkey, proxyType: uint, delay: Option<uint>`

evm-tests/src/contracts/pureProxy.ts
```diff
@@ -0,0 +1,55 @@
...
+                "internalType": "uint8[]",
+                "name": "call",
+                "type": "uint8[]"
+            }
+        ],
+        "name": "pureProxyCall",
+        "outputs": [],
+        "stateMutability": "nonpayable",
+        "type": "function"
+    }
```
camfairchild, line 34:
```suggestion
```
This should be in the generic-proxy precompile

evm-tests/src/contracts/pureProxy.ts
```diff
@@ -0,0 +1,55 @@
+export const IPURE_PROXY_ADDRESS = "0x000000000000000000000000000000000000080b";
+
+export const IPureProxyABI = [
+    {
+        "inputs": [],
+        "name": "createPureProxy",
+        "outputs": [],
```
camfairchild:
```suggestion
        "outputs": [
            {
                "internalType": "bytes32[]",
                "name": "",
                "type": "bytes32[]"
             }
        ],
```
This should output the address/substrate-pubkey of the pure proxy created

pallets/proxy/src/lib.rs
```diff
@@ -538,6 +520,123 @@ pub mod pallet {
...
+				Pays::No => Weight::zero(),
+			};
+			let base_weight = T::WeightInfo::proxy(T::MaxProxies::get())
+				.saturating_add(T::DbWeight::get().reads_writes(3, 1));
+			(base_weight.saturating_add(inner_call_weight), di.class)
+		})]
+        pub fn evm_proxy(
+            origin: OriginFor<T>,
+            force_proxy_type: Option<T::ProxyType>,
+            call: Box<<T as Config>::RuntimeCall>,
```
camfairchild: I don't understand why we have this function. We should be using the proxy_pallet calls within the precompile as much as possible

pallets/proxy/src/lib.rs
```diff
@@ -538,6 +520,123 @@ pub mod pallet {
...
+            let who = ensure_signed(origin)?;
+
+            let owner: T::AccountId = T::AddressMapping::into_account_id(evm_address);
+
+            ensure!(who == owner, Error::<T>::OriginNotMatchMappedEVM);
+
+            ensure!(
+                EVMProxies::<T>::get(evm_address).len() < T::MaxProxies::get() as usize,
+                Error::<T>::TooMany
+            );
```
camfairchild: we shouldn't be storing these in another map. the proxy pallet should already be handling this

camfairchild (changes requested):

open-junius:
@camfairchild just remembered why I design the solution like this way.
1. extra map. for js application, it can get the event and get which proxy account created. then use it. but for solidity, we can't query it easily. the existed map with the key of proxy account. So we need a new extrinsic to do it.  
2. to avoid someone use the substrate account to add proxy for evm address that he doesn't own, so add cross check.
3.  the same for evm_proxy.

camfairchild:
> just remembered why I design the solution like this way.
> 
>     1. extra map. for js application, it can get the event and get which proxy account created. then use it. but for solidity, we can't query it easily. the existed map with the key of proxy account. So we need a new extrinsic to do it.

we can use a map to hold _*exactly one*_ address, and then we can check that when we return to the EVM.

>     2. to avoid someone use the substrate account to add proxy for evm address that he doesn't own, so add cross check.
> 
>     3. the same for evm_proxy.

I don't understand. Is the request not sent from the EVM using the address as the origin? 
How would another user send a request from the matching origin?

evm-tests/src/contracts/pureProxy.ts
camfairchild: this needs to have the `delegate` arg. The caller needs to specify the proxy account to call on-behalf of

open-junius: I will update the parameter name. the first parameter is delegate or proxy account. the "real" account should be the mapped SS58 account from msg.sender.

PR #1708, opened 2025-06-02
Permissioned Babe NPoS Runtime

liamaharon:
Phase 3 in https://github.com/opentensor/subtensor/issues/1887

Now that we have a Hybrid Node that supports seamless Aura to Babe upgrades and block imports, we are finally ready to enact the Aura PoA to Babe NPoS runtime upgrade.

This PR includes
- Changing the runtime from Aura PoA to Babe NPoS consensus.
- Logic constraining NPoS participants to a whitelisted set (maintained by root)
- A Babe NPoS migration which initializes Babe NPoS with authorities from the Aura PoA block which executed the runtime upgrade
- A new admin-util extrinsic allowing customization of the emission "cut" allocated to NPoS validator
- Change to `run_coinbase` to set aside NPoS validator emissions

I have tested upgrades with fresh and baedeker runtimes.

Steps for this PR
- [ ] Review by Greg & Shamil
- [ ] Stand up network we can test it with 3rd parties running validators

Once confirmed no issues with the above, we can merge.

PR #1876, opened 2025-07-22
Hybrid Consensus Node

liamaharon:
Phase 1 in https://github.com/opentensor/subtensor/issues/1887.

I added the full NPoS dependencies in this PR primarily to reduce future merge conflicts, which recently been a bit time consuming to resolve.

## Hybrid Node

The Hybrid Node handles producing, syncing and validating both Aura and Babe based blocks. It contains logic to dynamically switch between Aura and Babe based processing, depending on the specific block it needs to process.

#### Node Service

The core changes are in the node service logic. Consensus specific service logic was migrated to a new `ConsensusMechanism` trait, which I have written both Aura and Babe implementation for.

#### Node RPC

Node rpc logic was made generic across consensus mechanisms by allowing the caller to pass the Frontier Consensus Data Provider to use, and custom RPC modules to instantiate. 

#### CLI / Command

A new cli parameter `--initial-consensus` was added, which determines which `ConsensusMechanism` to first start the node in. The node will dynamically switch between `ConsensusMechanism`s at runtime as needed to continue block production / validation / import. 

## TODO
- [x] Fix compiling with `runtime-benchmarks` feature flag 
- [x] Check `ConditionalEVMBlockImport` is up to date


node/src/cli.rs
```diff
@@ -13,6 +22,13 @@ pub struct Cli {
     #[arg(long, value_enum, ignore_case = true)]
     pub sealing: Option<Sealing>,
 
+    /// Whether to try Aura or Babe consensus on first start.
+    ///
+    /// After starting, the consensus used by the node will automatically
+    /// switch to whatever is required to continue validating / syncing.
+    #[arg(long, value_enum, ignore_case = true, default_value_t=SupportedConsensusMechanism::default())]
+    pub initial_consensus: SupportedConsensusMechanism,
```
shamil-gadelshin, line 32: Why do we need this? We know that the first consensus is aura and the next one is  babe. When should we add the parameter to the node?

node/src/command.rs
```diff
@@ -230,36 +233,97 @@ pub fn run() -> sc_cli::Result<()> {
...
+fn run_babe(arg_matches: &ArgMatches) -> Result<(), sc_cli::Error> {
+    let cli = Cli::from_arg_matches(arg_matches).expect("Bad arg_matches");
+    let runner = cli.create_runner(&cli.run)?;
+    match runner.run_node_until_exit(|config| async move {
+        let config = customise_config(arg_matches, config);
+        service::build_full::<BabeConsensus>(config, cli.eth, cli.sealing, None).await
+    }) {
+        Ok(_) => Ok(()),
+        Err(e) => {
+            if e.to_string().contains("BabeApi") {
```
shamil-gadelshin: What's about the actual error type check here? Can we use that instead?

node/src/consensus_mechanism.rs
shamil-gadelshin, line 1: Empty file? Likely an artifact.

gztensor, line 1: Is this intentional?

node/src/command.rs
```diff
@@ -178,7 +181,7 @@ pub fn run() -> sc_cli::Result<()> {
                 } = crate::service::new_partial(
                     &config,
                     &cli.eth,
-                    crate::service::build_manual_seal_import_queue,
```
shamil-gadelshin, line 180: Do we need to support this?

node/src/consensus/aura_wrapped_import_queue.rs
```diff
@@ -0,0 +1,140 @@
...
+where
+    C: ProvideRuntimeApi<B> + Send + Sync + sc_client_api::backend::AuxStore,
+    C::Api: BlockBuilderApi<B> + AuraApi<B, AuthorityId> + ApiExt<B>,
+    CIDP: CreateInherentDataProviders<B, ()> + Send + Sync,
+    CIDP::InherentDataProviders: InherentDataProviderExt + Send + Sync,
+{
+    async fn verify(&self, block: BlockImportParams<B>) -> Result<BlockImportParams<B>, String> {
+        let number: NumberFor<B> = *block.post_header().number();
+        log::debug!("Verifying block: {:?}", number);
+        if is_babe_digest(block.header.digest()) {
```
shamil-gadelshin, line 77: What if I propagate the custom block to some node with this header? Can I shut it down?

node/src/consensus/aura_wrapped_import_queue.rs
```diff
@@ -0,0 +1,140 @@
...
+        verifier,
+        Box::new(params.block_import),
+        params.justification_import,
+        params.spawner,
+        params.registry,
+    ))
+}
+
+fn is_babe_digest(digest: &Digest) -> bool {
+    digest
```
shamil-gadelshin, line 132: Can we simplify with the `matches!` macro here?  Something like this? 
`digest.log(|d| matches!(d, DigestItem::PreRuntime(engine_id, _) if *engine_id == BABE_ENGINE_ID)).is_some()`

node/src/consensus/babe_consensus.rs
```diff
@@ -0,0 +1,231 @@
...
+        Ok(())
+    }
+
+    fn rpc_methods(
+        &self,
+        client: Arc<FullClient>,
+        keystore: KeystorePtr,
+        select_chain: FullSelectChain,
+    ) -> Result<Vec<Methods>, sc_service::Error> {
+        if let Some(ref babe_worker_handle) = self.babe_worker_handle {
```
shamil-gadelshin, line 214: Is there an option to create the import queue in `new()`, renaming `build_queue` to `get_queue`?. The error "queue was already used" seems easier to reason about than two errors for babe_link and babe_worker_handle.

node/src/command.rs
```diff
@@ -230,36 +233,97 @@ pub fn run() -> sc_cli::Result<()> {
...
+        service::build_full::<BabeConsensus>(config, cli.eth, cli.sealing, None).await
+    }) {
+        Ok(_) => Ok(()),
+        Err(e) => {
+            if e.to_string().contains("BabeApi") {
+                log::info!(
+                    "💡 Chain is using Aura consensus. Switching to Aura service until Babe block is detected.",
+                );
+                run_aura(arg_matches)
+            } else if e.to_string().contains("lock hold by current process") {
```
shamil-gadelshin: Similar one here. Can we use the actual error type here?

node/src/command.rs
```diff
@@ -230,36 +233,97 @@ pub fn run() -> sc_cli::Result<()> {
...
+                log::warn!("Failed to aquire DB lock, trying again in 1s...");
+                std::thread::sleep(std::time::Duration::from_secs(1));
+                return run_babe(arg_matches);
```
shamil-gadelshin: Stack overflow on long db lock here.

shamil-gadelshin: Why do we have this condition in the first place?

node/src/consensus/aura_consensus.rs
```diff
@@ -0,0 +1,218 @@
...
+        task_manager.spawn_essential_handle().spawn(
+        "babe-switch",
+        None,
+        Box::pin(async move {
+            let client = client_clone;
+            let triggered = triggered_clone;
+            loop {
+                // Check if the runtime is Babe once per block.
+                if let Ok(c) = sc_consensus_babe::configuration(&*client) {
+                    if !c.authorities.is_empty() {
```
shamil-gadelshin, line 195: Can we hardcode the exact block number to switch?

shamil-gadelshin (changes requested):
An interesting approach, you've created the consensus engine abstraction on top of the existing crates! I have several questions about it:

- Do you have the whole roadmap of the consensus change? It would be easier to know what part of the code is missing and what we expect in the next parts. 
- How do you test the consensus change? It seems it's not possible until the next part is merged. 
- Am I correct that if I propagate the custom babe block I can switch the consensus from a random node? 
- What do you think about the specific block number as an explicit consensus engine boundary? This way we'll get rid of the pending verification, string error as a marker, etc.

liamaharon:
Thank you @shamil-gadelshin for your detailed review!

> Do you have the whole roadmap of the consensus change? It would be easier to know what part of the code is missing and what we expect in the next parts.

Yes, you can view the full changes (including runtime upgrade + migration) here: https://github.com/opentensor/subtensor/pull/1708

tldr; next changes after merging this are 
1. enacting a runtime upgrade migrating existing PoA authorities to NPoS authorities, and do not allow changes in that set
2. slowly adding new whitelisted parties into that set
3. opening the set up to anyone to validate

> How do you test the consensus change? It seems it's not possible until the next part is merged.

You can test them on the `node-decentralization` branch, following the same steps I did here: https://discord.com/channels/1120750674595024897/1387662637843742760/1391994494714515588

If you're interested in it let me know and I'll send you exact commands I used to set it up :)  

> Am I correct that if I propagate the custom babe block I can switch the consensus from a random node?

This is an interesting edge case I have not considered! I think you are correct, I would need to do additional checks on the Babe block rather than switching straight to Babe consensus...

> What do you think about the specific block number as an explicit consensus engine boundary? This way we'll get rid of the pending verification, string error as a marker, etc.

Yeah this is something I considered. Initially I tried to keep away from needing to hardcode any block numbers, keeping this as close as I could to being a "normal" runtime upgrade. However you raise good points about verification. I'll have a think about it.

node/src/cli.rs
liamaharon: After the node is sync'd up the a Babe block, node operator wants initial consensus to be Babe. 

I added it as a debug utility more than anything. Not fussed if you prefer we remove it.

node/src/command.rs
liamaharon: AFAIK it's only used for runtime-benchmarks, which I kept. I will double check with Nucleus team.

node/src/consensus/aura_wrapped_import_queue.rs
liamaharon: Good question. I don't think it will shutdown, but it might cause some other issues. Let me work on something more robust 👍🏼

node/src/consensus/babe_consensus.rs
liamaharon: I tried to get this working for a while, but opted against it because it would require serious changes to `new_partial`, and what `new_partial` means. Because to create the import queue (and babe link) requires `BIQ` params, which are not known at the time of initializing the `ConsensusMechanism`. 

If we wanted to initialise it in `new` we'd need to create the `ConsensusMechanism` inside `new_partial`, but that's not always desired, e.g. when running benchmark commands https://github.com/opentensor/subtensor/blob/4789a88e2d41b221e3735daa08323dab8208601c/node/src/command.rs#L178-L185

If you have any ideas happy to hear them out though!

node/src/command.rs
liamaharon: Sometimes when the node switches from Aura to Babe, I observed the Aura service hold onto the DB lock for at most 1 second before releasing it allowing the Babe service to start.

If it's held onto for longer than that, something would be seriously wrong

node/src/consensus/aura_consensus.rs
liamaharon: I prefer to reduce "magic numbers" that have the possibility to be misconfigured where possible. However if there is benefit to switching on block number instead of this I'm open to it

node/src/consensus/aura_wrapped_import_queue.rs
liamaharon: So a solution I think will work here is rather than having a wrapped Aura import queue where the verifier pauses on encountering a Babe block, we need the verifier to verify the Babe block using a BabeVerifier, while also checking that the authorities in the Babe block match the authorities in the last Aura block. 

Unfortunately, the `BabeVerifier` in vanilla Polkadot SDK is not public outside of its crate, so we are unable to construct it. 

For the purposes of unblocking this PR, I am going to make this piece of code use the Aura verifier when it encounters a Babe block for now, and add a TODO comment. I will keep the existing logic on the `node-decentralization` branch, so we can still test stuff e2e.

Once we upgrade the Polkadot SDK to our fork we can make the `BabeVerifier` public and use it to ensure junk Babe blocks can't mess up the node when it's in Aura mode.

I also need to upgrade the Babe service verifier to verify when it gets a genesis block, the authorities which signed that genesis block are the exact same authorities who signed the last Aura block in the chain.

I will include all this stuff in a full roadmap I'm working on.

node/src/cli.rs
shamil-gadelshin: Please, remove it if you don't need it or add a comment about its temporary status (if you need it to debug future code).

shamil-gadelshin (commented):
Thanks for the changes and the roadmap. It's easier to understand the general direction now. 

My current concern is the moment of the consensus change. We don't need flexibility for this event because it seems like "one time unidirectional change". If the hardcoded magic const makes the code simpler, let's use it. If it introduces more issues than it solves, let's use your current method (presence of babe authorities) or a similar external marker.

gztensor (dismissed):

node/src/command.rs
liamaharon: Seems no reason to keep it.

shamil-gadelshin:
```
2025-07-28 10:51:15 panicked at ./subtensor/runtime/src/lib.rs:2421:13:
not implemented    
2025-07-28 10:51:15 💤 Idle (2 peers), best: #10 (0x85b3…3ec4), finalized #8 (0x7206…08d9), ⬇ 1.3kiB/s ⬆ 1.7kiB/s    
2025-07-28 10:51:16 panicked at ./subtensor/runtime/src/lib.rs:2421:13:
not implemented    
2025-07-28 10:51:16 💤 Idle (2 peers), best: #10 (0x85b3…3ec4), finalized #8 (0x7206…08d9), ⬇ 1.6kiB/s ⬆ 1.4kiB/s    
2025-07-28 10:51:16 panicked at ./subtensor/runtime/src/lib.rs:2421:13:
not implemented  
```

@gztensor saw this error after the runtime upgrade. It seems that "babe_switch" worker asks BabeApi and gets "not implemented" 

Do we have a fix for this one?

liamaharon:
> @gztensor saw this error after the runtime upgrade. It seems that "babe_switch" worker asks BabeApi and gets "not implemented"
>
> Do we have a fix for this one?

@shamil-gadelshin  thanks for giving it another try! It will panic if you try to use the Babe node with an Aura runtime. This branch does not include the change to the new Babe NPoS runtime, which I suspect is why you see that panic. 

Please try following my demo here https://discord.com/channels/1120750674595024897/1387662637843742760/1399902628309106822 where I test it on the `node-decentralization` branch. That branch includes both the hybrid node changes from this branch, as well as the new Babe NPoS runtime.

Let me know how you go!

shamil-gadelshin (dismissed):
Runtime upgrade works without bugs. Warp sync works as expected.

shamil-gadelshin (approved):

gztensor (approved):

sam0x17:
reverted until we can test on mainnet clone and a few other things

PR #1880, opened 2025-07-23
fix/CRV3 Improvements

JohnReedV:
## Description

This PR:
- Modifies the drand pallet to keep only the latest `216_000` (1 week) pulses.
- Drand pallet will now prune a maximum of `100` of the oldest pulses every block. 
- Hotkey is added to`WeightsTlockPayload` and the chain can reveal both payload versions (with and without hotkey).
- The hotkey included in the payload must match the commit extrinsic signer to successfully reveal. 
- Reveal round is now `n + reveal_periods` rather than `n * reveal_periods`.
- Reveals can now occur on any block inside the tempo before the block on which the epoch runs, rather than exclusively immediately before the epoch runs. The requirement is that it is the correct epoch to reveal & the reveal round has been ingested. 
- Adds `commit-reveal-version` to enforce client upgrades. 
- Rejects `reveal_round`s that are in the past via the signed extension.


HudsonGraeme (approved):
LGTM
One observation that `weights_version` might be better applied in the reveal round rather than immediately when the hparam is set.. But that's out of scope.

sam0x17 (approved):

gztensor (approved):

basfroman:
Broken test will be fixed in this PR https://github.com/opentensor/bittensor/pull/2999

PR #1888, opened 2025-07-28
Fix insufficient liquidity issue

gztensor:
## Description

Fix insufficient liquidity issue when liquidity is insufficient only on one side (TAO for add_stake and Alpha for remove_stake).

## Type of Change
- [ ] Documentation update

## Checklist

### DIFF
--- a/pallets/swap/src/pallet/impls.rs
+++ b/pallets/swap/src/pallet/impls.rs
@@ -465,12 +465,17 @@ impl<T: Config> Pallet<T> {
         limit_sqrt_price: SqrtPrice,
         drop_fees: bool,
     ) -> Result<SwapResult, Error<T>> {
-        ensure!(
-            T::SubnetInfo::tao_reserve(netuid.into()) >= T::MinimumReserve::get().get()
-                && u64::from(T::SubnetInfo::alpha_reserve(netuid.into()))
+        match order_type {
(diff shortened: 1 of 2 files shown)


sam0x17 (approved):

pallets/swap/src/pallet/impls.rs
```diff
@@ -465,12 +465,17 @@ impl<T: Config> Pallet<T> {
...
+        match order_type {
+            OrderType::Buy => ensure!(
+                u64::from(T::SubnetInfo::alpha_reserve(netuid.into()))
                     >= T::MinimumReserve::get().get(),
-            Error::<T>::ReservesTooLow
-        );
+                Error::<T>::ReservesTooLow
+            ),
+            OrderType::Sell => ensure!(
+                T::SubnetInfo::tao_reserve(netuid.into()) >= T::MinimumReserve::get().get(),
```
shamil-gadelshin, line 475: Is it OK that we have the same constant for both sides? Their exchange rate can differ in orders of magnitude.

gztensor: That's a very good concern - thank you! Actually, it is OK to have the same constant because their $- or TAO value is irrelevant, the number is dictated by math precision.

PR #1918, opened 2025-08-04
fix alpha_in emission calculations

### LINKED ISSUE #1909: Incorrect use of `tao_in_ratio` instead of `default_tao_in_i` for "price < TAO emissions"
### Describe the bug

The "price < TAO emissions" mechanism uses `tao_in_ratio` instead of `default_tao_in_i`, causing improper triggering post-halving.

**Example:**
- TAO emissions: 0.0589 (post-halving, 0.5 block emission)
- ALPHA price: 0.08 TAO

This clearly shows that the ALPHA price is higher than the TAO emissions (mechanism should not trigger), but when looking at the TAO emission ratio (0.0589/0.5) being 0.1178. The price is below it so the mechanism would get triggered.

- Expected ALPHA injected: 0.73625 (0.0589 / 0.08)
- Actual ALPHA injected: 1
- Resulting TAO per ALPHA: 0.0589

Instead of being 0.08 TAO per ALPHA (0.0589 / 0.73625) to keep the price stable, it would become 0.0589 TAO per ALPHA (0.0589 / 1) creating downward pressure.

### To Reproduce

N/A

### Expected behavior

The mechanism should use `default_tao_in_i` instead of `tao_in_ratio`, representing "price < TAO emissions" not "price < TAO emissions ratio (%)".

### Screenshots

N/A

### Environment

N/A

### Additional context

N/A

dougsillars:
## Description
This fix aims to fix an issue (#1909) - while also simplifying the codebase.

old line 87 is incorrect.  after the tao halving the alpha_in_i must max out at block_emission (0.5). If 1.0 alpha were emitted, the tao/alpha emission would not be even, causing a liquidity imbalance and a change in price.

modeled in a spreadsheet - at the tao halvings, an incorrect amount of alpha_in is emitted.
image
https://docs.google.com/spreadsheets/d/1i4tn0IYaVhvBu8xT1IjiP4U7uuFzsOdD4awjKWckH-k/edit?usp=sharing

we could have changed line 87 to 

```
alpha_in_i = block_emission;
```
to resolve this.

but there is an error in line 105 as well
```
alpha_in_i = tao_in_i.safe_div_or(price_i, alpha_emission_i);
```

This needs to read:
```
alpha_in_i = tao_in_i.safe_div_or(price_i, block_emission);
```
or on a divide by zero, we'll add too much alpha_in into the liquidity pool.

Taking a step back - lets look at line 105 in context of subsidy.
when the subsidy is in place:
`tao_in_i = price_i*block_emission`

so we can rewrite line 105 (removing the safe div for the example:
`alpha_in_i = (tao_in_i)/(price_i)`

to:
`alpha_in_i = (price_i*block_emission)/(price_i)`
canceling the `price_i`
`alpha_in_i = block_emission`

Since line 87 and line 105 both return the same value, 

I added line 107 to account for both options.

alpha_in_i = tao_in_i.safe_div_or(price_i, block_emission);

With the alpha_in_i calcualted in just one place - we simplify the code.

 This also resolves the issue of using alpha_emission_i instead of block_emission.

## Related Issue(s)

- Closes #1909 

## Type of Change

- [ x] Bug fix (non-breaking change which fixes an issue)
- [ ] Documentation update

## Breaking Change

no breaking change.

## Checklist

## Additional Notes

### DIFF
--- a/pallets/subtensor/src/coinbase/run_coinbase.rs
+++ b/pallets/subtensor/src/coinbase/run_coinbase.rs
@@ -84,7 +84,6 @@ impl<T: Config> Pallet<T> {
             );
             if price_i < tao_in_ratio {
                 tao_in_i = price_i.saturating_mul(U96F32::saturating_from_num(block_emission));
-                alpha_in_i = alpha_emission_i;
                 let difference_tao: U96F32 = default_tao_in_i.saturating_sub(tao_in_i);
                 // Difference becomes buy.
                 let buy_swap_result = Self::swap_tao_for_alpha(
@@ -102,9 +101,10 @@ impl<T: Config> Pallet<T> {
                 is_subsidized.insert(*netuid_i, true);
             } else {
                 tao_in_i = default_tao_in_i;
-                alpha_in_i = tao_in_i.safe_div_or(price_i, alpha_emission_i);
                 is_subsidized.insert(*netuid_i, false);
             }
+            
+            alpha_in_i = tao_in_i.safe_div_or(price_i, block_emission);
             log::debug!("alpha_in_i: {:?}", alpha_in_i);
 
             // Get alpha_out.


mcjkula:
> This also resolves the issue of using alpha_emission_i instead of block_emission.

Why is that an issue? We should use `alpha_emission_i` for this. I don't understand the fix you are applying.

The reason for the subsidy is, that when the price is lower than TAO emissions then the `alpha_in` becomes larger than allowed, and without the limitation to `alpha_emission_i` it could become "infinitely" large.

Let's say my ALPHA price is 0.2 and my TAO emissions are 0.4, then I would need to inject 2 ALPHA into the pool to keep the price stable. To resolve this issue the subsidy is clipping the `alpha_in_i` to `alpha_emission_i` this means more TAO per ALPHA, so that the price increases back towards it's TAO emissions and the pool can continue to inject in balance.

With V3 the change was done to fix this mechanism, where it would downscale `tao_in_i` to `price_i * block_emission` and use the `difference_tao` to swap TAO for ALPHA. This means it's the same process of bringing the price back towards it's TAO emissions.

The problem with the fix, was that it compares the price towards `tao_in_ratio` which is not supposed to be used for this.

Issue #1887, opened 2025-07-27
[Meta] NPoS + Babe Roadmap


liamaharon:
As suggested by @shamil-gadelshin, I'm creating a roadmap describing the pathway to a world with decentralised NPoS + Babe block authoring and consensus.

The roadmap is split into "phases", with the idea of merging progress incrementally as it is ready, rather than accumulating all changes into a huge PR which we try to merge all at once.

## Phase 1: Hybrid Consensus Node

Since we plan to switch from Aura consensus to Babe consensus mid-chain, we require a node which is capable of importing and producing both Babe and Aura blocks. This, I am calling a "Hybrid Node". I recorded a video explaining why we need a Hybrid Node, and demonstrating an implementation of it [here](https://drive.google.com/file/d/1UjiJS-6Y-U0hP0rpm9900FK4aImAs0Vo/view).

The Hybrid Node in phase 1 supports block production and validation of both Aura and Babe based runtimes. It does NOT yet support seemless switching of Aura -> Babe in the event of a runtime upgrade, which is achieved in phase 2.

I'd like to get this merged without the work planned for phase 2, just because it is a non-breaking change, and will save me lots of time resolving merge conflicts in the future :) 

- [x] Hybrid Node PR: https://github.com/opentensor/subtensor/pull/1876

## Phase 2: Seemless Aura -> Babe Upgrade

There are a few additional things required for the Hybrid Node from phase 1 to support seemless Aura -> Babe runtime upgrades, which requires a slightly adjusted Polkadot SDK:
- The vulnerability raised by @shamil-gadelshin [here](https://github.com/opentensor/subtensor/pull/1876#discussion_r2228259301) requires the SDK to allow constructing a `BabeVerifier` outside of the crate.
- We must replace hardcoded assumptions that a Babe "Genesis" can only occur at block 0 with something more flexible.
- If `pallet_babe::CurrentSlot != timestamp_slot`, it should log a warning rather than panic in the runtime. This is because I found initializing `pallet_babe::CurrentSlot` in the runtime upgrade brittle, and want to remove any potential panic vectors however small they are. Note that after the runtime upgrade, the brittleness disappears and `pallet_babe::CurrentSlot` always equals `timestamp_slot`.
- Fix a small annoyance when trying to initialise the node logger multiple times.

### TODOs
[...]

PR #1927, opened 2025-08-07
Node Support for Aura -> Babe Runtime Upgrades

liamaharon:
Phase 2 in https://github.com/opentensor/subtensor/issues/1887

Phase 2 additions:
- Applies the [OTF Patched Polkadot SDK](https://github.com/paritytech/polkadot-sdk/compare/polkadot-stable2503-6...opentensor:polkadot-sdk:polkadot-stable2503-6-otf-patches) 
  - See the comment above the Cargo.toml patch explaining how it is applied
  - This is critical for Aura to Babe migrations
  - Removes the need for the https://github.com/opentensor/grandpa.git patch
- Added code to automatically sync Aura keystore with the Babe keystore
- Add `-otf` suffix to subtensor crates that share the same name as a polkadot-sdk crate. This is to prevent name collisions when applying patches, and ensure we don't accidentally use the wrong crate.
- Fixed [this vulnerability](https://github.com/opentensor/subtensor/pull/1876#discussion_r2228259301) raised by @shamil-gadelshin
- Fixed edge case issue where node would not automatically switch from Babe to Aura service

## Update Aug 26

I discovered there is an issue warp syncing a chain that contains an Aura to Babe migration. The warp sync would succeed up until the first Babe block is mined, and then revert to a regular sync.

### Root Cause

Polkadot SDK does not allow starting a service with warp sync if the database is in a partially synced state: https://github.com/opentensor/polkadot-sdk/blob/d13f915d8a1f55af53fd51fdb4544c47badddc7e/substrate/client/network/sync/src/strategy/warp.rs#L235-L252

In the initial implementation of this PR, while a node is warp syncing, it will restart part-way through the sync when it detects the first Babe block.

At the time of the service restart, the node has a partially synced database. This causes the previously referenced code to terminate the warp sync, and fall back to a regular sync.

### Solution

To resolve this issue, we must warp sync the entire chain (Aura AND Babe blocks) without restarting the service.

This is achieved in two parts:

First, I prevented the Aura service switching to a Babe service while the node is syncing. This was achieved by adding new code here: https://github.com/opentensor/subtensor/blob/a2340960f46f33b7ab52657d9cc35df7b58427d0/node/src/consensus/aura_consensus.rs#L222

Second, I added support for the node to import Babe blocks while running an Aura service. This was achieved by replacing the `AuraWrappedImportQueue` with a `HybridImportQueue` that contains 
- A `HybridBlockImport` that contains inner full implementations of `AuraBlockImport` and `BabeBlockImport`
- A `HybridVerifier` that contains inner full implementations of the `AuraVerifier` and `BabeVerifier` 
- An `import_queue` function that builds an `ImportQueue` implementation capable of completely importing both Aura and Babe blocks.

The Aura service is required to construct a `BabeConfiguration` to pass to the hybrid `import_queue`, so it can import the first Babe block it encounters. This required me to pull in some Babe runtime configuration from https://github.com/opentensor/subtensor/pull/1708 into this PR, specifically the [`BABE_GENESIS_EPOCH_CONFIG`](https://github.com/opentensor/subtensor/blob/a2340960f46f33b7ab52657d9cc35df7b58427d0/runtime/src/lib.rs#L383-L387) and [`EPOCH_DURATION_IN_BLOCKS`](https://github.com/opentensor/subtensor/blob/a2340960f46f33b7ab52657d9cc35df7b58427d0/runtime/src/lib.rs#L368).

With these runtime constants, we are able to construct what our initial Babe configuration will be while running an Aura service: https://github.com/opentensor/subtensor/blob/a2340960f46f33b7ab52657d9cc35df7b58427d0/node/src/consensus/aura_consensus.rs#L249-L285

### Summary

With the two changes described in "Solution", when a node is warp syncing, it will warp sync the entire chain (all Aura and Babe blocks) and import the state entirely before it switches to running a Babe service. This resolves the issue root cause of the issue, which is that the node would restart mid-warp-sync.

It is important this phase is merged prior to phase 3, so node operators have time to upgrade in advance of the runtime upgrade. 

## Steps to simulate Aura -> Babe migration with finney state

1. Set up https://github.com/opentensor/baedeker-for-subtensor. Ask Greg if you have questions.
2. Build Babe NPoS runtime from https://github.com/opentensor/subtensor/pull/1708
```bash
$ git checkout node-decentralization
$ cargo b -r -p node-subtensor && cp ./target/release/wbuild/node-subtensor-runtime/node_subtensor_runtime.compact.compressed.wasm ./babe-npos.wasm
```
4. Build node from this branch
```bash
$ git checkout hybrid-node
$ rm ./target/release/node-subtensor && cargo b -r -p node-subtensor
```

6. Run Baedeker using node from this branch
```bash
$ cd ../baedeker-for-subtensor
$ ./localnet-baedeker.sh
```

8. Upgrade to Babe NPoS runtime. Ask Liam if you have questions.

## QA Checklist
- [x] `--initial-consensus aura` gracefully switches to Babe post-upgrade
- [x] `--initial-consensus babe` gracefully switches to Aura pre-upgrade
  - [x] `devnet-ready` state/runtime
  - [x] `hybrid-node` state/runtime
  - [x] `baedeker-finney` state/runtime
- [x] Babe upgrade works when enacted in early blocks
  - [x] `devnet-ready` state/runtime
  - [x] `hybrid-node` state/runtime
- [x] Babe upgrade works when enacted in later blocks (where era would be >1)
  - [x] `devnet-ready` state/runtime
  - [x] `hybrid-node` state/runtime
  - [x] `baedeker-finney` state/runtime
- [x] Warp sync Aura & Babe chain works when switch happened in early blocks
  - [x] `devnet-ready`
  - [x] `hybrid-node`
- [x] Warp sync Aura & Babe chain works when switch happened in later blocks (where era would be >1)
  - [x] `devnet-ready` state/runtime
  - [x] `hybrid-node` state/runtime
  - [x] `baedeker-finney` state/runtime


node/src/consensus/aura_wrapped_import_queue.rs
```diff
@@ -47,44 +62,104 @@ impl<B: BlockT, C, CIDP, N> AuraWrappedVerifier<B, C, CIDP, N> {
...
+        let mut header = block.header.clone();
+        let seal = header
+            .digest_mut()
+            .pop()
+            .ok_or_else(|| "Header Unsealed".to_string())?;
+        let sig = seal
+            .as_babe_seal()
+            .ok_or_else(|| "Header bad seal".to_string())?;
+
+        let authorities = self.get_last_aura_authorities(block.header)?;
```
shamil-gadelshin: What will happen during the warp sync on the validator nodes with block and state pruning?

node/src/service.rs
```diff
@@ -94,6 +100,13 @@ pub fn new_partial(
             executor,
         )?;
 
+    // Prepare keystore for authoring Babe blocks.
```
shamil-gadelshin, line 103: Can we do that once?

shamil-gadelshin: Or is it a temporary code block?

shamil-gadelshin (commented):
I see you force-pushed the code recently, so I believe https://github.com/opentensor/subtensor/pull/1941 changes are integrated into this branch. 

Do you mind providing an additional description of this PR?

node/src/consensus/aura_wrapped_import_queue.rs
liamaharon: When nodes warp sync they do not execute runtime checks (`fn verify`), they only check the header hashes.

The Polkadot SDK Aura and Babe consensus logic actually uses the strategy when verifying candidate block authors: https://github.com/opentensor/polkadot-sdk/blob/881e32b6c9e4c794502fc1fa0f854a4e07ade187/substrate/client/consensus/aura/src/import_queue.rs#L192-L200

node/src/service.rs
liamaharon: What is your concern about leaving it there? It is idempotent and takes a negligible time to run. 

Unless there is a downside, it would be nice to leave it in there imo. It's nice for the node to always have the ability to perform the Aura to Babe migration, for example in case we want to reset testnet or something.

liamaharon:
Thanks for taking the time to review @shamil-gadelshin. I have updated the PR description to include some more detail, let me know if you have any questions about specific changes. I'm also happy to jump on a Nucleus Sync to through anything if that would be helpful.

node/src/service.rs
shamil-gadelshin: It looks like a part of the migration.
I would annotate it with a description of when it is safe to remove, or not at all if it will never be safe.

node/src/consensus/aura_wrapped_import_queue.rs
shamil-gadelshin: Ok, so we'll never end up in a situation when we request authorities for some old block with the pruned state.

liamaharon:
Added TODO to verify warp sync works across the Aura -> Babe upgrade block. @shamil-gadelshin do you have suggestions about the best way to test this? I am having some trouble warp syncing with baedeker

shamil-gadelshin (dismissed):
Tested runtime upgrade and warp sync.

shamil-gadelshin:
> Added TODO to verify warp sync works across the Aura -> Babe upgrade block. @shamil-gadelshin do you have suggestions about the best way to test this? I am having some trouble warp syncing with baedeker

I would test the local configuration first and testnet later. How are you going to approach this?

node/src/consensus/aura_wrapped_import_queue.rs
liamaharon: Yes. I guess unless someone manually overrides `--state-pruning` to a value <2 (default is 256) it might fail? but that seems extremely unlikely to be required.

liamaharon:
> I would test the local configuration first and testnet later. How are you going to approach this?

I'm having trouble getting local warp sync to work on `devnet-ready`. When I run with `RUST_LOG=sync=debug` the warp sync node logs this and disconnects its peers due to bad signature:

```
2025-08-18 12:43:21.290 DEBUG tokio-runtime-worker sync: Bad warp proof response: bad justification for header: invalid signature for precommit in grandpa justification
2025-08-18 12:43:21.290 DEBUG tokio-runtime-worker sync: 12D3KooWP4ZUiDpYWsajCUeiqrxPev2NcRUrgZ5ZaEXAkW6QLmtF disconnected
```

I'm going to spend some time debugging it this week before we merge this, and hopefully get a fix into `devnet-ready` so we can test warp sync properly before we merge the hybrid-node. 


edit: fixed it by changing `set_id` in `sc_consensus_grandpa::warp_proof::HardForks::new_initial_set_id(set_id)` to `0`.

sam0x17 (dismissed):

liamaharon:
Found an issue with warp sync across the Aura / Babe upgrade boundary. Almost done fixing it, will re-request review when it's ready.

node/src/consensus/aura_consensus.rs
```diff
@@ -150,8 +150,13 @@ impl ConsensusMechanism for AuraConsensus {
                     Ok((slot, timestamp))
                 };
 
-                let import_queue = super::aura_wrapped_import_queue::import_queue(
-                    crate::consensus::aura_wrapped_import_queue::ImportQueueParams {
+                // Aura needs the hybrid import queue, because it needs to
+                // 1. Validate the first Babe block it encounters before switching into Babe
```
shamil-gadelshin, line 155: Why only "first Babe block"?  Shouldn't it import all blocks from the target warp sync block (state import block) up to the head before starting importing the block header history?

Cargo.toml
```diff
@@ -48,18 +48,18 @@ unwrap-used = "deny"
...
-pallet-commitments = { default-features = false, path = "pallets/commitments" }
-pallet-registry = { default-features = false, path = "pallets/registry" }
-pallet-crowdloan = { default-features = false, path = "pallets/crowdloan" }
-pallet-subtensor = { default-features = false, path = "pallets/subtensor" }
-pallet-subtensor-swap = { default-features = false, path = "pallets/swap" }
-pallet-subtensor-swap-runtime-api = { default-features = false, path = "pallets/swap/runtime-api" }
-pallet-subtensor-swap-rpc = { default-features = false, path = "pallets/swap/rpc" }
+node-subtensor-runtime = { path = "runtime", default-features = false }
+pallet-admin-utils = { path = "pallets/admin-utils", default-features = false }
+pallet-collective-otf = { path = "pallets/collective", default-features = false }
```
shamil-gadelshin: We usually use "subtensor" as a middle word for pallets. Also, why does it cause issues? Do you mind describing the conflicts?

node/src/consensus/hybrid_import_queue.rs
```diff
@@ -0,0 +1,287 @@
...
+
+#[async_trait::async_trait]
+impl BlockImport<Block> for HybridBlockImport {
+    type Error = ConsensusError;
+
+    async fn check_block(
+        &self,
+        block: BlockCheckParams<Block>,
+    ) -> Result<ImportResult, Self::Error> {
+        self.inner_aura.check_block(block).await.map_err(Into::into)
```
shamil-gadelshin: Why do we check only `aura` blocks?

node/src/consensus/aura_consensus.rs
```diff
@@ -179,30 +196,36 @@ impl ConsensusMechanism for AuraConsensus {
...
+                        let syncing = sync_service.status().await.is_ok_and(|status| status.warp_sync.is_some() || status.state_sync.is_some());
+                        if !c.authorities.is_empty() && !syncing {
+                            log::info!("Babe runtime detected! Intentionally failing the essential handle `babe-switch` to trigger switch to Babe service.");
+                            if let Some(triggered) = triggered {
+                                triggered.store(true, std::sync::atomic::Ordering::SeqCst);
+                            };
+                            break;
+                        }
+                    };
+                    tokio::time::sleep(slot_duration.as_duration()).await;
```
shamil-gadelshin, line 232: What happens if the actual block duration exceeds the slot duration? And in general, if the trigger activates in unsynchronized way after the block production?

node/src/consensus/babe_consensus.rs
```diff
@@ -200,6 +219,7 @@ impl ConsensusMechanism for BabeConsensus {
         _task_manager: &mut TaskManager,
         _client: Arc<FullClient>,
         _triggered: Option<Arc<std::sync::atomic::AtomicBool>>,
```
shamil-gadelshin: There are at least three names for the same system-wide trigger. Please, unify them.

shamil-gadelshin (changes requested):
Do you mind expanding this? 
> "Warp sync only works from a fresh sync. This means, if the service switches from Aura -> Babe mid-warp-sync, the warp sync will not continue correctly.""


Please, consider adding more details for changes - the high level description helps but it lacks the technical reasoning. Maybe even a list of the added algorithms with descriptions. Also, consider adding a description for major commit updates - for example, the same PR was changed significantly compared to the previously approved version that required full retesting and rereviewing from the beginning. 

New features (example):
- babe-switch signal that passes message from aura consensus to babe consensus. It shut downs aura service and starts babe service that in turn ....
- hybrid import queue, it checks only....
- babe configuration parameters (EPOCH_DURATION) ..... 

Did you consider rare situations for consensus switch (end of the epoch, validator failure, etc)? I wonder are there situations when aura produces at least one block in presence of the babe configuration?

node/src/consensus/aura_consensus.rs
liamaharon: Good catch, that's an old comment from before `AuraConsensus` could import Babe blocks. I will update it.

node/src/consensus/hybrid_import_queue.rs
liamaharon: Aura and Babe `check_block` implementations are identical. I will add a comment.

node/src/consensus/aura_consensus.rs
liamaharon: > What happens if the actual block duration exceeds the slot duration?

It will wait one more "block_duration" period of time before it checks again.

> And in general, if the trigger activates in unsynchronized way after the block production?

It is actually unlikely that all nodes will restart "perfectly" synchronised with one another. In practice with baedeker state, their restart time vary by ~1 second. This is not a problem though, even in the case of some catastrophic disruption and they restart minutes apart, as soon as 2/3 of them are online they will begin finalizing again.

Cargo.toml
liamaharon: The conflict is that these crate names are identical with crates in `polkadot-sdk`.

This means when we patch the Parity `polkadot-sdk` crates with the OTF `polkadot-sdk`, there is a risk that we will accidentally overwrite our crates with the `polkadot-sdk` versions.

By making the names of our crates unique, it is impossible for us to accidentally patch them. 

I can change the name to put `-subtensor-` in the middle instead of `-otf` prefix if you prefer.

liamaharon:
> Please, consider adding more details for changes - the high level description helps but it lacks the technical reasoning. Maybe even a list of the added algorithms with descriptions. Also, consider adding a description for major commit updates - for example, the same PR was changed significantly compared to the previously approved version that required full retesting and rereviewing from the beginning.

Apologies you found my description not clear enough. I will write up the changes in more detail. I am also happy to jump onto a call with you to walk through anything if you might find that helpful. 

> Do you mind expanding this?
>
>> "Warp sync only works from a fresh sync. This means, if the service switches from Aura -> Babe mid-warp-sync, the warp sync will not continue correctly.""

Warp sync does not work if you try to start on a partially synced DB, it logs an error and returns to full sync. Please let me know if there is something in particular about that you find unclear.   

> Did you consider rare situations for consensus switch (end of the epoch, validator failure, etc)? I wonder are there situations when aura produces at least one block in presence of the babe configuration?

There are no circumstances where an Aura block could be produced after the Babe runtime upgrade. The runtime after that point is configured only for Babe.

liamaharon:
Hi @shamil-gadelshin, I have addressed your comments and also re-written the "Update Aug 26" section of the PR to be more in-depth and include technical reasoning. 

Please let me know if that is more clear, and you have any further questions.

As always, appreciate your time and help reviewing this PR

Cargo.toml
shamil-gadelshin: @gztensor What do you suggest?

node/src/consensus/babe_consensus.rs
shamil-gadelshin: I'm sorry for not being specific. I meant the following names for the same variable: "triggered", "custom_service_signal", "babe_switch", and their "_clone" versions.

shamil-gadelshin (changes requested):
> Hi @shamil-gadelshin, I have addressed your comments and also re-written the "Update Aug 26" section of the PR to be more in-depth and include technical reasoning.
> 
> Please let me know if that is more clear, and you have any further questions.
> 
> As always, appreciate your time and help reviewing this PR

Thank you! I was finally able to understand the warp sync issues and some of your decisions.

I have a couple of minor issues left (crate names and "triggered" variable). Otherwise, looks good.

node/src/consensus/babe_consensus.rs
liamaharon: Ah I see now, good idea let me do that 👍🏼

node/src/consensus/aura_consensus.rs
```diff
@@ -178,31 +195,44 @@ impl ConsensusMechanism for AuraConsensus {
...
+                        // Aura and Babe blocks. Wait until sync finishes before switching to the
+                        // Babe service to not break warp sync.
+						//
+						// Note that although unintuitive, it is required that we wait until BOTH
+						// warp sync and state sync are finished before we can safely switch to the
+						// Babe service. If we only wait for the "warp sync" to finish while state
+						// sync is still in progress prior to switching, the warp sync will not
+						// complete successfully.
+                        let syncing = sync_service.status().await.is_ok_and(|status| status.warp_sync.is_some() || status.state_sync.is_some());
+                        if !c.authorities.is_empty() && !syncing {
```
gztensor, line 223: Are there any caveats for not switching to Babe immediately?

shamil-gadelshin:
> The conflict is that these crate names are identical with crates in polkadot-sdk.
> This means when we patch the Parity polkadot-sdk crates with the OTF polkadot-sdk, there is a risk that we will accidentally overwrite our crates with the polkadot-sdk versions.
> By making the names of our crates unique, it is impossible for us to accidentally patch them.
> I can change the name to put -subtensor- in the middle instead of -otf prefix if you prefer.

After the conversation with @gztensor we confirmed that we use "-subtensor-" for such cases. @liamaharon , please, rename the crates when you have time.

node/src/consensus/aura_consensus.rs
liamaharon: Not anything I can think while syncing.

liamaharon:
Thanks @gregzaitsev and @shamil-gadelshin, I've replaced the `-otf` with `-subtensor-` as suggested.

shamil-gadelshin (dismissed):
Thank you! 

Consider merging the latest `devnet-ready` to counter conflicts created by the benchmarking bot.

liamaharon:
Thanks @shamil-gadelshin , I have merged devnet-ready and also updated the Polkadot SDK fork to include your `sc-transaction-pool` change.

gztensor (dismissed):

sam0x17:
just some more conflicts, btw will prob want to hold off on deploying this until next week because we are doing 2 quick deploys this week that need to be low-risk

sam0x17 (approved):

PR #1955, opened 2025-08-15
Update transaction priority

shamil-gadelshin:
This PR changes Subtensor transaction priority algorithm. Here is the current algorithm description.

# Transaction Priority

### Overview
In Subtensor, transaction priority is determined by custom transaction extensions, which alter or override the default Substrate SDK behavior. Extensions affecting transaction priority are:

- **`ChargeTransactionPaymentWrapper`** (wraps `ChargeTransactionPayment`)
- **`SubtensorTransactionExtension`**

Substrate SDK combines priorities from all transaction extensions using addition. 

---

### 1. `ChargeTransactionPaymentWrapper`
In the Substrate SDK, `ChargeTransactionPayment` normally calculates transaction priority based on:
- **Tip** — an extra fee paid by the sender.
- **Weight** — computational complexity of the transaction.
- **Dispatch class** — category of the transaction (`Normal`, `Operational`, `Mandatory`).

However, in Subtensor, `ChargeTransactionPaymentWrapper` **overrides** this logic.  
It replaces the dynamic calculation with a **flat priority scale** based only on the dispatch class.

#### Current priority values:
| Dispatch Class      | Priority Value    | Notes |
|---------------------|-------------------|-------|
| `Normal`            | `1`               | Standard transactions |
| `Mandatory`         | `1`               | Rarely used, same as `Normal` |
| `Operational`       | `10_000_000_000`  | Reserved for critical system extrinsics (e.g., `sudo` calls, `drand` pulses) |

---

### 2. `SubtensorTransactionExtension`
This extension introduces **special priority rules** for certain extrinsics, such as:
- `commit_weights`
- `reveal_weights`
- etc.

For these, priority is **boosted** using a time-sensitive factor:
1. Retrieve the **current block number**.
2. Retrieve the **block number from the map `LastUpdate`**.
3. Calculate the **difference** between them.
4. **Add** this difference to the transaction’s priority.

---

# Additional changes
- staking transactions have no additional priorities now
- DRAND extrinsics will get `DispatchClass::Operational`
- utility batching extrinsics will have `DispatchClass::Normal` even if they contain extrinsics with other dispatch classes
- following extrinsics have their dispatch class changed from `Operational` to `Normal`:
  - pallet_registry::set_identity
  - pallet_commitments::set_commitment
  - pallet_subtensor::swap_hotkey
  - pallet_subtensor::register_network
  - pallet_subtensor::set_childen
  - pallet_subtensor::schedule_swap_coldkey
  - pallet_subtensor::schedule_dissove_network
  - pallet_subtensor::register_network_with_identity
  - pallet_subtensor::unstake_all
  - pallet_subtensor::unstake_all_alpha
  - pallet_subtensor::swap_stake
  - pallet_subtensor::swap_stake_limit
  - pallet_subtensor::move_stake
  - pallet_subtensor::transfer_stake
  - pallet_subtensor::try_associate_hotkey
  - pallet_subtensor::start_call
  - pallet_subtensor::associate_evm_key
  - pallet_subtensor::recycle_alpha
  - pallet_subtensor::burn_alpha

The absolute majority of the remaining extrinsics with `DispatchClass::Operational` have `ensure_root` or similar origin verification now.


pallets/subtensor/src/transaction_extension.rs
```diff
@@ -0,0 +1,402 @@
...
+                salts_list,
+                version_keys,
+            }) => {
+                if Self::check_weights_min_stake(who, *netuid) {
+                    let num_reveals = uids_list.len();
+                    if num_reveals == values_list.len()
+                        && num_reveals == salts_list.len()
+                        && num_reveals == version_keys.len()
+                    {
+                        let provided_hashs = (0..num_reveals)
```
camfairchild: Typo
```suggestion
                        let provided_hashes = (0..num_reveals)
```

pallets/subtensor/src/transaction_extension.rs
```diff
@@ -0,0 +1,402 @@
...
+                    Err(CustomTransactionError::StakeAmountTooLow.into())
+                }
+            }
+            Some(Call::batch_reveal_weights {
+                netuid,
+                uids_list,
+                values_list,
+                salts_list,
+                version_keys,
+            }) => {
```
camfairchild, line 167: This could possibly be two extensions with one checking for the coldkey swap to invalidate. The other one handles just this priority logic

runtime/src/transaction_payment_wrapper.rs
```diff
@@ -0,0 +1,161 @@
...
+                    match info.class {
+                        DispatchClass::Normal => 1u64,
+                        DispatchClass::Mandatory => {
+                            // Mandatory extrinsics should be prohibited (e.g. by the [`CheckWeight`]
+                            // extensions), but just to be safe let's return the same priority as `Normal` here.
+                            1u64
+                        }
+                        DispatchClass::Operational => {
+                            // System calls
+                            10_000_000_000u64
```
camfairchild, line 89: Put drand below sudo

camfairchild (approved):

runtime/src/transaction_payment_wrapper.rs
camfairchild: .

camfairchild (changes requested):
Please see comment

ppolewicz:
Drand pulse submission has to have a check in validation (when inserting into mempool) to see if it's signed by the designated offline worker - if we don't do that then someone can spam these and flood out weight setting, staking, transfers etc

ppolewicz:
After a discussion with Shamil, we found out that it should be made slightly simpler and the transaction priority for setting weights should be removed until further notice (due to newly found attack vectors)

EDIT: after an even longer discussion with Vune, it was agreed to remove the priority for setting weights

pallets/subtensor/src/transaction_extension.rs
ppolewicz: @shamil-gadelshin In fact, coldkey swap check should be done in validation, not in transaction extension (in a separate PR)

shamil-gadelshin:
Last updates:
- remove weights extrinsics periory
- new extension for DRAND pulse priority

pallets/subtensor/src/transaction_extension.rs
shamil-gadelshin: I removed the second (priority) extension after the discussion with @ppolewicz

runtime/src/transaction_payment_wrapper.rs
shamil-gadelshin: I added a special transaction extension for `write_pulse` priority. Other extrinsics from drand pallet have sudo origin check.

camfairchild (approved):

sam0x17 (approved):

PR #1970, opened 2025-08-21
Add proxy call to precompile

### LINKED ISSUE #1967: Add `proxy.proxy()` to the precompile
If `proxy.proxy()` was part of the precompile, smart contracts would be able to work with subnet tokens in a similar way to how the uniswap contract works with ERC20 tokens using the `approve` method.

The proxy precompile would need to be a method that could take another precompile method, then call is on behalf of the original user to interact with subnet tokens.

It should work something like this pseudocode:

```solidity
interface IProxyPrecompile {
    function proxy(
        address accountToProxy,
        string memory proxyType,
        bytes memory encodedCall  // Raw calldata for the nested precompile call
    ) external returns (bool success);
}

interface ITransferStakePrecompile {
    function transferStake(
        bytes32 destinationColdkey,
        bytes32 hotkey,
        uint256 originNetuid,
        uint256 destinationNetuid,
        uint256 amount
    ) external returns (bool success);
}

contract StakeManager {
    IProxyPrecompile constant PROXY = IProxyPrecompile(0x...);
    ITransferStakePrecompile constant TRANSFER_STAKE = ITransferStakePrecompile(0x...);
    
    function transferStakeOnBehalf(
         address userAccount,
         bytes32 to,
         bytes32 hotkey,
         uint256 netuid,
         uint256 amount
    ) external {
        bytes memory transferStakeCall = abi.encodeWithSelector(
            ITransferStakePrecompile.transferStake.selector,
            to,
            hotkey,
            netuid,
            netuid,
            amount
        );
        
        bool success = PROXY.proxy(
            userAccount,
            "Staking",
            transferStakeCall
        );
        require(success, "Proxy call failed");
    }
}
```

bdmason:
The goal of the proxy precompile is to allow smart contracts to interact with the subtensor pallet by making other precompile calls on behalf of users. Much like DEXs can swap tokens on behalf of users using ERC20's allowance functionality, the proxy precompile allows smart contracts to perform actions on behalf of users, such as transferring stake.

This is just a draft right now, it still needs tests to ensure it works, and also to ensure that the elevated permissions inside precompile can't be abused. We need tests to prove the proxied call is scoped to the proxy type and the proxied account.

Closes #1967 

This won't work, so adding a more specific `proxyTransferStake` precompile method instead.


bdmason:
we're going to pallet_contracts instead

PR #1973, opened 2025-08-23
add cargo machete

sam0x17:
adds `cargo machete` to CI, will automatically find unused deps. Note that we get a lot of false positives from anything that uses a lot of macros (pallets) so keep that in mind


sam0x17:
too broken in this workspace because of all the macros

PR #1987, opened 2025-08-27
do not update (singular) BlockEmission for every subnet

mvds00:
do not update (singular) BlockEmission for every subnet but only on fetching Tao block emission

## Description

BlockEmission is updated in a utility function, which is very clearly the wrong place

## Type of Change

- [ ] Documentation update

## Checklist

## Additional Notes

This is actively being discussed in discord channel refactor-run_coinbase

### DIFF
--- a/pallets/subtensor/src/coinbase/block_emission.rs
+++ b/pallets/subtensor/src/coinbase/block_emission.rs
@@ -98,7 +98,13 @@ impl<T: Config> Pallet<T> {
     ///
     pub fn get_block_emission() -> Result<TaoCurrency, &'static str> {
         // Convert the total issuance to a fixed-point number for calculation.
-        Self::get_block_emission_for_issuance(Self::get_total_issuance().into()).map(Into::into)
+        let block_emission =
+            Self::get_block_emission_for_issuance(Self::get_total_issuance().into());
+        let block_emission_u64: u64 = block_emission.unwrap_or(0);
+        if BlockEmission::<T>::get() != block_emission_u64 {
+            BlockEmission::<T>::put(block_emission_u64);
+        }
+        block_emission.map(Into::into)
     }
 
     /// Returns the block emission for an issuance value.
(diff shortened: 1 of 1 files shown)


pallets/subtensor/src/coinbase/block_emission.rs
```diff
@@ -98,7 +98,13 @@ impl<T: Config> Pallet<T> {
     ///
     pub fn get_block_emission() -> Result<TaoCurrency, &'static str> {
         // Convert the total issuance to a fixed-point number for calculation.
-        Self::get_block_emission_for_issuance(Self::get_total_issuance().into()).map(Into::into)
+        let block_emission =
+            Self::get_block_emission_for_issuance(Self::get_total_issuance().into());
+        let block_emission_u64: u64 = block_emission.unwrap_or(0);
+        if BlockEmission::<T>::get() != block_emission_u64 {
+            BlockEmission::<T>::put(block_emission_u64);
```
open-junius, line 105: it is not a good idea to update anything in a get method.

camfairchild: I agree. We should just remove this if we don't read it anywhere

mvds00: Shall I rebase the PR so it can be merged? And drop the update of BlockEmission altogether, even remove it from the pallet?

sam0x17: If @camfairchild thinks we should remove we should prob remove so would be great if you could

camfairchild: if you want to now, go ahead and remove it altogether

dougsillars: what's the udpate on this one?  It is getting called 129 times per block... and now AI slop sites are saying 1 tao/block :D

PR #1993, opened 2025-08-28
Feat/mvds00 run coinbase more refactoring

mvds00:
## Description

Some more refactoring, not changing functionality (except some debug logging that is now gone).

## Type of Change

- [ ] Documentation update
- [x] Other (please describe): refactoring, making code more readable/understandable

## Checklist

## Additional Notes

This PR is under debate in discord channel refactor-run_coinbase

### DIFF
--- a/pallets/subtensor/src/coinbase/run_coinbase.rs
+++ b/pallets/subtensor/src/coinbase/run_coinbase.rs
@@ -123,9 +123,6 @@ impl<T: Config> Pallet<T> {
             alpha_in.insert(*netuid_i, alpha_in_i);
             alpha_out.insert(*netuid_i, alpha_out_i);
         }
-        log::debug!("tao_in: {tao_in:?}");
-        log::debug!("alpha_in: {alpha_in:?}");
-        log::debug!("alpha_out: {alpha_out:?}");
 
         // --- 4. Injection.
         // Actually perform the injection of alpha_in, alpha_out and tao_in into the subnet pool.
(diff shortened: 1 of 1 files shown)


pallets/subtensor/src/coinbase/run_coinbase.rs
```diff
@@ -118,50 +117,44 @@ impl<T: Config> Pallet<T> {
...
             alpha_out.insert(*netuid_i, alpha_out_i);
-        }
-        log::debug!("tao_in: {tao_in:?}");
-        log::debug!("alpha_in: {alpha_in:?}");
-        log::debug!("alpha_out: {alpha_out:?}");
 
-        // --- 4. Injection.
-        // Actually perform the injection of alpha_in, alpha_out and tao_in into the subnet pool.
-        // This operation changes the pool liquidity each block.
-        for netuid_i in subnets_to_emit_to.iter() {
```
gztensor: The reason this was made a separate loop was that injection back when we had uniswap v2 could change price, and this would disturb the balance between subnet emissions. We needed to calculate all emissions first using that prices that do not change (because emissions are proportional to prices), and then inject.

My concern here is that we are going to have a tokenomics fix that may or may not return to this behavior (i.e. change price on injection). It is safe to merge now, but will require more work for that update in the next step.

mvds00: Isn't this a YAGNI case? It was five minutes to join them, it will be five to split them, and _only_ if needed. The certain benefit of having the simplest code (for all people involved) far outweighs the possible cost (for one developer) of a possible future change.

sam0x17: FYI we want to merge this stuff but not this week, we have a bunch of stuff touching these areas coming in within the next 5 days so please let's revisit next week 🙏🏻

PR #1996, opened 2025-08-29
Ckburn

unconst:
## Description
Adds a constant CK burn parameter which auto burns alpha dividends between parent <--> child without the same coldkey.


pallets/admin-utils/src/lib.rs
```diff
@@ -107,6 +107,8 @@ pub mod pallet {
         BondsMovingAverageMaxReached,
         /// Only root can set negative sigmoid steepness values
         NegativeSigmoidSteepness,
+        /// Value not in allowed bounds.
+        ValueNotInBounds,
```
gztensor, line 111: We have InvalidValue error

pallets/admin-utils/src/lib.rs
```diff
@@ -665,12 +667,22 @@ pub mod pallet {
...
                 Error::<T>::SubnetDoesNotExist
             );
+            // Min burn must be less than 1 TAO.
+            ensure!(
+                min_burn < TaoCurrency::from(1_000_000_000),
+                Error::<T>::ValueNotInBounds
+            );
+            // Min burn must be less than max burn
+            ensure!(
+                min_burn > pallet_subtensor::Pallet::<T>::get_max_burn(netuid),
```
gztensor: Wrong way ensure

gztensor: Also, what if someone wants mom burn == max burn?

gztensor (changes requested):

pallets/admin-utils/src/lib.rs
```diff
@@ -1682,6 +1703,20 @@ pub mod pallet {
             pallet_subtensor::Pallet::<T>::set_owner_immune_neuron_limit(netuid, immune_neurons)?;
             Ok(())
         }
+
+        /// The extrinsic sets the difficulty for a subnet.
+        /// It is only callable by the root account or subnet owner.
+        /// The extrinsic will call the Subtensor pallet to set the difficulty.
```
gztensor: Outdated comment

pallets/subtensor/src/coinbase/run_coinbase.rs
```diff
@@ -818,23 +819,25 @@ impl<T: Config> Pallet<T> {
...
+                // The parent is from a different coldkey, we burn some proportion
+                burn_take = burn_take_proportion.saturating_mul(parent_emission);
+                child_take = child_take_proportion.saturating_mul(parent_emission);
+                parent_emission = parent_emission.saturating_sub(burn_take);
+                parent_emission = parent_emission.saturating_sub(child_take);
+                total_child_take = total_child_take.saturating_add(child_take);
+                SubnetAlphaOut::<T>::mutate(netuid, |total| {
+                    *total = total
+                        .saturating_sub(AlphaCurrency::from(burn_take.saturating_to_num::<u64>()));
+                });
```
camfairchild: Should use the burn helper here.
Would allow us to change up the burn logic later

pallets/admin-utils/src/lib.rs
open-junius: will remove it.

pallets/subtensor/src/coinbase/run_coinbase.rs
open-junius: add a help function to do it.

pallets/admin-utils/src/lib.rs
open-junius: removed.

JohnReedV (approved):

l0r1s (approved):

gztensor (approved):

Swamination-Yuma:
I believe the subnet control portion is this also referenced in #2045?
Also, the decision to allow subnets to control what childkeying does has a great impact on validators. Validators already set their own delegate take rate and child key take rate; why allow subnets to determine what childkeyers generate on the backs of those performing the validation work? The relationship is validator-to-validator. Subnets are highly likely to set this burn to zero, making the feature largely worthless when under subnet control.

PR #1998, opened 2025-08-29
Trim uids

unconst:
## Description
Allows subnets to trim the number of active UIDs while preserving immune UIDs (temporally and owner owned), compressing the UIDs to the left and migrating/clearing associated storage.

A new extrinsic + storage item has been defined for the minimum allowed number of UIDs of a subnet and we can't trim below this limit which is set to 64 to start with and will allow gradual controlled trimming on a per subnet basis to prevent abuse.


pallets/admin-utils/src/lib.rs
```diff
@@ -107,6 +107,8 @@ pub mod pallet {
         BondsMovingAverageMaxReached,
         /// Only root can set negative sigmoid steepness values
         NegativeSigmoidSteepness,
+        /// Value not in allowed bounds.
```
gztensor, line 110: We have InvalidValue error

pallets/admin-utils/src/lib.rs
```diff
@@ -665,12 +667,22 @@ pub mod pallet {
...
                 pallet_subtensor::Pallet::<T>::if_subnet_exist(netuid),
                 Error::<T>::SubnetDoesNotExist
             );
+            // Min burn must be less than 1 TAO.
+            ensure!(
+                min_burn < TaoCurrency::from(1_000_000_000),
+                Error::<T>::ValueNotInBounds
+            );
+            // Min burn must be less than max burn
+            ensure!(
```
gztensor: Wrong way ensure

pallets/admin-utils/src/lib.rs
```diff
@@ -1682,6 +1703,31 @@ pub mod pallet {
             pallet_subtensor::Pallet::<T>::set_owner_immune_neuron_limit(netuid, immune_neurons)?;
             Ok(())
         }
+
+        /// Sets the number of immune owner neurons
```
gztensor: Outdated comment

pallets/subtensor/src/subnets/uids.rs
```diff
@@ -107,6 +96,143 @@ impl<T: Config> Pallet<T> {
         IsNetworkMember::<T>::insert(new_hotkey.clone(), netuid, true); // Fill network is member.
     }
 
+    /// Appends the uid to the network.
```
gztensor: Outdated comment

pallets/subtensor/src/subnets/uids.rs
```diff
@@ -107,6 +96,143 @@ impl<T: Config> Pallet<T> {
...
+    pub fn clear_neuron(netuid: NetUid, neuron_uid: u16) {
```
gztensor: We clear weights set FOR the uid being cleared. Do we also need to clear weights set BY the uid being cleared?

gztensor (changes requested):

pallets/subtensor/src/subnets/uids.rs
lorismoulin: Not sure to understand, I get the impression the comment is to the wrong section maybe?

In `clear_neuron` it seems we only reset the value but weights are not touched, only bonds are removed.

In `trim_to_max_allowed_uids` we effectively clear the weights, bonds and also all connections to remaining neurons here:
https://github.com/opentensor/subtensor/pull/1998/files#diff-6e4716cd71324b7ff12eb21a83873e667f5c91b5fe9a18f64bc2fe2b8f748ff0R212

Is it about this part?

gztensor: When a validator sets weights, the weights get into the state as a value-vector associated with netuid+validator_uid tuple:

```
(netuid, validator_uid) => [(miner_uid_1, weight_1), (miner_uid_2, weight_2), ...]
```

In `trim_to_max_allowed_uids` we make the value-vector shorter, effectively removing the unneeded miner UIDs.

But we ignore the fact that UID that is being removed could also have been the validator and have set weights, so the corresponding key `(netuid, validator_uid)` remains in the map, but it needs to be fully removed.

l0r1s: This part is already handled, deleting the entry in the Weights map:
https://github.com/opentensor/subtensor/pull/1998/files#diff-6e4716cd71324b7ff12eb21a83873e667f5c91b5fe9a18f64bc2fe2b8f748ff0R206

gztensor (dismissed):

ales-otf (dismissed):

ales-otf (dismissed):
Confused about these "UID's". At first glance it looks like a hyperparameter. If so, you need to extend the `Hyperparameter` enum we introduced earlier and use `TransactionType::OwnerUpdateHyperparam`. You would also need to use `ensure_root_with_limits` to enforce the freeze window. Otherwise just ignore. At the same time I can see it was before and it's configured via a constant.

l0r1s:
> Confused about these "UID's". At first glance it looks like a hyperparameter. If so, you need to extend the `Hyperparameter` enum we introduced earlier and use `TransactionType::OwnerUpdateHyperparam`. You would also need to use `ensure_root_with_limits` to enforce the freeze window. Otherwise just ignore. At the same time I can see it was before and it's configured via a constant.

Yes, it is like an hyperparameter but with a custom rate limit of 1 time per month to avoid abuse given it kicks out neurons.

I added the ensure_root_with_limit for the 2nd extrinsic!

sam0x17 (approved):

PR #2008, opened 2025-09-02
Auto stake

unconst:
## Description
Allow miners to set an auto stake destination so that mining income can be automatically staked to a validator of their choice.


pallets/subtensor/src/coinbase/run_coinbase.rs
```diff
@@ -508,11 +510,16 @@ impl<T: Config> Pallet<T> {
...
-
-            // Increase stake for miner.
+            let destination: T::AccountId;
+            let owner: T::AccountId = Owner::<T>::get(&hotkey);
+            if AutoStakeDestination::<T>::contains_key(&owner) {
+                destination = AutoStakeDestination::<T>::get(&owner);
+            } else {
+                destination = hotkey.clone();
+            }
             Self::increase_stake_for_hotkey_and_coldkey_on_subnet(
```
camfairchild, line 516: Should we emit an event here? This makes it way harder to account mining emissions vs staking emissions

shamil-gadelshin: We don't emit events in such cases in this file.

pallets/subtensor/src/lib.rs
```diff
@@ -1081,8 +1081,9 @@ pub mod pallet {
     pub type StakingHotkeys<T: Config> =
         StorageMap<_, Blake2_128Concat, T::AccountId, Vec<T::AccountId>, ValueQuery>;
     #[pallet::storage] // --- MAP ( cold ) --> Vec<hot> | Returns the vector of hotkeys controlled by this coldkey.
-    pub type OwnedHotkeys<T: Config> =
-        StorageMap<_, Blake2_128Concat, T::AccountId, Vec<T::AccountId>, ValueQuery>;
+    pub type OwnedHotkeys<T: Config> = StorageMap<_, Blake2_128Concat, T::AccountId, Vec<T::AccountId>, ValueQuery>;
+    #[pallet::storage] // --- MAP ( cold ) --> hot | Returns the hotkey a coldkey will autostake to with mining rewards.
+    pub type AutoStakeDestination<T: Config> = StorageMap<_, Blake2_128Concat, T::AccountId, T::AccountId, ValueQuery>;
```
JohnReedV: I would make this an `OptionQuery` instead of `ValueQuery`

sam0x17 (approved):

PR #2017, opened 2025-09-04
Rate limit hyperparams setting

ales-otf:
## Description

- introduce transaction rate limits for hyperparams updates
- introduce freeze window for hyperparams updates (10 last blocks of the epoch by default, during which noone, including root, can update hyperparams)
- add ensure function equivalents (ensure root, sn owner, owner-or-root) with the transaction rate limiting checks
- add extrinsics to configure the update rate limit and the freeze window

## Related Issue(s)

- Closes #[issue number]

## Type of Change

- [X] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] Documentation update

## Breaking Change

any attempt to set a hyperparameter via the freeze window will be blocked even for the root. the default value is the last 10 blocks of the epoch.

## Checklist

- [X] I have run `cargo fmt` and `cargo clippy` to ensure my code is formatted and linted correctly

## Additional Notes


pallets/admin-utils/src/lib.rs
```diff
@@ -216,10 +213,18 @@ pub mod pallet {
             netuid: NetUid,
             serving_rate_limit: u64,
         ) -> DispatchResult {
-            pallet_subtensor::Pallet::<T>::ensure_subnet_owner_or_root(origin, netuid)?;
-
+            let maybe_owner = pallet_subtensor::Pallet::<T>::ensure_sn_owner_or_root_with_limits(
```
gztensor, line 216: Since we are using the ensure-record nearly everywhere, it looks like a "with" pattern, which can be achieved in a similar way that we use to setup tests:

1. Closure helper:

```rust
impl<T: Config> Pallet<T> {
    fn with_owner_or_root_limits<F>(origin: OriginFor<T>,
                            netuid: NetUid,
                            txs: &'static [TransactionType],
                            f: F) -> DispatchResult
    where
        F: FnOnce(MaybeOwnerOf<T>) -> DispatchResult, // use your actual MaybeOwner type
    {
        // Run the extrinsic logic.
        let res = f(...);

        // Only record rate-limit on success.
        if res.is_ok() {
            Self::record_owner_rl(maybe_owner, netuid, txs);
        }
        res
    }
}
```

2. Usage in extrinsics:

```rust
pub fn sudo_set_kappa(origin: OriginFor<T>, netuid: NetUid, kappa: u16) -> DispatchResult {
    pallet_subtensor::Pallet::<T>::ensure_subnet_owner_or_root(origin.clone(), netuid)?;

    Self::with_owner_or_root_limits(
        origin,
        netuid,
        &[TransactionType::OwnerHyperparamUpdate],
        |...| {
            ensure!(
                pallet_subtensor::Pallet::<T>::if_subnet_exist(netuid),
            );
            pallet_subtensor::Pallet::<T>::set_kappa(netuid, kappa);
            Ok(())
        },
    )
}
```

pallets/admin-utils/src/lib.rs
```diff
@@ -1611,14 +1813,57 @@ pub mod pallet {
...
+            ensure_root(origin)?;
+            pallet_subtensor::Pallet::<T>::set_owner_hyperparam_rate_limit(limit);
+            log::debug!("OwnerHyperparamRateLimitSet( limit: {limit:?} ) ");
+            Ok(())
+        }
+    }
+
+    impl<T: Config> Pallet<T> {
+        // Helper: if owner path, record last-blocks for the provided TransactionTypes
+        fn record_owner_rl(
```
gztensor: If not implement "with" pattern, would it make sense to move this fn to misc in pallet_subtensor?

gztensor (changes requested):

pallets/admin-utils/src/lib.rs
ales-otf: makes sense! thanks!

pallets/admin-utils/src/lib.rs
ales-otf: Thanks! The idea is good and I understand that your intention is to prevent developers from forgetting to call it and simplify things. But it's actually too complex. You still need to write the same code, but in a less readable way. The only benefit here is `record_owner_rl` call. But when it's done that way, it's a silent mutation. It introduces a side-effect. The function ensures the owner and checks rate limits, as you can judge by its name, but under the hood it can record into the store. To make it clear we would need to either rename the method to something like `with_owner...record` or introduce a new method, which introduce more code and more confusion. But we would need a method without that mutation anyway, to be able to just check, without recording. So we anyway need to separate this stuff. Separating into plane methods is simple and clean. And you can't hide this knowledge. If we introduce one more kind of functions with a record, we would need to make the intention of those functions clear anyway. Developers would still need to know, that you must call that method, if you need to record and call another if you don't. From this point of view there is no difference - just more code and confusion.

ales-otf:
btcli tests failures are expected. the tests are updated in [this pr](https://github.com/opentensor/bittensor/pull/3052)

gztensor (dismissed):

sam0x17 (dismissed):

gztensor (approved):

l0r1s (approved):

basfroman:
> btcli tests failures are expected. the tests are updated in [this pr](https://github.com/opentensor/bittensor/pull/3052)

btcli fix PR https://github.com/opentensor/btcli/pull/620#pullrequestreview-3203000502

open-junius (approved):

sam0x17 (approved):

PR #1980, opened 2025-08-26
Feat/mvds00 run coinbase refactoring

mvds00:
## Description

Various minor tweaks to improve readability and performance of code. Fixes an issue where BlockEmission would be repeatedly set with wrong values.

## Type of Change

- [x] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] Documentation update
- [x] Other (please describe): cleanup

## Checklist

- [x] I have performed a self-review of my own code
- [-] I have commented my code, particularly in hard-to-understand areas
- [x] I have run `cargo fmt` and `cargo clippy` to ensure my code is formatted and linted correctly
- [-] I have made corresponding changes to the documentation
- [x] My changes generate no new warnings
- [-] I have added tests that prove my fix is effective or that my feature works
- [-] New and existing unit tests pass locally with my changes
- [-] Any dependent changes have been merged and published in downstream modules

## Additional Notes

This is the first of a possible sequence of small PRs to refactor some subtensor code. This was discussed at OpenDev and in the new `#refactor-run_coinbase` discord channel.

PR #1951, opened 2025-08-14
Ignore zero alpha stake in `clear_small_nomination_if_required`

antlerminator:
## Description
`clear_small_nomination_if_required` now validates that `alpha_stake > 0` before it proceeds to clear the small nomination. This fixes the issue where the function would emit unnecessary `StakeRemoved` and `Deposit` events with zero values.

## Related Issue(s)

- Closes #1950

## Type of Change

- [x] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] Documentation update
- [ ] Other (please describe):

## Breaking Change

If this PR introduces a breaking change, please provide a detailed description of the impact and the migration path for existing applications.

## Checklist

- [x] I have performed a self-review of my own code
- [x] I have commented my code, particularly in hard-to-understand areas
- [x] I have run `cargo fmt` and `cargo clippy` to ensure my code is formatted and linted correctly
- [x] I have made corresponding changes to the documentation
- [x] My changes generate no new warnings
- [ ] I have added tests that prove my fix is effective or that my feature works
- [x] New and existing unit tests pass locally with my changes
- [x] Any dependent changes have been merged and published in downstream modules

## Screenshots (if applicable)

Please include any relevant screenshots or GIFs that demonstrate the changes made.

## Additional Notes

Please provide any additional information or context that may be helpful for reviewers.

PR #1944, opened 2025-08-13
Feat/validate reg network

camfairchild:
## Description

Adds tx-validation to the `register_network` call to prevent spam.
- checks if the last registration- rate-limit has passed   

Also makes it pays::yes in-case the tx slips through some other way.

## Type of Change

- [ ] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [x] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] Documentation update
- [ ] Other (please describe):

## Breaking Change

@opentensor/cortex -- need to add fee estimate to `btcli` for this
@opentensor/gyrus 

## Checklist

- [x] I have performed a self-review of my own code
- [x] I have commented my code, particularly in hard-to-understand areas
- [x] I have run `cargo fmt` and `cargo clippy` to ensure my code is formatted and linted correctly
- [x] I have made corresponding changes to the documentation
- [x] My changes generate no new warnings
- [ ] I have added tests that prove my fix is effective or that my feature works
- [x] New and existing unit tests pass locally with my changes
- [x] Any dependent changes have been merged and published in downstream modules

## Screenshots (if applicable)

Please include any relevant screenshots or GIFs that demonstrate the changes made.

## Additional Notes

Please provide any additional information or context that may be helpful for reviewers.

PR #2020, opened 2025-09-04
mainnet deploy 9/9/2025

sam0x17:
## What's Changed
- #2033 by @JohnReedV
- #1944 by @camfairchild
- #2027 by @open-junius
- #2029 by @open-junius
- #2016 by @JohnReedV
- #1981 by @open-junius
- #2028 by @open-junius
- #2018 by @open-junius
- #1991 by @open-junius
- #2019 by @shamil-gadelshin
- #1976 by @shamil-gadelshin
- #1721 by @shamil-gadelshin
- #1996 by @unconst
- #2014 by @shamil-gadelshin
- #1995 by @unconst
- #2013 by @open-junius
- #2008 by @unconst
- #2012 by @gztensor
- #1986 by @JohnReedV
- #1965 by @shamil-gadelshin
- #1999 by @mateusfigmelo
- #1997 by @mateusfigmelo
- #2001 by @mateusfigmelo
- #2000 by @mateusfigmelo
- #1990 by @JohnReedV
- #1994 by @mateusfigmelo
- #1992 by @prop-opentensor
- #1980 by @mvds00
- #1959 by @shamil-gadelshin
- #1951 by @antlerminator


gztensor (approved):

bdmason:
@open-junius @unconst are you intending to burn or recycle 18% of the parent dividends?

Normally when burning `SubnetAlphaOut` isn't updated, by leaving alpha in `SubnetAlphaOut` it ensures the burned alpha is inaccessible put still part of supply so can't be emitted later on.

If you take a look at [recycle_alpha.rs](https://github.com/opentensor/subtensor/blob/main/pallets/subtensor/src/staking/recycle_alpha.rs#L131) you'll see how burning leaves `SubnetAlphaOut` untouched, whilst recycling reduces it. See the comment from @camfairchild 

In your PR to reduce parent dividends by 18% you added a new method to pallets/subtensor/src/staking/helpers.rs

```rust
pub fn burn_subnet_alpha(netuid: NetUid, amount: AlphaCurrency) {
    SubnetAlphaOut::<T>::mutate(netuid, |total| {
        *total = total.saturating_sub(amount);
    });
}
```

This contradicts what the existing `do_burn_alpha` method is doing.

camfairchild (dismissed):
https://github.com/opentensor/subtensor/pull/2020#issuecomment-3265021824

open-junius:
> @open-junius @unconst are you intending to burn or recycle 18% of the parent dividends?
> 
> Normally when burning `SubnetAlphaOut` isn't updated, by leaving alpha in `SubnetAlphaOut` it ensures the burned alpha is inaccessible put still part of supply so can't be emitted later on.
> 
> If you take a look at [recycle_alpha.rs](https://github.com/opentensor/subtensor/blob/main/pallets/subtensor/src/staking/recycle_alpha.rs#L131) you'll see how burning leaves `SubnetAlphaOut` untouched, whilst recycling reduces it. See the comment from @camfairchild
> 
> In your PR to reduce parent dividends by 18% you added a new method to pallets/subtensor/src/staking/helpers.rs
> 
> ```rust
> pub fn burn_subnet_alpha(netuid: NetUid, amount: AlphaCurrency) {
>     SubnetAlphaOut::<T>::mutate(netuid, |total| {
>         *total = total.saturating_sub(amount);
>     });
> }
> ```
> 
> This contradicts what the existing `do_burn_alpha` method is doing.

@unconst Please confirm that if we should reduce the SubnetAlphaOut for child fee burn case.

bdmason:
The 18% reduction in income to validators using parent keys is a huge change, particularly for Taostats. Perhaps a simpler approach could solve the problems:

Why not set a minimum take of 5% and a minimum child take of 5%? It makes it impossible for a non-infra running validator to outperform the best infra running validator.

Also, we could send the child take to the validator coldkey as profit rather than distribute it as staking dividends like we do now _(this is now it used to work if I recall correctly),_ so they are incentivised whilst not being able to boost APY by being the first one to get popular with parent keys.

bdmason:
> > @open-junius @unconst are you intending to burn or recycle 18% of the parent dividends?
> > Normally when burning `SubnetAlphaOut` isn't updated, by leaving alpha in `SubnetAlphaOut` it ensures the burned alpha is inaccessible put still part of supply so can't be emitted later on.
> > If you take a look at [recycle_alpha.rs](https://github.com/opentensor/subtensor/blob/main/pallets/subtensor/src/staking/recycle_alpha.rs#L131) you'll see how burning leaves `SubnetAlphaOut` untouched, whilst recycling reduces it. See the comment from @camfairchild
> > In your PR to reduce parent dividends by 18% you added a new method to pallets/subtensor/src/staking/helpers.rs
> > ```rust
> > pub fn burn_subnet_alpha(netuid: NetUid, amount: AlphaCurrency) {
> >     SubnetAlphaOut::<T>::mutate(netuid, |total| {
> >         *total = total.saturating_sub(amount);
> >     });
> > }
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > This contradicts what the existing `do_burn_alpha` method is doing.
> 
> @unconst Please confirm that if we should reduce the SubnetAlphaOut for child fee burn case.

In that method `burn_subnet_alpha` you're actually recycling it (which I assume is probably the real intention here). I would be careful of having this method in a generic helper file too as it might be misused in the future, this method is only useful during the distribution of emission as it's simply offsetting the `SubnetAlphaOut` increase from earlier in the distribution logic.

open-junius:
> The 18% reduction in income to validators using parent keys is a huge change, particularly for Taostats. Perhaps a simpler approach could solve the problems:
> 
> Why not set a minimum take of 5% and a minimum child take of 5%? It makes it impossible for a non-infra running validator to outperform the best infra running validator.
> 
> Also, we could send the child take to the validator coldkey as profit rather than distribute it as staking dividends like we do now _(this is now it used to work if I recall correctly),_ so they are incentivised whilst not being able to boost APY by being the first one to get popular with parent keys.

@ppolewicz any comment from you?

camfairchild:
> Why not set a minimum take of 5% and a minimum child take of 5%? It makes it impossible for a non-infra running validator to outperform the best infra running validator.

Total would be ~~\~9.025%~~ ~9.75% which is basically the same as the 9% of a real Validator.
Possibly could be higher with a similar premise.

> Also, we could send the child take to the validator coldkey as profit rather than distribute it as staking dividends like we do now _(this is now it used to work if I recall correctly),_ so they are incentivised whilst not being able to boost APY by being the first one to get popular with parent keys.

It's a burn so that rakebacks aren't possible

bdmason:
> > Why not set a minimum take of 5% and a minimum child take of 5%? It makes it impossible for a non-infra running validator to outperform the best infra running validator.
> 
> Total would be ~9.025% which is basically the same as the 9% of a real Validator. Possibly could be higher with a similar premise.
> 
> > Also, we could send the child take to the validator coldkey as profit rather than distribute it as staking dividends like we do now _(this is now it used to work if I recall correctly),_ so they are incentivised whilst not being able to boost APY by being the first one to get popular with parent keys.
> 
> It's a burn so that rake backs aren't possible

A trivialised example to help illustrate it:

- Vali A runs infra
- Vali B sets A as child
- Their total stakes are equal
- Both have 5% take and Vali A has 5% CK take
- Total dividends for the epoch are 200 Alpha
- Vali A gets 105 Alpha, 5 Alpha is CK take, 5 is normal take
- Vali B gets 95 Alpha, 4.75 is normal take
- Vali A APY is based off 95
- Vali B APY is based off 90.25

5% better relative APY for the infra running validator
52.75% better relative profit for the infra running validator

Now Vali A decides to use his profits to buy more parents, so same as above, but with an extra parent Vali C, same total stake as A & B for easy math.

- Total dividends for the epoch are 300 Alpha
- Vali A gets 110 Alpha, 10 Alpha is CK take, 5 is normal take
- Vali B gets 95 Alpha, 4.75 is normal take
- Vali C gets 95 Alpha, 4.75 is normal take
- Vali A APY is based off 95
- Vali B APY is based off 90.25
- Vali C APY is based off 90.25

Still 5% better relative APY for the infra running validator, the APY doesn't scale with rake backs. You can still buy dominance in a subnet with rake backs though, but this is a subnet problem, they need to incentivise diverse validation somehow, a bit like TaoHash did for a short while with BTC.

Having a small minimum take and child take keeps the infra running validators at the top of the yield table which is fair, and the more parents they attract the more profit they make, which is also fair... infra isn't free.

Trying to protect infra running validators is obviously very important for the network, but we can't ignore the fact that the network's leading explorer is run from proceeds from validation, and slashing it by 18% isn't a trivial change. The wider adoption of the network is still in its infancy so fees for services will take time to outweigh validator emission.

camfairchild:
> Having a small minimum take and child take keeps the infra running validators at the top of the yield table which is fair, and the more parents they attract the more profit they make, which is also fair... infra isn't free.

The idea is to have mainly them earning the take. Because they're contributing the Validation that they're being incentivized to contribute.

> Trying to protect infra running validators is obviously very important for the network, but we can't ignore the fact that the network's leading explorer is run from proceeds from validation, and slashing it by 18% isn't a trivial change. The wider adoption of the network is still in its infancy so fees for services will take time to outweigh validator emission.

It doesn't make sense to cut out funding for anything that isn't an actual Validator.

Otherwise we end up funding tao.bot as the top Validator.

And then we have Validators making more emission than OTF while OTF runs infra *and* maintains the chain.

bdmason:
> The idea is to have mainly them earning the take. Because they're contributing the Validation that they're being incentivized to contribute.

In the example above the infra running validators do earn much more take, it goes straight to them as profit.

> It doesn't make sense to cut out funding for anything that isn't an actual Validator.
> 
> Otherwise we end up funding tao.bot as the top Validator.
> 
> And then we have Validators making more emission than OTF while OTF runs infra _and_ maintains the chain.

It would be good for OTF to force a take, currently the top 3 validators are tao.bot (0% take), taostats (9%), OTF (18%).

The changes I'm proposing would make your APY more competitive and your income for funding development significantly higher (by putting both take _and_ child take on your coldkey).

With the extra income from child take you could afford to drop your normal take and compete with APY again.

We can't avoid the fact that significant parts of the ecosystem are funded by validator take right now, so whilst long term it might make sense to eventually force all take to infra running validators, in the short term less harmful to the ecosystem to simply guarantee high performing infra running validators can't be beaten on APY and take home profits, but we can't do this with a minimum take of 0%.

CapriciousSage:
> Having a small minimum take and child take keeps the infra running validators at the top of the yield table which is fair, and the more parents they attract the more profit they make, which is also fair... infra isn't free.

The original reason for the 18% was based on the APY difference of a real-infra validator running at a 9% take and a (zero-overhead) "CHK-optimised" vali running at 0% take, and the hypothetical scenario of them rapidly becoming the largest validator on the ecosystem, then using that success to shill their ERC-20 sh*tcoin (again, totally hypothetical scenario *cough*).

Second, the reason that mandating a take instead of a recycle doesn't work is that instead of encouraging more validators to run infra, it would inevitably lead to LESS validators running real infra through a process of "he who who gets biggest first wins forever".

- The larger the vali, the larger the vtrust advantage it often gets, making it more attractive to CHK to.
- The more other valis CHK, the more the target's APY will skyrocket through all of the forced CHK takes
- The more that one vali's APY skyrockets through all the forced CHK takes, the more it absorbs all the stake from the other legit valis until eventually there is only one vali left running infra (and no one has any choice but to CHK to them because they control kappa1, and launching a competing validator will have terrible vtrust with performance lower than the fees for CHKing).

By having a tax (recycle), it ensures that two perfectly-operating legitimate validators with the same take have the same base APY. If they choose to fuddle about with their CHK Take, that's their call (and market dynamics will do their thing), but mandating the CHK take only serves to collapse the subnet.

Now, as the validator that actually has a super-majority on *quite* a lot of the smaller subnets, a fixed CHK take would actually be fantastic for our profitability, but having us (RT21) as the only validator on 2/3 of Bittensor is a really really bad idea for the decentralisation of the network. We're meant to be moving towards "Proof of Stake", not "Proof of Sage" :-P

> Trying to protect infra running validators is obviously very important for the network, but we can't ignore the fact that the network's leading explorer is run from proceeds from validation, and slashing it by 18% isn't a trivial change. The wider adoption of the network is still in its infancy so fees for services will take time to outweigh validator emission.

The issue here isn't the CHK Take or the CHK Tax... it's the fact that the ecosystems only real block explorer is funded by a validator that doesn't validate, and we're arguing about keeping a broken system broken to ensure funding to key infrastructure.

Taostats provides essential ecosystem infrastructure that NEEDS to be funded. In any other project of this age, they would be funded through ecosystem grants or other ecosystem funding mechanisms that would be put to governance. Until we can replace mog's dev team with AI super-agents running on a subnet, wages (and servers) need to be paid, and the ecosystem needs a better way of picking up the tab here.

Meanwhile, if taostats wants to generate (additional) funding with a validator, then they need to run a proper validator.

> It would be good for OTF to force a take, currently the top 3 validators are tao.bot (0% take), taostats (9%), OTF (18%). [...] we can't do this with a minimum take of 0%.

As an aside, if we moved back to a minimum take of 9%, it would reduce the afformentioned gap between the "real infra" vali and the currently-0% "CHK-only" vali, meaning that a 9% tax on CHK (instead of 18%) would largely eliminate the remaining delta. We'd need to make sure that certain WC valis are eliminated though, as the few remaining instances that WC is still out-performing the legit valis creates a CHK-optimisation path that would still probably put them slightly ahead of legit-infra valis.

bdmason:
> The larger the vali, the larger the vtrust advantage it often gets, making it more attractive to CHK to.

I'm not an expert on vtrust, but it sounds like the real issue here is that vtrust is too tightly coupled with stake weight, the weights of smaller validators need more weight.

> Meanwhile, if taostats wants to generate (additional) funding with a validator, then they need to run a proper validator.

We did, then the network incentivised us not to by introducing child keys and expanding the number of subnets to 128. You more than anyone know how many of those subnets are worth the effort.

> As an aside, if we moved back to a minimum take of 9%, it would reduce the afformentioned gap between the "real infra" vali and the currently-0% "CHK-only" vali, meaning that a 9% tax on CHK (instead of 18%) would largely eliminate the remaining delta. We'd need to make sure that certain WC valis are eliminated though, as the few remaining instances that WC is still out-performing the legit valis creates a CHK-optimisation path that would still probably put them slightly ahead of legit-infra valis.

That's not the worst middle ground. 18% tax is too high, and whilst there's no proven solution to WC and no chance of removing child keying in the immediate future, 0% take as you pointed out allows those who do very little for the network to win the APY wars.

If the vtrust "biggest first wins" issue can be resolved, it's still a lot simpler to introduce sensible minimum take and child take for now. Once WC is dealt with and the quality of subnets improves, removing child keying completely (or at least making it only for your own hotkeys) will force all validators to validate, but right now it's not practical.

bdmason:
It looks like the tax is now going to be 0%. See: https://github.com/opentensor/subtensor/commit/44bc6a9e9cba7081d042d5ff33f7a00dd66d9c38

And the burn is going to be a recycle. See: https://github.com/opentensor/subtensor/pull/2029/files

It would be good to get some confirmation of this a while before it reaches mainnet. The tax has a significant effect on the distribution of funds on the chain, and the burn effects the explorer.

open-junius:
> It looks like the tax is now going to be 0%. See: [44bc6a9](https://github.com/opentensor/subtensor/commit/44bc6a9e9cba7081d042d5ff33f7a00dd66d9c38)
> 
> And the burn is going to be a recycle. See: https://github.com/opentensor/subtensor/pull/2029/files
> 
> It would be good to get some confirmation of this a while before it reaches mainnet. The tax has a significant effect on the distribution of funds on the chain, and the burn effects the explorer.

Yes. I will ask team to review the #2029. then everything will be in mainnet as expected.

JohnReedV (approved):

PR #2033, opened 2025-09-09
Events for Incentive Emitted To Miners

JohnReedV:
## Description
Emit and event for 
- Incentive alpha emitted to miners in the epoch
- Auto Stakes to hotkey

### DIFF
--- a/pallets/subtensor/src/coinbase/run_coinbase.rs
+++ b/pallets/subtensor/src/coinbase/run_coinbase.rs
@@ -514,13 +514,13 @@ impl<T: Config> Pallet<T> {
             let owner: T::AccountId = Owner::<T>::get(&hotkey);
             let destination = AutoStakeDestination::<T>::get(&owner).unwrap_or(hotkey.clone());
 
-            Self::deposit_event(Event::<T>::AutoStakeAdded(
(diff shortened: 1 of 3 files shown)


pallets/subtensor/src/coinbase/run_coinbase.rs
```diff
@@ -513,6 +513,13 @@ impl<T: Config> Pallet<T> {
 
             let owner: T::AccountId = Owner::<T>::get(&hotkey);
             let destination = AutoStakeDestination::<T>::get(&owner).unwrap_or(hotkey.clone());
+
+            Self::deposit_event(Event::<T>::AutoStakeAdded(
+                netuid,
```
camfairchild, line 518: Add miner hotkey here

camfairchild (changes requested):

pallets/subtensor/src/coinbase/run_coinbase.rs
JohnReedV: `destination` is the hotkey if there is no autostake key. Should we add both regardless?

sam0x17 (approved):

PR #2041, opened 2025-09-10
set auto hotkey per subnet

### LINKED ISSUE #2038: Add getter for auto-stake hotkey and make auto-staking per-netuid

The auto-staking implementation as it is has two limitations (IMO):
- Miners cannot check what their auto-stake destination is set to. There is a `set_coldkey_auto_stake_hotkey` but no corresponding getter, which would also be ideal for the BTCLI/SDK.
- The auto-stake is currently implemented globally across all subnets, it means miners can only set one auto-stake destination for their entire coldkey. This forces miners to choose a single validator across all subnets they mine on (assuming they use the same coldkey), but a miner may want to auto-stake on netuid 1 but not on netuid 50. It also causes their auto-stake to get no dividends should the set destination not CHK or not be a validator on the subnet, while it may be on a different subnet (where the miner also mines).

- Implement `get_coldkey_auto_stake_hotkey(coldkey, netuid)` that allows miners to query their current auto-stake destination for a specific subnet.
- Modify the auto-stake destination to work per-netuid instead of globally

_No response_

### Additional context

I also think that the auto-stake destination should be the miner/UID's hotkey by default, then they can change it on the netuid, and to turn it "off" they would set it back to their miner's hotkey

open-junius:
## Description
Add auto hotkey for coldkey per subnet.

## Related Issue(s)

- Closes #2038 

## Type of Change

- [ ] Documentation update

## Breaking Change

## Checklist

## Additional Notes


pallets/subtensor/src/lib.rs
```diff
@@ -1140,9 +1140,16 @@ pub mod pallet {
...
+    #[pallet::storage] // --- DMAP ( cold, netuid )--> hot | Returns the hotkey a coldkey will autostake to with mining rewards.
+    pub type AutoStakeDestination<T: Config> = StorageDoubleMap<
+        _,
+        Blake2_128Concat,
+        T::AccountId,
+        Identity,
+        NetUid,
+        T::AccountId,
+        OptionQuery,
+    >;
```
lorismoulin, line 1152: Is this code already in use? Because it could requires a migration

l0r1s (dismissed):
No sure if there is a need for a migration, if this is not already used, you can ignore

pallets/subtensor/src/lib.rs
open-junius: Thanks for the reminder. I was just wandering if we should add the old default key to all subnet, it is better to add migration process.

basfroman:
Potential problem is that autostaking can now be done on a hot key that's not only not registered in the subnet, but doesn't even exist in the chain.
Also, in this impl, the subnet may not exist either, but autostaking will be completed successfully.
Where will the staking be done if the key doesn't exist?

open-junius:
> Potential problem is that autostaking can now be done on a hot key that's not only not registered in the subnet, but doesn't even exist in the chain. Also, in this impl, the subnet may not exist either, but autostaking will be completed successfully. Where will the staking be done if the key doesn't exist?

will add the check in extrinsic.

sam0x17 (approved):

PR #2048, opened 2025-09-11
hyperparameter refactor

sam0x17:
* set rate limit to be based on tempo
* rate limit default => 2 tempos
* make rate limit _apply_ individually for each hyperparameter instead of one global rate limit
* all hyperparameters still use the same rate limit


pallets/subtensor/src/migrations/migrate_owner_hparam_rl_to_epochs.rs
lorismoulin, line 1: Not sure if this migration is needed if the StorageValue doesn't exists yet? I checked devnet/testnet and I couldn't find it

l0r1s (dismissed):

pallets/subtensor/src/migrations/migrate_owner_hparam_rl_to_epochs.rs
ales-otf: Thanks, removed

open-junius (approved):

l0r1s (approved):

PR #2053, opened 2025-09-12
Root claim

shamil-gadelshin:
# Root Claim

This PR introduces a **root claim** mechanism. It replaces the automatic selling of root-alpha dividends with accumulation. Users can either:

* **Manually claim** their accumulated share, or
* **Enable auto-sell**, which automatically sells the accumulated share daily and transfers the proceeds to their account.

## New extrinsics

* `set_root_claim_type` — Set root-claim mode: `auto` or `manual`.
* `claim_root` — Manually claim accumulated root dividends.

## Storage changes

* `PendingRootAlphaDivs` — Tracks root-alpha dividends between epochs.
* `RootClaimType` — User’s preferred claim mode (`auto` or `manual`).
* `RootClaimable` — User’s claimable share of alpha dividends.
* `RootClaimed` — Accounting helper for alpha dividends (stake changes, claimed amounts).
* `StakingColdkeysByIndex` — Supports randomized distribution of auto-claims per block.
* `StakingColdkeys` — Supports root-dividends distribution.
* `NumStakingColdkeys` — Count to support randomized auto-claim distribution per block.
* `NumRootClaim` — Number of auto-claims to fulfill for a subnet in the current epoch.
* `AlphaMapLastKey` — Supports staking coldkey migration.

### Removed (obsolete storage maps)

* `TaoDividendsPerSubnet`
* `PendingAlphaSwapped`

## Original PRs

* [https://github.com/opentensor/subtensor/pull/1223](https://github.com/opentensor/subtensor/pull/1223)
* [https://github.com/opentensor/subtensor/pull/1232](https://github.com/opentensor/subtensor/pull/1232)

---


pallets/subtensor/src/staking/helpers.rs
```diff
@@ -329,6 +329,56 @@ impl<T: Config> Pallet<T> {
...
+
+                new_starting_key = Some(Alpha::<T>::hashed_key_for(key));
+            }
+
+            // Restart the process if it's the last batch
+            if new_iteration {
+                new_starting_key = None;
+            }
+
+            AlphaMapLastKey::<T>::put(new_starting_key);
```
camfairchild, line 375:
```suggestion
            let mut new_starting_raw_key = None;
            let new_iteration = keys.len() < ALPHA_MAP_BATCH_SIZE;

            // Check and remove alphas if necessary
            for key in keys {
                let (_, coldkey, netuid) = key.clone();

                if netuid == NetUid::ROOT {
                    Self::maybe_add_coldkey_index(&coldkey);
                }

                new_starting_raw_key = Some(Alpha::<T>::hashed_key_for(key));
            }

            // Restart the process if it's the last batch
            if new_iteration {
                new_starting_raw_key = None;
            }

            AlphaMapLastKey::<T>::put(new_starting_raw_key);
``` 
This naming is more consistent with what's happening 
`new_start_key` -> `new_starting_raw_key`

pallets/subtensor/src/lib.rs
```diff
@@ -1051,17 +1116,6 @@ pub mod pallet {
         ValueQuery,
         DefaultZeroAlpha<T>,
     >;
-    #[pallet::storage] // --- DMAP ( netuid, hotkey ) --> u64 | Last total root dividend paid to this hotkey on this subnet.
```
camfairchild, line 1040: do we need to clear this in a migration?

pallets/subtensor/src/staking/claim_root.rs
```diff
@@ -0,0 +1,335 @@
...
+                weight.saturating_accrue(Self::do_root_claim(coldkey.clone()));
+            }
+
+            continue;
+        }
+
+        weight
+    }
+
+    pub fn change_root_claim_type(coldkey: &T::AccountId, new_type: RootClaimTypeEnum) {
```
camfairchild, line 346: should we remove the entry in the coldkey index if they're not wanting Auto-claim?

pallets/subtensor/src/coinbase/run_coinbase.rs
```diff
@@ -567,37 +544,31 @@ impl<T: Config> Pallet<T> {
...
             // Give the validator their take.
-            log::debug!("hotkey: {hotkey:?} tao_take: {tao_take:?}");
-            let validator_stake = Self::increase_stake_for_hotkey_and_coldkey_on_subnet(
+            log::debug!("hotkey: {hotkey:?} alpha_take: {alpha_take:?}");
+            let _validator_stake = Self::increase_stake_for_hotkey_and_coldkey_on_subnet(
                 &hotkey,
                 &Owner::<T>::get(hotkey.clone()),
-                NetUid::ROOT,
-                tou64!(tao_take).into(),
+                netuid,
```
camfairchild, line 542: This is potentially a big deal.
We're giving root Validators alpha only instead of TAO.

pallets/subtensor/src/staking/claim_root.rs
```diff
@@ -0,0 +1,335 @@
...
+        // Get total stake on this hotkey on root.
+        let total: I96F32 =
+            I96F32::saturating_from_num(Self::get_stake_for_hotkey_on_subnet(hotkey, NetUid::ROOT));
+
+        // Get increment
+        let increment: I96F32 = I96F32::saturating_from_num(amount)
+            .checked_div(total)
+            .unwrap_or(I96F32::saturating_from_num(0.0));
+
+        if u64::from(amount) > total.saturating_to_num::<u64>() {
```
camfairchild, line 55: While I think this should never happen we should probably add it to their alpha stake or something.

pallets/subtensor/src/staking/claim_root.rs
```diff
@@ -0,0 +1,335 @@
...
+            log::error!("Not enough root stake. NetUID = {netuid}");
+
+            return;
+        }
+
+        // Increment claimable for this subnet.
+        RootClaimable::<T>::mutate(hotkey, |claimable| {
+            claimable
+                .entry(netuid)
+                .and_modify(|total| *total = total.saturating_add(increment))
```
camfairchild:
```suggestion
                .and_modify(|claim_total| *claim_total = claim_total.saturating_add(increment))
```

pallets/subtensor/src/staking/helpers.rs
```diff
@@ -326,6 +326,56 @@ impl<T: Config> Pallet<T> {
...
+        let mut new_starting_raw_key = AlphaMapLastKey::<T>::get();
+        let mut starting_key = None;
+        if new_starting_raw_key.is_none() {
+            starting_key = Alpha::<T>::iter_keys().next();
+            new_starting_raw_key = starting_key.as_ref().map(Alpha::<T>::hashed_key_for);
+        }
+
+        if let Some(starting_raw_key) = new_starting_raw_key {
+            // Get the key batch
+            let mut keys = Alpha::<T>::iter_keys_from(starting_raw_key)
```
camfairchild, line 343: Does this handle if the starting raw key is no longer in the map?

pallets/subtensor/src/coinbase/run_coinbase.rs
shamil-gadelshin: Isn't it almost the same because of the conversion above?

pallets/subtensor/src/staking/helpers.rs
shamil-gadelshin: To the best of my knowledge - it will just iterate to the next key (byte slice).

pallets/subtensor/src/staking/claim_root.rs
shamil-gadelshin: It will create holes in indexing that will affect random claims distribution.

pallets/subtensor/src/staking/claim_root.rs
shamil-gadelshin: Correct me if I'm wrong, but it will be the same, as we'll give everything to the validator owner. 

I introduced this check during the testing. Incorrect test setup produced unexpected results. It's unlikely that we have this on mainnet. And if we do have this condition  - we chose a safe route. 

I'm open to suggestions here.

pallets/subtensor/src/staking/helpers.rs
shamil-gadelshin: I looked into the implementation of `iter_keys_from`. It seems that it will `iter_after` indeed. 
```
	fn next_storage_key(&mut self, key: &[u8]) -> Option<StorageKey> {
		self.overlay.iter_after(key).find_map(|(k, v)| v.value().map(|_| k.to_vec()))
	}

	/// Returns an iterator over the keys (in lexicographic order) following `key` (excluding `key`)
	/// alongside its value.
	pub fn iter_after(&mut self, key: &[u8]) -> impl Iterator<Item = (&[u8], &mut OverlayedValue)> {
		self.top.changes_after(key)
	}
```

pallets/subtensor/src/staking/claim_root.rs
camfairchild: is there a way around that? There's an issue where I could stake-unstake and create a ton of index entries

pallets/subtensor/src/staking/claim_root.rs
camfairchild: Like the remainder goes to the vali owner? That sounds right

shamil-gadelshin: Here is the fix: [Handle case for stake mismatch.](https://github.com/opentensor/subtensor/pull/2053/commits/ab5a274af81fe54d0a47f6d96930c4dece501727)

Swamination-Yuma:
Making a note from OpenDev - to assure this handles dereg liquidation scenarios.

pallets/subtensor/src/swap/swap_hotkey.rs
```diff
@@ -516,15 +516,7 @@ impl<T: Config> Pallet<T> {
...
-        let old_hotkey_tao_dividends = TaoDividendsPerSubnet::<T>::get(netuid, old_hotkey);
-        let new_hotkey_tao_dividends = TaoDividendsPerSubnet::<T>::get(netuid, new_hotkey);
-        TaoDividendsPerSubnet::<T>::remove(netuid, old_hotkey);
-        TaoDividendsPerSubnet::<T>::insert(
-            netuid,
-            new_hotkey,
-            old_hotkey_tao_dividends.saturating_add(new_hotkey_tao_dividends),
-        );
-        weight.saturating_accrue(T::DbWeight::get().reads_writes(2, 2));
+        // Tao dividends were removed
```
camfairchild, line 498: we need to swap over the pools to the new hotkey

pallets/subtensor/src/swap/swap_coldkey.rs
```diff
@@ -188,6 +188,11 @@ impl<T: Config> Pallet<T> {
                 );
                 // Remove the value from the old account.
                 Alpha::<T>::remove((&hotkey, old_coldkey, netuid));
```
camfairchild, line 190: we need to swap over the pool entry here for the coldkey

pallets/subtensor/src/staking/claim_root.rs
```diff
@@ -0,0 +1,338 @@
...
+        for i in coldkeys_to_claim.iter() {
+            weight.saturating_accrue(T::DbWeight::get().reads(1));
+            if let Ok(coldkey) = StakingColdkeysByIndex::<T>::try_get(i) {
+                weight.saturating_accrue(Self::do_root_claim(coldkey.clone()));
+            }
+
+            continue;
+        }
+
+        weight
```
camfairchild, line 343: maybe we should try to grab more if some don't meet the claim threshold

camfairchild (changes requested):
we need to handle subnet de-reg as well. We should just claim the whole pool when the de-reg happens

pallets/subtensor/src/migrations/migrate_remove_tao_dividends.rs
```diff
@@ -0,0 +1,60 @@
...
+        return weight;
+    }
+
+    log::info!(
+        "Running migration '{}'",
+        String::from_utf8_lossy(&migration_name)
+    );
+
+    // Remove obsolete map entries
+    let weight1 = remove_prefix::<T>("TaoDividendsPerSubnet");
```
open-junius, line 47: 1. Should we also remove the PendingRootDivs?
2. Should we distribute the token in  PendingRootDivs and PendingAlphaSwapped? they are swapped, but not go to owner/validator/miner yet.

camfairchild:
Note from nucleus call:
- [ ] make sure auto claim threshold is per-subnet. Don't claim tiny amounts of any subnet automatically

pallets/subtensor/src/migrations/migrate_remove_tao_dividends.rs
shamil-gadelshin: I removed `PendingRootDivs`. We just remove those maps.

pallets/subtensor/src/staking/claim_root.rs
shamil-gadelshin: It introduces a rather awkward interface. I increased the auto-key claims instead. We had roughly a day previously, a week after a conversation with Const. I doubled the value.

pallets/subtensor/src/staking/claim_root.rs
shamil-gadelshin: We could reindex these maps each 6-12month.

shamil-gadelshin (commented):
> make sure auto claim threshold is per-subnet. Don't claim tiny amounts of any subnet automatically

We have one threshold for all operations. Is there a reason to introduce a map of thresholds? 

UPDATE: I implemented the per-subnet threshold anyway.

sam0x17 (dismissed):

sam0x17:
looks like some CI failures tho

Swamination-Yuma:
Has someone communicated with the Bittensor SDK v10 devs to account for any changes (for the auto/manual setting and the storage map adds & removals)? May be nothing, but I was not sure offhand if directly handled since no mention in the v10 migration guide.

shamil-gadelshin:
> Has someone communicated with the Bittensor SDK v10 devs to account for any changes (for the auto/manual setting and the storage map adds & removals)? May be nothing, but I was not sure offhand if directly handled since no mention in the v10 migration guide.

Yes. The SDK team was going to work on the required changes after the merge. cc @basfroman

gztensor (dismissed):

sam0x17 (dismissed):

pallets/subtensor/src/macros/dispatches.rs
```diff
@@ -2309,5 +2311,120 @@ mod dispatches {
...
+        ) -> DispatchResult {
+            let coldkey: T::AccountId = ensure_signed(origin)?;
+
+            Self::maybe_add_coldkey_index(&coldkey);
+
+            Self::change_root_claim_type(&coldkey, new_root_claim_type);
+            Ok(())
+        }
+
+        /// --- Sets the root claim type for the coldkey.
```
bdmason: this description doesn't match the function (same for both function descriptions below)

pallets/subtensor/src/coinbase/root.rs
```diff
@@ -313,8 +315,7 @@ impl<T: Config> Pallet<T> {
         // --- 15. Mechanism step / emissions bookkeeping.
         FirstEmissionBlockNumber::<T>::remove(netuid);
         PendingEmission::<T>::remove(netuid);
-        PendingRootDivs::<T>::remove(netuid);
-        PendingAlphaSwapped::<T>::remove(netuid);
+        PendingRootAlphaDivs::<T>::remove(netuid);
```
antlerminator, line 318: `PendingRootAlphaDivs::<T>::remove(netuid)` is called twice: line 318 & 351

pallets/subtensor/src/coinbase/run_coinbase.rs
```diff
@@ -197,27 +197,14 @@ impl<T: Config> Pallet<T> {
...
-                if let Ok(ok_result) = swap_result {
-                    let root_tao = ok_result.amount_paid_out;
-                    // Accumulate root divs for subnet.
-                    PendingRootDivs::<T>::mutate(*netuid_i, |total| {
-                        *total = total.saturating_add(root_tao);
-                    });
-                }
+                PendingRootAlphaDivs::<T>::mutate(*netuid_i, |total| {
+                    *total = total.saturating_add(tou64!(root_alpha).into());
+                });
```
antlerminator, line 205: This change is not reflected in the description of the 6th step of `run_coinbase` on line 177...

pallets/subtensor/src/coinbase/run_coinbase.rs
```diff
@@ -251,25 +238,15 @@ impl<T: Config> Pallet<T> {
                 PendingEmission::<T>::insert(netuid, AlphaCurrency::ZERO);
 
                 // Get and drain the subnet pending root divs.
```
antlerminator:
```suggestion
                // Get and drain the subnet pending root alpha divs.
```

pallets/subtensor/src/coinbase/run_coinbase.rs
```diff
@@ -251,25 +238,15 @@ impl<T: Config> Pallet<T> {
...
-                let pending_swapped = PendingAlphaSwapped::<T>::get(netuid);
-                PendingAlphaSwapped::<T>::insert(netuid, AlphaCurrency::ZERO);
+                let pending_root_alpha = PendingRootAlphaDivs::<T>::get(netuid);
+                PendingRootAlphaDivs::<T>::insert(netuid, AlphaCurrency::ZERO);
 
                 // Get owner cut and drain.
                 let owner_cut = PendingOwnerCut::<T>::get(netuid);
                 PendingOwnerCut::<T>::insert(netuid, AlphaCurrency::ZERO);
 
                 // Drain pending root divs, alpha emission, and owner cut.
```
antlerminator:
```suggestion
                // Drain pending root alpha divs, alpha emission, and owner cut.
```

pallets/subtensor/src/coinbase/run_coinbase.rs
```diff
@@ -375,22 +352,22 @@ impl<T: Config> Pallet<T> {
         log::debug!("total_root_divs: {total_root_divs:?}");
         log::debug!("total_alpha_divs: {total_alpha_divs:?}");
 
-        // Compute root divs as TAO. Here we take
-        let mut tao_dividends: BTreeMap<T::AccountId, U96F32> = BTreeMap::new();
+        // Compute root alpha divs. Here we take
```
antlerminator, line 355: On line 309 the comment should be also changed to reflect `root divs` -> `root alpha divs`. Consistency is a cool thing!

pallets/subtensor/src/coinbase/run_coinbase.rs
```diff
@@ -550,37 +527,31 @@ impl<T: Config> Pallet<T> {
...
+            let alpha_take: U96F32 =
+                Self::get_hotkey_take_float(&hotkey).saturating_mul(root_alpha);
             // Remove take prop from root_tao
-            root_tao = root_tao.saturating_sub(tao_take);
+            root_alpha = root_alpha.saturating_sub(alpha_take);
             // Give the validator their take.
-            log::debug!("hotkey: {hotkey:?} tao_take: {tao_take:?}");
-            let validator_stake = Self::increase_stake_for_hotkey_and_coldkey_on_subnet(
+            log::debug!("hotkey: {hotkey:?} alpha_take: {alpha_take:?}");
+            let _validator_stake = Self::increase_stake_for_hotkey_and_coldkey_on_subnet(
```
antlerminator, line 539:
```suggestion
           Self::increase_stake_for_hotkey_and_coldkey_on_subnet(
```

Is there a reason we keep the variable here?

pallets/subtensor/src/coinbase/run_coinbase.rs
```diff
@@ -550,37 +527,31 @@ impl<T: Config> Pallet<T> {
...
+
+            Self::increase_root_claimable_for_hotkey_and_subnet(
                 &hotkey,
-                NetUid::ROOT,
-                tou64!(root_tao).into(),
+                netuid,
+                tou64!(root_alpha).into(),
             );
+
             // Record root dividends for this validator on this subnet.
```
antlerminator:
```suggestion
            // Record root alpha dividends for this validator on this subnet.
```

pallets/subtensor/src/coinbase/run_coinbase.rs
```diff
@@ -550,37 +527,31 @@ impl<T: Config> Pallet<T> {
...
-        // Distribute root tao divs.
-        let _ = TaoDividendsPerSubnet::<T>::clear_prefix(netuid, u32::MAX, None);
-        for (hotkey, mut root_tao) in tao_dividends {
+        // Distribute root alpha divs.
+        for (hotkey, mut root_alpha) in root_alpha_dividends {
             // Get take prop
-            let tao_take: U96F32 = Self::get_hotkey_take_float(&hotkey).saturating_mul(root_tao);
+            let alpha_take: U96F32 =
+                Self::get_hotkey_take_float(&hotkey).saturating_mul(root_alpha);
             // Remove take prop from root_tao
```
antlerminator:
```suggestion
            // Remove take prop from root_alpha
```

pallets/subtensor/src/macros/dispatches.rs
```diff
@@ -2309,5 +2311,120 @@ mod dispatches {
...
+        /// * RootClaimTypeSet;
+        /// 	- On the successfully setting the root claim type for the coldkey.
+        ///
+        #[pallet::call_index(123)]
+        #[pallet::weight((
+            Weight::from_parts(4_000_000, 0).saturating_add(T::DbWeight::get().writes(1_u64)),
+            DispatchClass::Operational,
+            Pays::Yes
+        ))]
+        pub fn sudo_set_num_root_claims(origin: OriginFor<T>, new_value: u64) -> DispatchResult {
```
antlerminator: Incorrect description

pallets/subtensor/src/macros/dispatches.rs
```diff
@@ -2309,5 +2311,120 @@ mod dispatches {
...
+        /// * RootClaimTypeSet;
+        /// 	- On the successfully setting the root claim type for the coldkey.
+        ///
+        #[pallet::call_index(124)]
+        #[pallet::weight((
+            Weight::from_parts(5_711_000, 0).saturating_add(T::DbWeight::get().writes(1_u64)),
+            DispatchClass::Operational,
+            Pays::Yes
+        ))]
+        pub fn sudo_set_root_claim_threshold(
```
antlerminator: Incorrect description

pallets/subtensor/src/staking/claim_root.rs
```diff
@@ -0,0 +1,407 @@
...
+        } else {
+            owed.saturating_to_num::<u64>()
+        };
+
+        if owed_u64 == 0 {
+            log::debug!(
+                "root claim on subnet {netuid} is skipped: {owed:?} for h={hotkey:?},c={coldkey:?}"
+            );
+            return; // no-op
+        }
```
antlerminator: This feels off. There's already a helper for clamping owed to 0 that's actually never used...

In addition based on their respective usages, `get_root_owed_for_hotkey_coldkey_float`, `get_root_claimable_for_hotkey_coldkey`, and `get_root_owed_for_hotkey_coldkey` could be merged together to unbloat the file.

pallets/subtensor/src/staking/claim_root.rs
```diff
@@ -0,0 +1,407 @@
...
+    }
+
+    pub fn root_claim_on_subnet(
+        hotkey: &T::AccountId,
+        coldkey: &T::AccountId,
+        netuid: NetUid,
+        root_claim_type: RootClaimTypeEnum,
+        ignore_minimum_condition: bool,
+    ) {
+        // Substract the root claimed.
```
antlerminator:
```suggestion
        // Subtract the root claimed.
```

pallets/subtensor/src/staking/claim_root.rs
```diff
@@ -0,0 +1,407 @@
...
+        coldkey: &T::AccountId,
+        netuid: NetUid,
+    ) -> I96F32 {
+        let claimable = Self::get_root_claimable_for_hotkey_coldkey(hotkey, coldkey, netuid);
+
+        // Attain the root claimed to avoid overclaiming.
+        let root_claimed: I96F32 =
+            I96F32::saturating_from_num(RootClaimed::<T>::get((hotkey, coldkey, netuid)));
+
+        // Substract the already claimed alpha.
```
antlerminator:
```suggestion
        // Subtract the already claimed alpha.
```

pallets/subtensor/src/staking/claim_root.rs
```diff
@@ -0,0 +1,407 @@
...
+            weight.saturating_accrue(Self::root_claim_on_subnet_weight(root_claim_type.clone()));
+        });
+
+        weight
+    }
+
+    pub fn add_stake_adjust_root_claimed_for_hotkey_and_coldkey(
+        hotkey: &T::AccountId,
+        coldkey: &T::AccountId,
+        amount: u64,
```
antlerminator, line 247: Shouldn't this be `AlphaCurrency`, similarly to how it's done in `remove_stake_adjust_root_claimed_for_hotkey_and_coldkey`?

pallets/subtensor/src/staking/helpers.rs
```diff
@@ -326,6 +326,56 @@ impl<T: Config> Pallet<T> {
         });
     }
 
+    /// The function clears Alpha map in batches. Each run will check ALPHA_MAP_CLEAN_BATCH_SIZE
+    /// alphas. It keeps the alpha value stored when it's >= than MIN_ALPHA.
+    /// The function uses AlphaMapCleanLastKey as a storage for key iterator between runs.
```
antlerminator:
```suggestion
    /// The function uses AlphaMapLastKey as a storage for key iterator between runs.
```

pallets/subtensor/src/tests/coinbase.rs
```diff
@@ -1001,7 +947,7 @@ fn test_drain_base_with_subnet_with_two_stakers_registered_and_root_different_am
         SubtensorModule::drain_pending_emission(
             netuid,
             pending_alpha,
-            pending_tao,
+            //   pending_tao,
```
antlerminator:
```suggestion
```

There are multiple issues exactly like that in this file

pallets/subtensor/src/lib.rs
```diff
@@ -1323,13 +1383,9 @@ pub mod pallet {
     /// --- MAP ( netuid ) --> pending_emission
     pub type PendingEmission<T> =
         StorageMap<_, Identity, NetUid, AlphaCurrency, ValueQuery, DefaultPendingEmission<T>>;
-    #[pallet::storage]
     /// --- MAP ( netuid ) --> pending_root_emission
```
antlerminator:
```suggestion
    /// --- MAP ( netuid ) --> pending_root_alpha_emission
```

pallets/subtensor/src/lib.rs
```diff
@@ -1833,6 +1889,53 @@ pub mod pallet {
...
+
+    #[pallet::storage] // --- MAP ( hot ) --> MAP(netuid ) --> claimable_dividends | Root claimable dividends.
+    pub type RootClaimable<T: Config> = StorageMap<
+        _,
+        Blake2_128Concat,
+        T::AccountId,
+        BTreeMap<NetUid, I96F32>,
+        ValueQuery,
+        DefaultRootClaimable<T>,
+    >;
```
antlerminator, line 1907: Shouldn't this be renamed to `RootClaimableRatio` to reflect it better what this exactly is? This wording is already used elsewhere in the code

antlerminator (commented):
Pardon for the intrusion but I've heard this PR was close/ready to be merged yet after reading it I found a bunch of things that could be addressed 😬

pallets/subtensor/src/lib.rs
```diff
@@ -1833,6 +1889,53 @@ pub mod pallet {
         ValueQuery,
     >;
 
+    #[pallet::storage] // --- MAP(netuid ) --> Root claim threshold
+    pub type RootClaimableThreshold<T: Config> =
+        StorageMap<_, Blake2_128Concat, NetUid, I96F32, ValueQuery, DefaultMinRootClaimAmount<T>>;
+
+    #[pallet::storage] // --- MAP ( hot ) --> MAP(netuid ) --> claimable_dividends | Root claimable dividends.
+    pub type RootClaimable<T: Config> = StorageMap<
```
antlerminator: Numbers in this storage variable are increment-only and are cleaned up only when subnet is derreged. Doesn't this lead to unnecessary growth of chain size? Even minimal growth is still net negative

shamil-gadelshin:
> Pardon for the intrusion but I've heard this PR was close/ready to be merged yet after reading it I found a bunch of things that could be addressed 😬

Thanks for the review. I addressed the majority of your comments.

sam0x17 (dismissed):

pallets/subtensor/src/macros/dispatches.rs
```diff
@@ -2309,5 +2311,104 @@ mod dispatches {
...
+            DispatchClass::Operational,
+            Pays::Yes
+        ))]
+        pub fn sudo_set_num_root_claims(origin: OriginFor<T>, new_value: u64) -> DispatchResult {
+            ensure_root(origin)?;
+
+            ensure!(
+                new_value > 0 && new_value <= MAX_NUM_ROOT_CLAIMS,
+                Error::<T>::InvalidNumRootClaim
+            );
```
basfroman:
```suggestion
                new_value <= MAX_NUM_ROOT_CLAIMS,
                Error::<T>::InvalidNumRootClaim
            );
```

We need to allow the option to set NumRootClaim=0, not 1 (as a minimum).
It's very difficult and time-consuming to create 100-1000 "stakers" just to TRY to prevent a single Cold Key from ending up in the random auto claim pool. But even a large number of delegate keys doesn't guarantee a successful test - flaky behavior will still occur.
Allowing the option to set 0 guarantees the ability to test two behaviors: with and without a random auto-claim pool.

Or you could implement some function/constant for this purpose like this
```rust
pub const MIN_NUM_ROOT_CLAIMS: u64 = prod_or_fast!(1, 0);
```

Otherwise, we won't be able to implement 100% stable tests.

shamil-gadelshin: I implemented "zero as acceptable value that disables auto-claim" option.

antlerminator:
Per discussion on the Open Dev call and the overall initiative of improvement of bookkeeping in Bittensor, can we please implement proper bookkeeping from the start for root claim? Back when I was reviewing this PR I saw two possibilities:
- Emit event/events with the exact amounts of alpha that's being received or tao that was received due to selling
- Add new storage variable that tracks the amount of alpha/tao that was just received, e.g., `LastRootClaimed` where for non-root netuids it'd track alpha received and for root netuid it'd track amount of tao received by selling other alphas (this name is not 100% ideal though)

Relevant discussion from CoR: https://discord.com/channels/1120750674595024897/1245487232928714863/1431026852008427572

pallets/subtensor/src/staking/claim_root.rs
```diff
@@ -0,0 +1,420 @@
...
+
+        match root_claim_type {
+            //  Increase stake on root
+            RootClaimTypeEnum::Swap => {
+                // Swap the alpha owed to TAO
+                let owed_tao = match Self::swap_alpha_for_tao(
+                    netuid,
+                    owed_u64.into(),
+                    T::SwapInterface::min_price::<TaoCurrency>(),
+                    false,
```
dougsillars, line 168: so we are charging a swap fee on root stakeholders now?

dougsillars (commented):
it looks like root claim now charges fees to convert to tao?

bdmason:
Hey @shamil-gadelshin it looks like (apologies if I've misunderstood) the following is correct:

1. user unstakes from root
2. `RootClaimed` is _reduced_ by what wasn't claimed but was claimable
3. user can restake and claim what they forgot to claim before step 1

If this is correct, there is likely to be an awful lot of dust that's effectively burned, but not actually burned as it's still technically accessible, though a user may have to earn more divs first to get over the threshold required to make a claim.

Would it not be better to just claim any outstanding divs when the user unstakes? Whilst there would be more operations on the unstake call, there won't be a never ending accumulation of storage in `RootClaimed` for all the dust left behind.

sam0x17 (approved):

pallets/subtensor/src/staking/claim_root.rs
Swamination-Yuma: Adding fees has not been part of the public description of how this would function, afaik. Previous open discussions were that the default would function as it is now (just the option to continue to receive in TAO). The recent Discord announcement does say "If you do nothing, your claims use Swap", but this could be construed as the two options of root_claim_type and not the traditional swap with fees.
Is the intention for all root stakers to pay fees to receive dividends?

shamil-gadelshin:
> Per discussion on the Open Dev call and the overall initiative of improvement of bookkeeping in Bittensor, can we please implement proper bookkeeping from the start for root claim? Back when I was reviewing this PR I saw two possibilities:
> 
> * Emit event/events with the exact amounts of alpha that's being received or tao that was received due to selling
> * Add new storage variable that tracks the amount of alpha/tao that was just received, e.g., `LastRootClaimed` where for non-root netuids it'd track alpha received and for root netuid it'd track amount of tao received by selling other alphas (this name is not 100% ideal though)
> 
> Relevant discussion from CoR: https://discord.com/channels/1120750674595024897/1245487232928714863/1431026852008427572

We plan to update this feature after the release.

leemcmullen:
@shamil-gadelshin The PR description is confusing, it says:

`set_root_claim_type — Set root-claim mode: auto or manual.`

and 

`RootClaimType — User’s preferred claim mode (auto or manual).`

Neither of which are true. 

In fact, `RootClaimType` is either `Swap` or `Keep`. There _is_ an enum with `auto` and `manual` options, but that's called `RootClaimFrequencyEnum` and doesn't seem to be used anywhere else in the subtensor codebase?

antlerminator:
@shamil-gadelshin doing this after the release will mean that the off-chain bookkeeping is impossible between the release and that update... Basically the historical state will be "corrupted" so why not do it now? This is not a huge change in itself but it's impossible to do by non-core devs

PR #2059, opened 2025-09-15
Add pallet-contracts integration with Subtensor chain extension

silva-fj:
## Summary
This PR integrates the `pallet-contracts` module into the Subtensor runtime, enabling wasm smart contract functionality with a custom chain extension for accessing Subtensor-specific features.

### Key Features
  1. **Custom Chain Extension**: The following calls are exposed
    - SubtensorModule::add_stake
    - SubtensorModule::remove_stake
    - SubtensorModule::unstake_all 
    - SubtensorModule::unstake_all_alpha
    - SubtensorModule::move_stake
    - SubtensorModule::transfer_stake
    - SubtensorModule::swap_stake
    - SubtensorModule::add_stake_limit
    - SubtensorModule::remove_stake_limit
    - SubtensorModule::swap_stake_limit
    - SubtensorModule::remove_stake_full_limit
    - SubtensorModule::set_coldkey_auto_stake_hotkey
    - Proxy::add_proxy (Staking)
    - Proxy::remove_proxy (Staking)

  2. **Restricted Call Filter**: Whitelisted operations
       - `Proxy::proxy`
   
  3. **Conservative Configuration**
     - Memory limits: 1GB for runtime, 2GB for validators
     - Maximum code size: 128KB
     - Call stack depth: 5 frames
     - Transient storage: 1MB
 
 ## Breaking Changes

  None - This is an additive change that doesn't affect existing functionality.


masterdubs:
This proposal is groundbreaking, it open lot of native doors integration.
As soon as it is merged we will use it in Hippius for alpha handling.

docs/contracts.md
```diff
@@ -0,0 +1,66 @@
...
+```
+
+### Call Filter
+
+For security, contracts can only dispatch a limited set of runtime calls:
+
+**Whitelisted Calls:**
+- `SubtensorModule::move_stake` - Move stake between hotkeys
+- `SubtensorModule::transfer_stake` - Transfer stake between accounts
+- `Proxy::proxy` - Execute proxy calls
```
camfairchild, line 90: We should whitelist a lot more calls than this

bdmason: Agreed, but we spoke to @gregzaitsev about being cautious in the approach, adding them as needed at first. This is just what we need for our current project.

runtime/src/lib.rs
```diff
@@ -1590,6 +1591,76 @@ impl pallet_crowdloan::Config for Runtime {
...
+            validator_runtime_memory: 1024 * 1024 * 1024 * 2,
+            ..Default::default()
+        },
+        ..Default::default()
+    }
+}
+
+parameter_types! {
+    pub const ContractDepositPerItem: Balance = deposit(1, 0);
+    pub const ContractDepositPerByte: Balance = deposit(0, 1);
```
camfairchild: This is using the proxy/multisig deposit formula
So around `0.001 TAO` per byte and `0.02 TAO` per storage item.

docs/contracts.md
camfairchild: We should allow all the `SubtensorModule::*stake*` extrinsics and `Balances::*` extrinsics

Also `Proxy::add_proxy` and `Proxy::create_pure`

runtime/src/lib.rs
```diff
@@ -1590,6 +1591,76 @@ impl pallet_crowdloan::Config for Runtime {
...
+    pub const ContractDefaultDepositLimit: Balance = deposit(1024, 1024 * 1024);
+    pub ContractsSchedule: pallet_contracts::Schedule<Runtime> = contracts_schedule::<Runtime>();
+    pub const CodeHashLockupDepositPercent: Perbill = Perbill::from_percent(0);
```
camfairchild: This needs to be non-zero per docs: https://docs.rs/pallet-contracts/latest/pallet_contracts/pallet/trait.Config.html#associatedtype.CodeHashLockupDepositPercent

This example has it at 30pc
https://github.com/paritytech/substrate/blob/033d4e86cc7eff0066cd376b9375f815761d653c/bin/node/runtime/src/lib.rs#L1292

camfairchild (changes requested):

runtime/src/lib.rs
silva-fj: Thanks for reviewing, I was hoping for feedback on this (as well as the deposit config) before making the PR ready for review as these specific to the chain's economy. I've seen a few examples of production chains that use different pc values but it seems that 30 is the default reference 👍🏼

runtime/src/lib.rs
silva-fj: I have updated the deposit format to mirror the substrate example reference: 

_15% of the existential deposit per key, 6% per byte._

silva-fj:
> We should allow all the SubtensorModule::*stake* extrinsics and Balances::* extrinsics
Also Proxy::add_proxy and Proxy::create_pure

I've included the staking related dispatchables to the whitelist. The pallet-contracts exposes some host functions related to balances, e.g: https://docs.rs/pallet-contracts/latest/pallet_contracts/api_doc/trait.Current.html#tymethod.transfer, if there is a need for something specific from `Balances`, please let me know.

Keep in mind that when `call_runtime` is executed in a smart contract, the origin of the extrinsic is the contract (account) itself and not the account calling the contract.

docs/contracts.md
silva-fj: I've included the staking related dispatchables to the whitelist. The pallet-contracts exposes some host functions related to balances, e.g: https://docs.rs/pallet-contracts/latest/pallet_contracts/api_doc/trait.Current.html#tymethod.transfer, if there is a need for something specific from Balances, please let me know.

Keep in mind that when call_runtime is executed in a smart contract, the origin of the extrinsic is the contract (account) itself and not the account calling the contract.

camfairchild:
I saw something in the docs saying the APIs for the extrinsics we allow has to remain stable.

We solve this in the frontier (EVM) pallet by using precompiles with stable APIs but the underlying extrinsics can change.

Possibly we should do the same because I'm not sure if we can guarantee the stability of any of the extrinsics at-present

camfairchild:
Also, @silva-fj let's actually remove the `proxy::create_pure` because I'm uncertain how the contract would be able to grab the resultant Pure proxy account that was created. It's only in an event created after the fact

silva-fj:
> I saw something in the docs saying the APIs for the extrinsics we allow has to remain stable.
> 
> We solve this in the frontier (EVM) pallet by using precompiles with stable APIs but the underlying extrinsics can change.
> 
> Possibly we should do the same because I'm not sure if we can guarantee the stability of any of the extrinsics at-present

Yes that's correct, the whitelisted dispatchables should remain stable. This is not exactly as critical as on EVM contracts because ink contracts can be upgradable (just like the runtime). But ideally a stable API should be provided. 

This can be accomplished with chain extensions, by exposing an API that remains stable while the underlying dispatchables could be changed

camfairchild:
> This is not exactly as critical as on EVM contracts because ink contracts can be upgradable (just like the runtime).

EVM contracts use an upgradeable pattern as well. 
 
> This can be accomplished with chain extensions, by exposing an API that remains stable while the underlying dispatchables could be changed

Possibly we can also just agree that the API is not stable and any contracts implementing the calls can check the runtime version to see if it has changed/if their current contract version (self-imposed) supports it.

i.e. like a guard clause

sam0x17:
are you satisfied with this as is @camfairchild ?

sam0x17:
just a few tiny merge conflicts

silva-fj:
> Possibly we can also just agree that the API is not stable and any contracts implementing the calls can check the runtime version to see if it has changed/if their current contract version (self-imposed) supports it.

I agree with that, contract developers using unstable APIs should know that they need to handle potential breaking changes. Just checking the runtime version would not be enough as you can't tell if a specific API has changed based on that, but making the contract upgradable, e.g: using [set_code](https://docs.rs/pallet-contracts/latest/pallet_contracts/pallet/enum.Call.html#variant.set_code) would solve the problem

camfairchild:
> > Possibly we can also just agree that the API is not stable and any contracts implementing the calls can check the runtime version to see if it has changed/if their current contract version (self-imposed) supports it.
> 
> I agree with that, contract developers using unstable APIs should know that they need to handle potential breaking changes. Just checking the runtime version would not be enough as you can't tell if a specific API has changed based on that, 

Why wouldn't the runtime version signal a potential breaking change? I get it wouldn't be specific but it is the thing that signals a change to the runtime.

> but making the contract upgradable, e.g: using [set_code](https://docs.rs/pallet-contracts/latest/pallet_contracts/pallet/enum.Call.html#variant.set_code) would solve the problem

Also, what about contracts that aren't upgradeable. There is going to be a few for sure. 
My point in raising this concern was that we offer a *more stable* API for the EVM and right now nothing like that for `pallet-contracts` but I do think that would be super beneficial

silva-fj:
> Why wouldn't the runtime version signal a potential breaking change? I get it wouldn't be specific but it is the thing that signals a change to the runtime.
> 

Sorry, that's not what I said so I'm not sure I understand your comment. Maybe you can provide an example of how a contract can use the runtime version to workaround a breaking change in the API. Let's say you create a contract that uses "Foo" API that is not stable and deploy it on chain, at the time the runtime version is "X". Then, there is a runtime upgrade and the new version is "Y", what should the contract do?

> Also, what about contracts that aren't upgradeable. There is going to be a few for sure. 

Non-upgradable contracts should not use APIs that are known to be "unstable", but as with any contract, that's up to the developer.

> My point in raising this concern was that we offer a _more stable_ API for the EVM and right now nothing like that for `pallet-contracts` but I do think that would be super beneficial

I think that's a valid concern and a good idea. I mentioned before that with pallet-contracts you can expose chain extensions to accomplish the same of what's available with the precompiles for EVM and you said this: 

_"Possibly we can also just agree that the API is not stable and any contracts implementing the calls can check the runtime version to see if it has changed/if their current contract version (self-imposed) supports it."_ 

Which in my opinion it is fine, if a contract will use an API that is known to be unstable or that can change in the future, it should also support `set_code` to upgrade the contract's code accordingly. 

It would be helpful if you or anyone else involved could provide the minimum requirements for getting this going so we can adjust it and get it done as it will benefit many people. I don't know what would be useful to the community upfront, so I would suggest adding functionality incrementally as needed and not exposing everything from the start.

camfairchild:
> > Why wouldn't the runtime version signal a potential breaking change? I get it wouldn't be specific but it is the thing that signals a change to the runtime.
> > 
> 
> Sorry, that's not what I said so I'm not sure I understand your comment. Maybe you can provide an example of how a contract can use the runtime version to workaround a breaking change in the API. Let's say you create a contract that uses "Foo" API that is not stable and deploy it on chain, at the time the runtime version is "X". Then, there is a runtime upgrade and the new version is "Y", what should the contract do?

They could pause until they upgrade. But ya that's not a good change for a lot of the contract developers (non upgradable).

> > My point in raising this concern was that we offer a _more stable_ API for the EVM and right now nothing like that for `pallet-contracts` but I do think that would be super beneficial
> 
> I think that's a valid concern and a good idea. I mentioned before that with pallet-contracts you can expose chain extensions to accomplish the same of what's available with the precompiles for EVM and you said this:  
> _"Possibly we can also just agree that the API is not stable and any contracts implementing the calls can check the runtime version to see if it has changed/if their current contract version (self-imposed) supports it."_ 
> 
> Which in my opinion it is fine, if a contract will use an API that is known to be unstable or that can change in the future, it should also support `set_code` to upgrade the contract's code accordingly. 

Ya it's just for the contracts that don't use upgradable, which a lot will, as is common in the space.

> It would be helpful if you or anyone else involved could provide the minimum requirements for getting this going so we can adjust it and get it done as it will benefit many people. I don't know what would be useful to the community upfront, so I would suggest adding functionality incrementally as needed and not exposing everything from the start.

imo, we provide a versioned interface. We've been discussing this in CoR as well.

docs/contracts.md
```diff
@@ -0,0 +1,105 @@
...
+| 14 | `InsufficientLiquidity` | Insufficient liquidity for swap operation |
+| 15 | `SameNetuid` | Source and destination subnets are the same |
+| 16 | `ProxyTooMany` | Too many proxies registered |
+| 17 | `ProxyDuplicate` | Proxy already exists |
+| 18 | `ProxyNoSelfProxy` | Cannot add self as proxy |
+| 19 | `ProxyNotFound` | Proxy relationship not found |
+
+### Call Filter
+
+For security, contracts can only dispatch a limited set of runtime calls:
```
camfairchild:
```suggestion
For security, contracts can only directly dispatch a limited set of runtime calls:
```
This wording would be more clear

docs/contracts.md
```diff
@@ -0,0 +1,105 @@
+# Smart Contracts on Subtensor
```
camfairchild:
```suggestion
# WebAssembly Smart Contracts
```

docs/contracts.md
```diff
@@ -0,0 +1,105 @@
...
+
+## Overview
+
+Subtensor now supports smart contract functionality through the integration of `pallet-contracts`, enabling developers to deploy and execute WebAssembly (WASM) smart contracts on the network. Contracts are written in [ink!](https://use.ink/), a Rust-based embedded domain-specific language (eDSL) for writing smart contracts on Substrate-based chains.
```
camfairchild:
```suggestion
Subtensor now supports WebAssembly (WASM) smart contract functionality through the integration of `pallet-contracts`, enabling developers to deploy and execute WASM smart contracts on the network. Contracts are written in [ink!](https://use.ink/), a Rust-based embedded domain-specific language (eDSL) for writing smart contracts on Substrate-based chains. For compatibility, WASM contracts can also be compiled from Solidity using [Solang](https://github.com/hyperledger-solang/solang).
```

docs/wasm-contracts.md
```diff
@@ -0,0 +1,105 @@
+# Smart Contracts on Subtensor
+
+## Overview
+
+Subtensor now supports smart contract functionality through the integration of `pallet-contracts`, enabling developers to deploy and execute WebAssembly (WASM) smart contracts on the network. Contracts are written in [ink!](https://use.ink/), a Rust-based embedded domain-specific language (eDSL) for writing smart contracts on Substrate-based chains.
+
```
camfairchild, line 9:
```suggestion
> [!NOTE]  
> If you're looking for information on EVM contracts, please see the documentation: https://docs.learnbittensor.org/evm-tutorials
```
Not sure if we have a better place to link to

docs/contracts.md
camfairchild: Possibly rename the file to `wasm-contracts.md` to avoid confusion

sam0x17 (dismissed):

sam0x17:
clippy failing

sam0x17 (approved):

ligdeve:
can we expose more functions , like what evm precompiles have done?
at leatst these two functions are very useful when calculating alpha price in the wasm contract.
```
    function simSwapTaoForAlpha(
        uint16 netuid,
        uint64 tao
    ) external view returns (uint256);

    function simSwapAlphaForTao(
        uint16 netuid,
        uint64 alpha
    ) external view returns (uint256);

```

PR #2073, opened 2025-09-22
Fix/Use `netuid_index` in selective metagraph

JohnReedV:
## Description
This PR fixes a bug where netuid was incorrectly used in the metagraph instead of `netuid_index`.

### DIFF
--- a/pallets/subtensor/src/rpc_info/metagraph.rs
+++ b/pallets/subtensor/src/rpc_info/metagraph.rs
@@ -1453,7 +1453,7 @@ impl<T: Config> Pallet<T> {
             Some(SelectiveMetagraphIndex::Incentives) => SelectiveMetagraph {
                 netuid: netuid.into(),
                 incentives: Some(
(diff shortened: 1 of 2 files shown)


pallets/subtensor/src/rpc_info/metagraph.rs
```diff
@@ -1453,7 +1453,7 @@ impl<T: Config> Pallet<T> {
             Some(SelectiveMetagraphIndex::Incentives) => SelectiveMetagraph {
                 netuid: netuid.into(),
                 incentives: Some(
-                    Incentive::<T>::get(NetUidStorageIndex::from(netuid))
+                    Incentive::<T>::get(NetUidStorageIndex::from(netuid_index))
```
gztensor: netuid_index is already `NetUidStorageIndex`, no need for this type conversion here.

gztensor (changes requested):

gztensor (approved):

l0r1s (approved):

PR #2114, opened 2025-10-08
Remove Duplicate HyperParameter setter

JohnReedV:
## Description
This PR removes a duplicate hyperparameter extrinsic in admin-utils.


pallets/admin-utils/src/tests/mod.rs
```diff
@@ -1745,48 +1745,48 @@ fn test_sets_a_lower_value_clears_small_nominations() {
...
+//             AdminUtils::sudo_set_subnet_owner_hotkey(random_account, netuid, new_hotkey),
+//             DispatchError::BadOrigin
+//         );
+
+//         assert_noop!(
+//             AdminUtils::sudo_set_subnet_owner_hotkey(root, netuid, new_hotkey),
+//             DispatchError::BadOrigin
+//         );
+//     });
+// }
```
ppolewicz, line 1756:
```suggestion
```
Trust VCS. If you'll ever need this code back, it will be in the repo.

pallets/admin-utils/src/lib.rs
```diff
@@ -1686,11 +1686,7 @@ pub mod pallet {
             netuid: NetUid,
             hotkey: <T as frame_system::Config>::AccountId,
         ) -> DispatchResult {
-            pallet_subtensor::Pallet::<T>::ensure_subnet_owner(origin.clone(), netuid)?;
-            pallet_subtensor::Pallet::<T>::set_subnet_owner_hotkey(netuid, &hotkey);
-
-            log::debug!("SubnetOwnerHotkeySet( netuid: {netuid:?}, hotkey: {hotkey:?} )");
-            Ok(())
+            pallet_subtensor::Pallet::<T>::do_set_sn_owner_hotkey(origin, netuid, &hotkey)
         }
```
ppolewicz, line 1660:
```suggestion
```
There is another function that does the same thing in the same file, one has "sn" in the name and the other has "_subnet_". Do we need both? What for?

I'm not sure which one Cortex uses, I'd keep that one and remove the other one.

ppolewicz:
Actually, I looked at the copy of the code on the `main` branch - perhaps there were some changes earlier in `devnet-ready` that this PR is cleaning up, but it was not linked?

pallets/admin-utils/src/tests/mod.rs
JohnReedV: I commented it out instead of deleting it to make it more difficult for somebody to re-use the extrinsic index. Substrate doesn't like that.

pallets/admin-utils/src/lib.rs
JohnReedV: Cortex uses neither at the moment.

sam0x17 (approved):

l0r1s (approved):

PR #2122, opened 2025-10-09
Add linter rule to flag expect. Remove expects/allow exceptions.

gztensor:
## Description

Avoid using `expect` anywhere in the on-chain code.

## Type of Change
- [ ] Documentation update

## Checklist

### DIFF
--- a/pallets/proxy/src/lib.rs
+++ b/pallets/proxy/src/lib.rs
@@ -455,7 +455,7 @@ pub mod pallet {
                     T::AnnouncementDepositFactor::get(),
                     pending.len(),
                 )?
-                .ok_or(Error::<T>::AnnouncementDepositInvariantViolated)?; // <- new error instead of expect
+                .ok_or(Error::<T>::AnnouncementDepositInvariantViolated)?;
 
                 *deposit = new_deposit;
                 Ok::<(), DispatchError>(())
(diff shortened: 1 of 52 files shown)


sam0x17 (dismissed):

pallets/proxy/src/lib.rs
```diff
@@ -113,6 +113,7 @@ pub enum DepositKind {
 }
 
 #[frame::pallet]
+#[allow(clippy::expect_used)]
```
lorismoulin, line 116: I'm trying to grasp why this is an `allow` here and not a `deny` with specific `allow` where needed?

gztensor: Because if I deny, it flags `expect` used in FRAME macros, so I can't allow them in place. This way I can allow everywhere, so it doesn't flag FRAME macros, but I can deny on the implementation level, which still covers our code.

gztensor: Globally it is denied in workspace settings in Cargo.toml

lorismoulin: Make sense, thanks!

l0r1s (approved):

sam0x17 (approved):

PR #2126, opened 2025-10-10
Remove governance deadcode

lorismoulin:
This PR removes the dead code from legacy and non-working governance system in preparation for the new governance system.


ales-otf (dismissed):

runtime/src/lib.rs
```diff
@@ -1685,12 +1510,22 @@ pub type TransactionExtensions = (
...
+}
+
 type Migrations = (
     // Leave this migration in the runtime, so every runtime upgrade tiny rounding errors (fractions of fractions
     // of a cent) are cleaned up. These tiny rounding errors occur due to floating point coversion.
     pallet_subtensor::migrations::migrate_init_total_issuance::initialise_total_issuance::Migration<
         Runtime,
     >,
+    // Remove storage from removed governance pallets
+    frame_support::migrations::RemovePallet<TriumviratePalletStr, RocksDbWeight>,
```
shamil-gadelshin, line 1526: Why do we set RockDBWeight?

sam0x17 (dismissed):

ales-otf (dismissed):

runtime/src/lib.rs
lorismoulin: @shamil-gadelshin Good catch, all the weights across the code base are RocksDbWeight but we migrated to ParityDb as default, I will open an issue so we can handle this in another PR

shamil-gadelshin (dismissed):

basfroman:
Fix failed e2e test from SDK https://github.com/opentensor/bittensor/pull/3101

sam0x17 (approved):

open-junius (approved):

PR #2139, opened 2025-10-16
kappa root only

JohnReedV:
## Description
Make the `Kappa` hyper-parameter only settable by root.
Migrate all `Kappa` entries back to the default value of `0.5`.

### DIFF
--- a/pallets/admin-utils/src/lib.rs
+++ b/pallets/admin-utils/src/lib.rs
@@ -581,23 +581,13 @@ pub mod pallet {
         .saturating_add(<T as frame_system::Config>::DbWeight::get().reads(3_u64))
         .saturating_add(<T as frame_system::Config>::DbWeight::get().writes(1_u64)))]
         pub fn sudo_set_kappa(origin: OriginFor<T>, netuid: NetUid, kappa: u16) -> DispatchResult {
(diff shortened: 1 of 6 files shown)


pallets/admin-utils/src/lib.rs
```diff
@@ -543,27 +543,17 @@ pub mod pallet {
...
+        .saturating_add(<T as frame_system::Config>::DbWeight::get().reads(1_u64))
         .saturating_add(<T as frame_system::Config>::DbWeight::get().writes(1_u64)))]
         pub fn sudo_set_kappa(origin: OriginFor<T>, netuid: NetUid, kappa: u16) -> DispatchResult {
-            let maybe_owner = pallet_subtensor::Pallet::<T>::ensure_sn_owner_or_root_with_limits(
-                origin,
-                netuid,
-                &[Hyperparameter::Kappa.into()],
-            )?;
-
+            ensure_root(origin)?;
```
lorismoulin, line 550: Are we allowed to execute this in the last 10 blocks of a tempo @ppolewicz ? If no, there is the `ensure_root_with_rate_limit`, name is strange but it only allows in the freeze window, there is no rate limit

l0r1s (approved):

sam0x17 (approved):

PR #2140, opened 2025-10-16
Global subsidies based on ema prices

gztensor:
## Description

No root sell when EMA prices sum up to lower than 1.0.

## Type of Change
- [ ] Documentation update

## Checklist

### DIFF
--- a/pallets/subtensor/src/coinbase/run_coinbase.rs
+++ b/pallets/subtensor/src/coinbase/run_coinbase.rs
@@ -54,6 +54,7 @@ impl<T: Config> Pallet<T> {
         // --- 3. Get subnet terms (tao_in, alpha_in, and alpha_out)
         // Computation is described in detail in the dtao whitepaper.
         let mut tao_in: BTreeMap<NetUid, U96F32> = BTreeMap::new();
+        let mut tao_issued: BTreeMap<NetUid, U96F32> = BTreeMap::new();
         let mut alpha_in: BTreeMap<NetUid, U96F32> = BTreeMap::new();
         let mut alpha_out: BTreeMap<NetUid, U96F32> = BTreeMap::new();
         // Only calculate for subnets that we are emitting to.
@@ -115,6 +116,7 @@ impl<T: Config> Pallet<T> {
             }
             // Insert values into maps
             tao_in.insert(*netuid_i, tao_in_i);
+            tao_issued.insert(*netuid_i, default_tao_in_i);
             alpha_in.insert(*netuid_i, alpha_in_i);
             alpha_out.insert(*netuid_i, alpha_out_i);
         }
@@ -150,10 +152,13 @@ impl<T: Config> Pallet<T> {
             TotalStake::<T>::mutate(|total| {
                 *total = total.saturating_add(tao_in_i.into());
             });
+            // TAO issued
+            let tao_issued_i: TaoCurrency =
+                tou64!(*tao_issued.get(netuid_i).unwrap_or(&asfloat!(0))).into();
             TotalIssuance::<T>::mutate(|total| {
-                *total = total.saturating_add(tao_in_i.into());
+                *total = total.saturating_add(tao_issued_i.into());
             });
-            // Adjust protocol liquidity based on new reserves
+            // Adjust protocol liquidity based added reserves
             T::SwapInterface::adjust_protocol_liquidity(*netuid_i, tao_in_i, alpha_in_i);
         }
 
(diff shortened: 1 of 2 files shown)


camfairchild (dismissed):

pallets/subtensor/src/tests/coinbase.rs
camfairchild, line 288: the issuance should increase by the full emission, no? We should still be emitting the same amount of TAO. It's just where it goes that is different.

camfairchild (dismissed):

sam0x17 (approved):

pallets/subtensor/src/coinbase/run_coinbase.rs
```diff
@@ -62,11 +75,7 @@ impl<T: Config> Pallet<T> {
             // Get initial alpha_in
             let mut alpha_in_i: U96F32;
             let mut tao_in_i: U96F32;
-            let tao_in_ratio: U96F32 = default_tao_in_i.safe_div_or(
-                U96F32::saturating_from_num(block_emission),
-                U96F32::saturating_from_num(0.0),
-            );
-            if price_i < tao_in_ratio {
+            if subsidy_mode {
                 tao_in_i = price_i.saturating_mul(U96F32::saturating_from_num(block_emission));
```
mcjkula, line 79: The `tao_in_i` should only be changed for subnets where ALPHA price < TAO emissions (EMA / sum of EMAs), otherwise it would inject more TAO, and would result in more than 1 TAO/block being emitted. The only reason to change the `tao_in_i` according to the `ALPHA price * Block Emission` is so that one can calculate how much TAO should be used to swap for ALPHA (to bring the ALPHA price back up).

Here a scenario:
> SoS is 0.8 and there is a subnet A that is overpriced, so ALPHA price is 0.5 while TAO emissions are 0.25, the chain would inject 0.5 TAO into the subnet because it looks at `ALPHA price * Block Emission`, instead of the 0.25 TAO that he should get based on TAO emissions. The `difference_tao_in` would clamp to 0 because it would be otherwise negative (`tao_in > default_tao_in`). Then subnet B has 0.75 TAO emissions but an ALPHA price of 0.3, so the chain would use 0.3 TAO to inject, and 0.45 TAO to swap for ALPHA to push ALPHA price towards TAO emissions (still using 0.75 TAO in total).

> Thus subnet A gets 0.5 TAO and subnet B gets 0.75 TAO, so the chain emits 1.25 TAO/block

If one were to use ALPHA prices only as commented at line 38 I think this wouldn't happen, because then there is no discrepancy between ALPHA price and EMA price, as only ALPHA price is used.

pallets/subtensor/src/coinbase/run_coinbase.rs
```diff
@@ -198,8 +209,7 @@ impl<T: Config> Pallet<T> {
             let pending_alpha: U96F32 = alpha_out_i.saturating_sub(root_alpha);
             log::debug!("pending_alpha: {pending_alpha:?}");
             // Sell root emission through the pool (do not pay fees)
-            let subsidized: bool = *is_subsidized.get(netuid_i).unwrap_or(&false);
-            if !subsidized {
+            if !subsidy_mode {
```
mcjkula, line 212: That could technically stay so that the root proportion on "overpriced" subnets (ALPHA price > TAO emissions) is not countering the SoS to increase at/above 1

pallets/subtensor/src/coinbase/run_coinbase.rs
```diff
@@ -103,6 +110,7 @@ impl<T: Config> Pallet<T> {
             }
             // Insert values into maps
             tao_in.insert(*netuid_i, tao_in_i);
+            tao_issued.insert(*netuid_i, default_tao_in_i);
```
mcjkula, line 113: As commented above that wouldn't be true, when ALPHA price > TAO emissions, with the global subsidy, because the `tao_in_i` (`ALPHA price * Block Emission`) would be larger than `default_tao_in_i` (which is based on EMA price).

pallets/subtensor/src/coinbase/run_coinbase.rs
mcjkula: The reason to downscale TAO is if ALPHA needs to be downscaled, so that the injections can maintain the price, while the `difference_tao_in` is used to swap for ALPHA (increasing the ALPHA price towards TAO emissions), which it doesn't when ALPHA price > TAO emissions

mcjkula (changes requested):

gztensor:
Closing in favor of https://github.com/opentensor/subtensor/pull/2096 fix

PR #2145, opened 2025-10-20
Random Transaction Ordering

JohnReedV:
## Description
Randomizes the transaction selection order **per block**.  
On each proposal the node derives a per-block salt from a secure random seed and the parent block hash, then the pool orders ready txs by `blake2_256(salt || hash)`.  

**Requires Node Upgrade**


sam0x17 (dismissed):

sam0x17 (approved):

sam0x17:
closing because unfortunately 0% slippage makes this more exploitable than the regular state

PR #2150, opened 2025-10-21
Add pallet-rate-limiting

ales-otf:
## Description

This PR introduces `pallet-rate-limiting` as a uniform rate-limiting solution and migrates a part of legacy rate limiting to this pallet.

Rate limits are set with `pallet_rate_limiting::set_rate_limit(target, scope, limit)`. `target` is either one call or a group of calls. `scope` is optional context used to select the configured span (for example, `netuid`). If a call/group does not need context, it can be configured directly at runtime with `target + limit` and no resolver data.

There are standalone calls and groups of calls. Groups are defined at runtime, and each call can be either standalone or within a group, but not both. Groups allow multiple calls to share rate-limiting behavior. Depending on mode (`ConfigOnly`, `UsageOnly`, `ConfigAndUsage`), calls share config, usage tracking, or both.

For calls that need additional context, resolvers provide it. Limits can be configured either globally per target, or scoped per target+scope. The role of `ScopeResolver` is to provide that scope context (for example, `netuid`) so the extension can select the correct scoped limit entry. `ScopeResolver` can also adjust span (for example, tempo scaling) and define bypass behavior. `UsageResolver` resolves usage key(s) so `LastSeen` is tracked with additional context (for example, per account/per subnet/per mechanism), not only by target.

Enforcement happens via `UnwrappedRateLimitTransactionExtension`. It first unwraps nested calls (`sudo`, `proxy`, `utility`, `multisig`), then delegates each inner call to `RateLimitTransactionExtension`. The extension checks the resolved target/scope and compares current block vs `LastSeen`. If within span, validation fails with `InvalidTransaction::Custom(1)`. On successful dispatch, the extension writes `LastSeen` for resolved usage key(s). This is what enforces rate limiting for subsequent calls.

Other pallets should use `rate-limiting-interface` (`RateLimitingInterface`) to read limits and last-seen state without depending on pallet internals. Writes through this interface should be avoided as much as possible because they introduce side-effects. The expected write path is the transaction extension itself. Manual `set_last_seen` writes are only for cases where usage must be updated outside the normal rate-limited call path.

`pallet-rate-limiting` is instanceable and is intended to be used as one instance per pallet, with local scope/usage types and resolvers. This runtime currently uses one centralized instance for `pallet-subtensor` + `pallet-admin-utils` as a transitional setup. Migration and resolvers already exist for grouped and standalone legacy limits, but this PR migrates only grouped legacy limits (`GroupedRateLimitingMigration`). Standalone migration/cleanup is deferred to a follow-up PR.

### How to review:

I tried to organize this PR so that each major change is represented by a single commit, while avoiding meaningless commits as much as possible. You can refer to the commit history and review changes related to specific limits.

1. To review how data is migrated: `runtime/src/migrations/rate_limiting::commits_grouped`. This is where you'll find everything considered "legacy" rate-limiting. You can then review each migration from the list separately.
2. To review whether behavior is covered correctly: the migration above + `runtime/src/rate_limiting`. There you'll find the resolver implementations and can review how different calls are bypassed, adjusted to tempo, and scoped (for limits: only by `NetUid`; for last-seen timestamp: various cases).
3. For proof of correctness and behavior: `runtime/tests/rate_limiting`. These are integration tests at the extrinsic level, where transaction extensions are involved. You can verify whether rate-limiting behavior is consistent with legacy behavior and whether all cases are covered.

## Type of Change

- [ ] Documentation update

### Deprecated extrinsics with their equivalents in `pallet-rate-limiting`

| extrinsic (`pallet-admin-utils`) | `pallet-rate-limiting::set_rate_limit` params |
|---|---|
| `sudo_set_tx_rate_limit(tx_rate_limit)` | `target = Group(GROUP_SWAP_KEYS), scope = None, limit = Exact(tx_rate_limit)` |
| `sudo_set_serving_rate_limit(netuid, serving_rate_limit)` | `target = Group(GROUP_SERVE), scope = Some(netuid), limit = Exact(serving_rate_limit)` |
| `sudo_set_weights_set_rate_limit(netuid, weights_set_rate_limit)` | `target = Group(GROUP_WEIGHTS_SET), scope = Some(netuid), limit = Exact(weights_set_rate_limit)` |
| `sudo_set_network_rate_limit(limit)` | `target = Group(GROUP_REGISTER_NETWORK), scope = None, limit = Exact(limit)` |
| `sudo_set_tx_delegate_take_rate_limit(tx_rate_limit)` | `target = Group(GROUP_DELEGATE_TAKE), scope = None, limit = Exact(tx_rate_limit)` |
| `sudo_set_owner_hparam_rate_limit(epochs)` | `target = Group(GROUP_OWNER_HPARAMS), scope = Some(netuid), limit = Exact(epochs)` |

You can find the values of `GROUP_*` constants in `common/src/rate_limiting.rs`.

From the client's perspective, you can query `pallet-rate-limiting::Groups` storage to list all groups with their IDs and configuration, or `pallet-rate-limiting::GroupNameIndex` to get the ID of a particular group by name.

### Removed storages from `pallet-subtensor`

On the client side, use `pallet-rate-limiting::Limits` storage to fetch limits, or `pallet-rate-limiting::LastSeen` to fetch last-seen timestamps.

- `NetworkRateLimit` -> `Limits({ Group: GROUP_REGISTER_NETWORK })`
- `OwnerHyperparamRateLimit` -> `Limits({ Group: GROUP_OWNER_HPARAMS })`
- `ServingRateLimit` -> `Limits({ Group: GROUP_SERVE })` then scoped value for `netuid`
- `StakingOperationRateLimiter` -> `LastSeen({ Group: GROUP_STAKING_OPS }, { ColdkeyHotkeySubnet: { coldkey, hotkey, netuid } })`
- `TxDelegateTakeRateLimit` -> `Limits({ Group: GROUP_DELEGATE_TAKE })`
- `TxRateLimit` -> `Limits({ Group: GROUP_SWAP_KEYS })`
- `WeightsSetRateLimit` -> `Limits({ Group: GROUP_WEIGHTS_SET })` then scoped value for `netuid`

### `pallet-subtensor::Config` changes

- added `type RateLimiting: RateLimitingInterface`
- removed:
  - `InitialServingRateLimit`
  - `InitialTxRateLimit`
  - `InitialTxDelegateTakeRateLimit`
  - `InitialNetworkRateLimit`

### Removed events

They moved to `pallet-rate-limiting::RateLimitSet` event like this:

- `NetworkRateLimitSet` -> `{ target: Group(GROUP_REGISTER_NETWORK), scope: None, limit: Exact(span) }`
- `OwnerHyperparamRateLimitSet` -> `{ target: Group(GROUP_OWNER_HPARAMS), scope: None, limit: Exact(epochs) }`
- `ServingRateLimitSet` -> `{ target: Group(GROUP_SERVE), scope: Some(netuid), limit: Exact(span) }`
- `TxDelegateTakeRateLimitSet` -> `{ target: Group(GROUP_DELEGATE_TAKE), scope: None, limit: Exact(span) }`
- `TxRateLimitSet` -> `{ target: Group(GROUP_SWAP_KEYS), scope: None, limit: Exact(span) }`
- `WeightsSetRateLimitSet` -> `{ target: Group(GROUP_WEIGHTS_SET), scope: Some(netuid), limit: Exact(span) }`

### Additional changes

`get_network_lock_cost()` now uses `T::RateLimiting::last_seen(GROUP_REGISTER_NETWORK, None)` instead of legacy `RateLimitKey::NetworkLastRegistered`.

`pallet_admin_utils::sudo_set_tx_rate_limit` -> `pallet_rate_limiting::set_rate_limit(... GROUP_SWAP_KEYS)` is now **limit + 1**:

- before (`pallet_admin_utils::sudo_set_tx_rate_limit(N)`) - **delta <= limit**: a swap was still blocked when exactly `N` blocks had passed, and it became allowed only after `N + 1` blocks (**delta > limit**);
- now (`pallet_rate_limiting::set_rate_limit(... GROUP_SWAP_KEYS, Exact(S))`) - **delta < span**: a swap is blocked while fewer than `S` blocks have passed, and it is allowed at exactly `S` blocks (**delta >= span**).

So to keep the same real wait time as old `N`, you must set `S = N + 1` (`N = 0` stays `0`).

The reason for this change is that `pallet-rate-limiting` uses one unified comparison rule (**delta < span**) for all limits. We keep that consistent and compensate for this specific legacy boundary in migration (`+1`).

## Checklist

## ToDo:
- [x] Add dev TS tests for each new group to confirm Rate Limiting doesn't have a regression
- [ ] Re-generate weights for all the pallets
- [x] Security audit
- [ ] Mainnet clone testing
  - [ ] Test all extrinsics rate limit with old runtime
  - [ ] Test all extrinsics rate limit with new runtime
  - [ ] Confirm no difference


ales-otf (commented):
these migrations were not removed, but moved to the runtime, because they depend on rate-limiting, but i didn't want to introduce a dependency on pallet-rate-limiting:

- migrate_network_lock_cost_2500
- migrate_network_lock_reduction_interval
- migrate_obsolete_rate_limiting_last_blocks_storage

ales-otf:
here are some additional details:

1. the pallet-rate-limiting code itself is about **600 LOC** (if we don't take the docs, storages and configs). it's all located in `pallets/rate-limiting/src/lib.rs`. 
   - what's really worth checking there additionaly to the `Pallet` impl is the tx-extension implementation (`pallets/rate-limiting/src/tx_extension.rs`) and especially `validate_*` functions. that part if stripped to just implementation (without marker-like traits and tests) is just about **150-200 LOC**

2. the changes in pallet-subtensor/pallet-admin-utils are deprecations of legacy rate-limiting with very few exclusions when time-stamp reads from pallet-rate-limiting (via interface) were needed and as far as i remember a single manual write was there

3. the migration + "resolvers" (read the description) in the runtime are the most crucial parts - they are the places where inconsistency with the legacy behavior might be introduced. the tests for the behavior is located in the `runtime/tests/rate_limiting.rs`. the migration is the biggest part, but I organized it in a script-like structure so that could be easy to review it. without tests, the utility code and the standalones migration (that will go to another PR) it's about **700 LOC**.

4. the rest are the precompiles + tests + contract tests. the precompiles have only changes related to the tx-extension enforcements and updates of the api

runtime/tests/common/mod.rs
```diff
@@ -0,0 +1,4 @@
+#![allow(dead_code)]
```
shamil-gadelshin: Why do we allow(dead_code)?

pallets/rate-limiting/runtime-api/src/lib.rs
```diff
@@ -0,0 +1,27 @@
...
+use subtensor_macros::freeze_struct;
+use subtensor_runtime_common::BlockNumber;
+
+#[cfg(feature = "std")]
+use serde::{Deserialize, Serialize};
+
+#[cfg_attr(feature = "std", derive(Serialize, Deserialize))]
+#[derive(Clone, Debug, Decode, Encode, Eq, PartialEq, TypeInfo)]
+#[freeze_struct("871e006597542bec")]
+pub struct RateLimitRpcResponse {
```
shamil-gadelshin: This structure requires extended docs.

pallets/rate-limiting/src/lib.rs
```diff
@@ -0,0 +1,1390 @@
...
+            clear_usage: bool,
+        ) -> DispatchResult {
+            T::AdminOrigin::ensure_origin(origin)?;
+
+            Self::ensure_call_registered(&transaction)?;
+            let target = Self::config_target(&transaction)?;
+            let tx_target = RateLimitTarget::Transaction(transaction);
+            let usage_target = Self::usage_target(&transaction)?;
+
+            match &scope {
```
shamil-gadelshin, line 1233: Seems like if-let-else.

pallets/rate-limiting/src/lib.rs
```diff
@@ -0,0 +1,1390 @@
...
+                if let Ok(resolved) = Self::usage_target(&identifier) {
+                    resolved
+                } else {
+                    return;
+                }
+            }
+            _ => raw_target,
+        };
+
+        match block {
```
shamil-gadelshin, line 1385: Seems like if-let-else

pallets/subtensor/src/staking/decrease_take.rs
```diff
@@ -52,10 +52,6 @@ impl<T: Config> Pallet<T> {
         // --- 4. Set the new take value.
```
shamil-gadelshin, line 52: Numeration is broken now.

pallets/subtensor/src/staking/increase_take.rs
```diff
@@ -52,19 +52,6 @@ impl<T: Config> Pallet<T> {
         let max_take = MaxDelegateTake::<T>::get();
         ensure!(take <= max_take, Error::<T>::DelegateTakeTooHigh);
```
shamil-gadelshin, line 54: Numeration is broken.

pallets/subtensor/src/swap/swap_hotkey.rs
```diff
@@ -46,31 +49,22 @@ impl<T: Config> Pallet<T> {
...
 
-        // 8. Swap LastTxBlock
-        let last_tx_block: u64 = Self::get_last_tx_block(old_hotkey);
-        Self::set_last_tx_block(new_hotkey, last_tx_block);
-        weight.saturating_accrue(T::DbWeight::get().reads_writes(1, 1));
-
-        // 9. Swap LastTxBlockDelegateTake
-        let last_tx_block_delegate_take: u64 = Self::get_last_tx_block_delegate_take(old_hotkey);
-        Self::set_last_tx_block_delegate_take(new_hotkey, last_tx_block_delegate_take);
+        // 8. Swap last-seen
```
shamil-gadelshin, line 92: Numeration is broken.

pallets/subtensor/src/tests/mechanism.rs
```diff
@@ -1388,101 +1384,6 @@ fn test_crv3_above_mechanism_count_fails() {
     });
 }
 
-#[test]
-fn test_do_commit_crv3_mechanism_weights_committing_too_fast() {
```
shamil-gadelshin, line 1392: Why did you remove this test?

gztensor (dismissed):

sam0x17 (dismissed):

pallets/subtensor/src/utils/rate_limiting.rs
```diff
@@ -80,7 +85,10 @@ impl TransactionType {
     /// Get the block number of the last transaction for a specific key, and transaction type
     pub fn last_block<T: Config>(&self, key: &T::AccountId) -> u64 {
         match self {
-            Self::RegisterNetwork => Pallet::<T>::get_network_last_lock_block(),
+            Self::RegisterNetwork => {
```
shamil-gadelshin, line 88: Don't we deprecate all of this?

pallets/subtensor/src/utils/rate_limiting.rs
```diff
@@ -89,13 +97,14 @@ impl TransactionType {
     /// type
     pub fn last_block_on_subnet<T: Config>(&self, hotkey: &T::AccountId, netuid: NetUid) -> u64 {
         match self {
-            Self::RegisterNetwork => Pallet::<T>::get_network_last_lock_block(),
+            Self::RegisterNetwork => {
```
shamil-gadelshin, line 100: Don't we deprecate all of this?

runtime/src/lib.rs
```diff
@@ -244,7 +251,7 @@ pub const VERSION: RuntimeVersion = RuntimeVersion {
     spec_version: 376,
     impl_version: 1,
     apis: RUNTIME_API_VERSIONS,
-    transaction_version: 1,
+    transaction_version: 2,
```
shamil-gadelshin: Why did you increase this parameter?

runtime/src/lib.rs
```diff
@@ -1576,16 +1609,38 @@ pub type TransactionExtensions = (
     ChargeTransactionPaymentWrapper<Runtime>,
     SudoTransactionExtension<Runtime>,
     pallet_subtensor::SubtensorTransactionExtension<Runtime>,
-    pallet_drand::drand_priority::DrandPriority<Runtime>,
-    frame_metadata_hash_extension::CheckMetadataHash<Runtime>,
+    (
+        pallet_drand::drand_priority::DrandPriority<Runtime>,
+        frame_metadata_hash_extension::CheckMetadataHash<Runtime>,
+    ),
+    rate_limiting::UnwrappedRateLimitTransactionExtension,
```
shamil-gadelshin, line 1749: Shouldn't we use the last extension as a nested one?

pallets/rate-limiting/src/tx_extension.rs
```diff
@@ -0,0 +1,889 @@
...
+        let bypass = <T as Config<I>>::LimitScopeResolver::should_bypass(origin, call);
+        let should_record =
+            bypass.record_usage && Pallet::<T, I>::should_record_usage(&identifier, &usage_target);
+
+        if bypass.bypass_enforcement {
+            return Ok(should_record.then_some((usage_target, usage, true)));
+        }
+
+        let usage_keys: BTreeSet<Option<<T as Config<I>>::UsageKey>> = match usage.clone() {
+            None => {
```
shamil-gadelshin, line 165: Seems like if-let-else

pallets/subtensor/src/staking/remove_stake.rs
```diff
@@ -235,6 +242,25 @@ impl<T: Config> Pallet<T> {
...
+                        hotkey: hotkey.clone(),
+                        netuid,
+                    };
+                    if let Some(last_seen) = T::RateLimiting::last_seen(
+                        rate_limiting::GROUP_STAKING_OPS,
+                        Some(usage_key),
+                    ) {
+                        let current = <frame_system::Pallet<T>>::block_number();
+                        if current.saturating_sub(last_seen) < span {
+                            continue;
```
shamil-gadelshin, line 253: Why do we continue instead of returning an error?

pallets/subtensor/src/staking/remove_stake.rs
```diff
@@ -235,6 +242,25 @@ impl<T: Config> Pallet<T> {
             }
             // If not Root network.
             if !netuid.is_root() {
+                // Manually filter out rate-limited subnets.
```
shamil-gadelshin, line 238: Why do we add a new feature here?

shamil-gadelshin (changes requested):
A couple of comments:

- IMHO, we should break large PRs into meaningful pieces in the future. 
- It makes sense to test it on a mainnet clone and verify that all limit classes are migrated successfully. 
- I don't see the latest "add_stake_burn" (subnet_buyback) limit as part of the migration here. 
- probably, we should use if-let-else instead of match-some-none.
- There are multiple broken comment numerations after the refactoring

ales-otf:
> A couple of comments:
> 
> * IMHO, we should break large PRs into meaningful pieces in the future.
> * It makes sense to test it on a mainnet clone and verify that all limit classes are migrated successfully.
> * I don't see the latest "add_stake_burn" (subnet_buyback) limit as part of the migration here.
> * probably, we should use if-let-else instead of match-some-none.
> * There are multiple broken comment numerations after the refactoring

Thank you for review! Replying to this one, will go next comments one by one afterwards.

- yes, agree. the problem with this one is caused by the intention to reduce the number of migrations, but generally - yes. no more than 5000 LOC anymore (i'll try less), until some crazy Cargo.lock rebuild
- i've covered it with the integration tests, but i'll check, thanks
- it's a standalone call - this PR covers only grouped calls. you introduced it, when i'd already finished writing migrations and had considered to split it to the standalone/grouped parts. and since standalone calls go to the following PR, i didn't add the migration for that call yet
- i checked the places you mentioned and i disagree with that, but it's mostly a style preference. i prefer the match in those cases, because on my opinion it's easier to read and follow. we probably need a style guide to prevent such collisions and to also reduce git conflicts, but since we don't have it yet, i'll stick to the match in this pr
- yeah, well i thought it's ok for the comments, but i'll fix them

pallets/rate-limiting/runtime-api/src/lib.rs
ales-otf: oh, it's good you pointed to this, because it needs an update

pallets/subtensor/src/tests/mechanism.rs
ales-otf: it's a rate-limiting test and covered by integration tests (`commit_crv3_mechanism_weights_are_rate_limited_per_mechanism`).

pallets/subtensor/src/utils/rate_limiting.rs
ales-otf: not yet, because not all rate limits are migrated

pallets/subtensor/src/utils/rate_limiting.rs
ales-otf: same here

runtime/src/lib.rs
ales-otf: well, i was sure that it's needed when we introduce a new transaction-extension. but i'll double check it

runtime/src/lib.rs
ales-otf: sorry, what do you mean?

pallets/subtensor/src/staking/remove_stake.rs
ales-otf: because previously it was [a part of `vlidate_remove_stake`](https://github.com/opentensor/subtensor/blob/b2e2cdebaf39c1badfc1af1cd0b5de5d0ddebb4c/pallets/subtensor/src/staking/stake_utils.rs#L1036) and for the previous question `validate_remove_stake` is [`continue` if `.is_err`](https://github.com/opentensor/subtensor/blob/b2e2cdebaf39c1badfc1af1cd0b5de5d0ddebb4c/pallets/subtensor/src/staking/remove_stake.rs#L253)

pallets/subtensor/src/staking/remove_stake.rs
ales-otf: replied below

pallets/subtensor/src/staking/remove_stake.rs
ales-otf: this exclusion needed a manual check because it's an iteration over netuids

runtime/tests/common/mod.rs
ales-otf: https://github.com/opentensor/subtensor/pull/2150/changes/0563aca9e58e66e8d8fa61aeb318864421ab5cbf

pallets/rate-limiting/runtime-api/src/lib.rs
ales-otf: https://github.com/opentensor/subtensor/pull/2150/changes/2186ab9055919aa3d1b1fcb188579fd640b34103

pallets/subtensor/src/tests/mechanism.rs
ales-otf: https://github.com/opentensor/subtensor/blob/2186ab9055919aa3d1b1fcb188579fd640b34103/runtime/tests/rate_limiting.rs#L621

pallets/rate-limiting/src/lib.rs
ales-otf: oh, you mean this. i disagree. match is natural and rusty here, makes it "labeled" - easy to read and follow. if-let-else would make it more if-else c-style.

pallets/rate-limiting/src/lib.rs
ales-otf: i disagree again - match is much better.

pallets/rate-limiting/src/tx_extension.rs
ales-otf: well, yeah. the same as with previous cases.

runtime/src/lib.rs
shamil-gadelshin: Why did you put drand and metadata extensions as a nested tuple instead of metadata and your extension? When we add a new one, we'll just add it to the last tuple.

ales-otf: ah. there was no reason for that. intuitively i felt like i made a change, so my change didn't belong to the older extensions. and the reason why i introduced a tuple is the limitation of the extensions number in this type. i'll change it like you suggested.

pallets/subtensor/src/swap/swap_hotkey.rs
ales-otf: [405465b](https://github.com/opentensor/subtensor/pull/2150/commits/405465b5fae8f0d96b84e0be0f1ef23c499a3c1c)

pallets/subtensor/src/staking/increase_take.rs
ales-otf: [405465b](https://github.com/opentensor/subtensor/pull/2150/commits/405465b5fae8f0d96b84e0be0f1ef23c499a3c1c)

pallets/subtensor/src/staking/decrease_take.rs
ales-otf: [405465b](https://github.com/opentensor/subtensor/pull/2150/commits/405465b5fae8f0d96b84e0be0f1ef23c499a3c1c)

runtime/src/lib.rs
ales-otf: [3a0438f](https://github.com/opentensor/subtensor/pull/2150/commits/3a0438fd1c19bc81ca995f104ce1e54f0b9d3fa7)

runtime/src/lib.rs
shamil-gadelshin: I don't think we need to change this parameter. AFAIU, it's reserved for the tx format change.

runtime/src/lib.rs
ales-otf: oh i need more changes for that, this commit changed only a single place. but a realized it when i started to resolve git conflicts 😄 but you can still check this commit—the rest is "compiler-driven" anyway

runtime/src/lib.rs
ales-otf: per the docs it looks like it's mostly needed in cases, when we want to change a pallet/call index or some existing call's api.

it also [mentions](https://paritytech.github.io/polkadot-sdk/master/sc_executor/struct.RuntimeVersion.html) transaction semantics changes. that could be the case, because new tx-extension can decline transactions that doesn't pass the rate-limiter, but semantically the call's behavior didn't change - it was rate-limited before. so from this perspective - yes, we should not change this number. i'll revert it

ales-otf: codex is stubborn though about this 😄 

> You’re right to question it. With the doc you quoted, the key is this:
> 
>   - If existing calls are truly fully compatible at user level, transaction_version does not need to change.
>   - tx-extension change alone is not automatically a bump trigger.
> 
>   For your migration specifically, the deciding point is:
> 
>   - Legacy checks were inside dispatch path.
>   - New checks are in tx-extension validation path.
> 
>   Even if final business outcome is “rate-limited call fails” in both versions, transaction semantics are not identical:
> 
>   - before: tx could be included and fail at dispatch,
>   - now: tx can be rejected as invalid before inclusion.
> 
>   That is a user-visible semantic difference in transaction behavior, so bumping transaction_version is defensible.
> 
>   If you can prove no such observable change (same acceptance, inclusion, fee/result behavior), then you can skip the bump. But with dispatch-time -> tx-validation-time enforcement move, it usually counts
>   as changed semantics.

ales-otf: [6ac10a2](https://github.com/opentensor/subtensor/pull/2150/commits/6ac10a2b9b6a37ccf12f44304451ab7a0320cec3)

node/src/mev_shield/author.rs
```diff
@@ -390,8 +390,12 @@ where
         >::from(0u64)),
         node_subtensor_runtime::sudo_wrapper::SudoTransactionExtension::<runtime::Runtime>::new(),
         pallet_subtensor::SubtensorTransactionExtension::<runtime::Runtime>::new(),
+        // Keep the same order while staying under the 12-item tuple limit.
         pallet_drand::drand_priority::DrandPriority::<runtime::Runtime>::new(),
-        frame_metadata_hash_extension::CheckMetadataHash::<runtime::Runtime>::new(false),
+        (
+            frame_metadata_hash_extension::CheckMetadataHash::<runtime::Runtime>::new(false),
+            node_subtensor_runtime::rate_limiting::UnwrappedRateLimitTransactionExtension::new(),
+        ),
```
l0r1s: This requires a node upgrade but should be fixed when the MEV shield is upgraded because we won't have extension node side anymore

pallets/rate-limiting/rpc/src/lib.rs
```diff
@@ -0,0 +1,82 @@
...
+        extrinsic: Vec<u8>,
+        at: Option<<Block as BlockT>::Hash>,
+    ) -> RpcResult<Option<RateLimitRpcResponse>> {
+        let api = self.client.runtime_api();
+        let at = at.unwrap_or_else(|| self.client.info().best_hash);
+
+        api.get_rate_limit(at, pallet, extrinsic)
+            .map_err(|e| Error::RuntimeError(format!("Unable to fetch rate limit: {e:?}")).into())
+    }
+}
```
l0r1s, line 82: In the documentation they seems to indicate that implementing custom RPC is somewhat deprecated and they recommend combining a runtime API with `state_call` which makes it much easier for dApps and doesn't require any node upgrade when update/changes are being made. 

That could be something to look into: https://paritytech.github.io/polkadot-sdk/master/polkadot_sdk_docs/reference_docs/custom_runtime_api_rpc/index.html

Wdyt?

ales-otf: I'll take a look. Thanks!

ales-otf: BTW it looks like I forgot to add the rpc to the node!

ales-otf: but as I understand it the point is to stop adding the RPC to the node. so I can just remove it and keep only runtime api? then the clients will use `state_call` and pass the runtime api method instead, right?

pallets/rate-limiting/src/lib.rs
```diff
@@ -0,0 +1,1390 @@
...
+            }
+        }
+    }
+
+    #[pallet::pallet]
+    #[pallet::without_storage_info]
+    pub struct Pallet<T, I = ()>(PhantomData<(T, I)>);
+
+    #[deny(clippy::expect_used)]
+    impl<T: Config<I>, I: 'static> Pallet<T, I> {
```
l0r1s, line 600: This is a bit confusing to me that we have the struct Pallet here and the impl Pallet here before calls.

The well adopted pattern across susbtrate pallets (including all of parity pallets) is to put the impl Pallet out of the mod pallet that way we can easily skim through the pallet storage/calls without looking at the full implementation.

It may be only personal taste though, just got surprised to see this first when scrolling!

pallets/rate-limiting/src/lib.rs
```diff
@@ -0,0 +1,1390 @@
...
+        /// `usage_key`: `None` checks global usage, `Some(key)` checks keyed usage.
+        ///
+        /// Returns [`Error::MissingScope`] when scoped config exists but no scope is provided.
+        /// Returns `Ok(true)` on bypass; otherwise every enforced scope must pass.
+        pub fn is_within_limit(
+            origin: &DispatchOriginOf<<T as Config<I>>::RuntimeCall>,
+            call: &<T as Config<I>>::RuntimeCall,
+            identifier: &TransactionIdentifier,
+            scopes: &Option<BTreeSet<<T as Config<I>>::LimitScope>>,
+            usage_key: &Option<<T as Config<I>>::UsageKey>,
```
l0r1s, line 613: Some helpers type could probably be extracted to make this more readable, it seems to be reused multiple times

pallets/rate-limiting/rpc/src/lib.rs
l0r1s: @ales-otf yes, that is what I understand from the documentation. Probably something we can discuss and document somewhere to make it known for the ecosystem and enforce this pattern in the future?

pallets/rate-limiting/runtime-api/src/lib.rs
```diff
@@ -0,0 +1,44 @@
...
+    /// Call is assigned to a group.
+    Grouped {
+        group_id: GroupId,
+        limit: RateLimitConfigRpcResponse,
+    },
+}
+
+sp_api::decl_runtime_apis! {
+    pub trait RateLimitingRuntimeApi {
+        fn get_rate_limit(pallet: Vec<u8>, extrinsic: Vec<u8>) -> Option<RateLimitRpcResponse>;
```
l0r1s, line 42: May need a comment for this function to understand what exactly the params are because this is something that will be public facing if we go the route of using state_call with Runtime APIs

pallets/rate-limiting/src/lib.rs
```diff
@@ -0,0 +1,1390 @@
...
+
+            for (target, rule) in &self.limit_setting_rules {
+                LimitSettingRules::<T, I>::insert(target, rule.clone());
+            }
+        }
+    }
+
+    #[pallet::pallet]
+    #[pallet::without_storage_info]
+    pub struct Pallet<T, I = ()>(PhantomData<(T, I)>);
```
l0r1s, line 597: Why do you think we should have multiple instances of the pallet? Isn't it more cumbersome than just having a single instance that manage all rate limits across all pallets? My impression is that it can become problematic if we need to have 1:1 instance for each pallet.

If I take the collective pallet as an example, it has instances because group A and group B are fundamentally different given they can't share the same on-chain storage but in the case of rate limits, this distinction is already handled by the concept of group if I understand correctly.

I don't see much benefits of having instances but I may be missing something specific?

runtime/src/rate_limiting/mod.rs
```diff
@@ -0,0 +1,604 @@
...
+        Self(pallet_rate_limiting::RateLimitTransactionExtension::<Runtime>::new())
+    }
+}
+
+impl UnwrappedRateLimitTransactionExtension {
+    pub fn new() -> Self {
+        Self::default()
+    }
+
+    fn unwrap_nested_calls(call: &RuntimeCall) -> Vec<&RuntimeCall> {
```
l0r1s, line 503: It is missing Crowdloan and Scheduler and it is kind of error prone if new type of calls are added and we forget to update this code.

It could leverage the DispatchGuard but I'm not sure how it could be done because validate needs to pass information to post_dispatch to write the updated limits, right?

runtime/src/rate_limiting/mod.rs
```diff
@@ -0,0 +1,604 @@
...
+        >>::Val,
+    >;
+    type Pre = Vec<
+        <pallet_rate_limiting::RateLimitTransactionExtension<Runtime> as TransactionExtension<
+            RuntimeCall,
+        >>::Pre,
+    >;
+
+    fn weight(&self, _call: &RuntimeCall) -> Weight {
+        Weight::zero()
```
l0r1s, line 563: Probably not zero if we do some storage reads? Could be benchmarked

pallets/rate-limiting/src/lib.rs
ales-otf: yes, the idea is to have an instance per pallet, where rate-limiting is needed. because otherwise you would need huge resolvers implementations that will have all the pallets inside. it makes the code more isolated. right now we can't isolate admin from subtensor-module though, because they are too coupled, but in the future (when the refactoring is done) i imagined having an instance per pallet. also the initial intention was to introduce macro so you could add attributes/decorators to calls and they would produce needed resolvers, and in that case multiple instances would be needed, but it doesn't look like a right direction anymore.

runtime/src/rate_limiting/mod.rs
ales-otf: yes, i have no idea how to work around scheduler. will take a look at crowdloan, thanks!

this is definitely not the solution, it's just a workaround. because the solution of this problem is out of scope of this task.

pallets/rate-limiting/src/lib.rs
```diff
@@ -0,0 +1,1390 @@
...
+    #[pallet::storage]
+    #[pallet::getter(fn group_members)]
+    pub type GroupMembers<T: Config<I>, I: 'static = ()> = StorageMap<
+        _,
+        Blake2_128Concat,
+        <T as Config<I>>::GroupId,
+        GroupMembersOf<T, I>,
+        ValueQuery,
+    >;
+
```
l0r1s, line 359: Wouldn't this be more efficient using a StorageDoubleMap?

pallets/rate-limiting/src/lib.rs
l0r1s: > because otherwise you would need huge resolvers implementations that will have all the pallets inside

Yes but this is preferable imho, what do you see this as an issue? It would be scoped to a single place, somewhere where you already have access to all pallets, runtime/ dir for example, this is already where the extension is defined.

Having 1 instance per pallet means you have N times the admin extrinsics per pallet, N times the transaction extensions (and fee that are paid for + metadata encoded in the extrinsic being sent) and it will create storage fragmentation for no real benefits.

If a new pallet want to add rate limiting, it would need to add another pallet, another extension, etc... while having a single central rate limiting pallet would allow 2 extrinsic call and that's it, you have rate limit for your pallet. Much easier to configure, much easier to reasonate about, complexity is isolated to a single place instead of spanning it everywhere. Group and TransactionIdentifier have really low chance of collusion across pallets and extrinsics, it would also allow for cross pallets shared configuration for simple limits.

If you look at pallets that are used as a cross cutting concern, we have the Preimage pallet. Not every pallet that needs storage define it's own storage pallet instance, they just use the interface provided by the the Preimage pallet and this is much easier to reason about.

It will also complexify use and confuse users when used through future governance.

pallets/rate-limiting/src/lib.rs
```diff
@@ -0,0 +1,1390 @@
...
+            GroupMembers::<T, I>::mutate(group, |members| members.remove(identifier))
+        }
+    }
+
+    #[pallet::call]
+    #[deny(clippy::expect_used)]
+    impl<T: Config<I>, I: 'static> Pallet<T, I> {
+        /// Registers a call for rate limiting and seeds its initial configuration.
+        #[pallet::call_index(0)]
+        #[pallet::weight(T::DbWeight::get().reads_writes(3, 3))]
```
l0r1s, line 902: The extrinsic weights only include reads/writes but are missing the fixed weight values, is it normal?

pallets/rate-limiting/src/lib.rs
```diff
@@ -0,0 +1,1390 @@
...
+            group: <T as Config<I>>::GroupId,
+            read_only: bool,
+        ) -> DispatchResult {
+            T::AdminOrigin::ensure_origin(origin)?;
+
+            Self::ensure_call_registered(&transaction)?;
+            Self::ensure_group_details(group)?;
+
+            let current = CallGroups::<T, I>::get(transaction);
+            ensure!(current.is_none(), Error::<T, I>::CallAlreadyInGroup);
```
l0r1s, line 1063: It's not possible to change the group of an existing call without first removing it?

Couldn't the assign_call_to_group and remove_call_to_group be unified to only have something like set_call_group?

pallets/rate-limiting/src/lib.rs
```diff
@@ -0,0 +1,1390 @@
...
+                        Self::deposit_event(Event::CallGroupUpdated {
+                            transaction,
+                            group: None,
+                        });
+                    }
+                }
+            }
+
+            if clear_usage {
+                let _ = LastSeen::<T, I>::clear_prefix(usage_target, u32::MAX, None);
```
l0r1s, line 1275: Is LastSeen bounded? Extrinsic weights indicate 4/4 but if this is not the case, it should be parameterized.

runtime/src/rate_limiting/mod.rs
l0r1s: > because the solution of this problem is out of scope of this task

Could you explain? My understanding is that you use the transaction extension validate to check limits and you use the post_dispatch to update usage? post_dispatch is always O(1)?

runtime/src/rate_limiting/mod.rs
```diff
@@ -0,0 +1,604 @@
...
+        info: &DispatchInfoOf<RuntimeCall>,
+        post_info: &mut PostDispatchInfo,
+        len: usize,
+        result: &DispatchResult,
+    ) -> Result<(), TransactionValidityError> {
+        for entry in pre {
+            pallet_rate_limiting::RateLimitTransactionExtension::<Runtime>::post_dispatch(
+                entry, info, post_info, len, result,
+            )?;
+        }
```
l0r1s, line 603: If there are some storage writes, we need to account for them in the extension weight fn

pallets/rate-limiting/src/lib.rs
```diff
@@ -0,0 +1,1390 @@
...
+        /// Default rule applied when a target does not have an explicit entry in
+        /// [`LimitSettingRules`].
+        type DefaultLimitSettingRule: Get<Self::LimitSettingRule>;
+
+        /// Origin checker invoked when setting a rate limit, parameterized by the stored rule.
+        type LimitSettingOrigin: EnsureLimitSettingRule<OriginFor<Self>, Self::LimitSettingRule, Self::LimitScope>;
+
+        /// Scope identifier used to namespace stored rate limits.
+        type LimitScope: Parameter + Clone + PartialEq + Eq + Ord + MaybeSerializeDeserialize;
+
```
l0r1s, line 239: This is kind of hardcoded even though it is behind a config type, no? Right now, it is defined as NetUid, but what will happen if we decide to have some calls scoped to NetUid and some other calls scoped to NetUid + MechId?

ales-otf: This is the same as with the usage key. It's now NetUid only, because we didn't have other use cases for that. if you want to introduce something like NetUid + MechId, you'd introduce an enum, something like:

```rust
enum Scope {
    NetUid(NetUid),
    NetUidMechId(NetUid, MechId),
}
```

pallets/rate-limiting/src/lib.rs
ales-otf: I just don't think that mixing concerns is a good practice. It's harder to debug, at very least, harder to navigate, to read. Separating the limiter instances doesn't mean to make it span across the codebase producing spaghetti - in opposite it means scoping/namespacing and organizing.

Generally I was thinking about restructuring how we config pallets into a module per config. Of course it's a hell of modules, but that would make the code much more organized and easier to navigate. In that case having a rate-limiting related code in a module with configs for a pallet would be very convenient.

> Having 1 instance per pallet means you have N times the admin extrinsics per pallet

I think it's easier, then navigate and manage all the groups and limit configs for the entire system. Like, you have `staking-rate-limit`, `weights-set-rate-limit`, etc. and you just see the instance and you know that all the groups there is just about this pallet: `staking-rate-limit::set_rate_limit`, `weights-set-rate-limit::set_rate_limit`, etc. And it simplifies the limit scope type and the usage key.

> N times the transaction extensions (and fee that are paid for + metadata encoded in the extrinsic being sent)
> ...
> while having a single central rate limiting pallet would allow 2 extrinsic call and that's it,

These are fair points. But from the algorithm complexity perspective it's the same. Only applied calls are handled anyway. And introducing a new pallet is not a thing that we do often. Extending it with rate-limiting is just an additional update. Yes, it's not that simple like with a single instance, but it's also not a big deal.

Anyway having this pallet instantiable doesn't mean we already have all these issues. We still have a single central instance. And instancing just provides more room on which way you want to organize your code. It's not a problem right now and how we will organize it in our codebase, we can consider later.

runtime/src/rate_limiting/mod.rs
ales-otf: I meant the solution of the nested calls problem. Having an unwraper tx-extension is a workaround - not a solution. Solution would make us forget about this problem and taking care about nested calls handling. I think the dispatch guard is the solution.

sam0x17:
just a few merge conflicts now

ales-otf:
> just a few merge conflicts now

yeah, i have them after almost every push to `devnet-ready`... anyway i still need to address Loris' comments and adopt a few things we introduced while this PR was in review

PR #2165, opened 2025-10-30
use origin caller in extension

open-junius:
## Description
We should use the original caller as origin to call extension like add_stake.
Currently, the extrinsic take the contract address as origin.

## Related Issue(s)

- Closes #[issue number]

## Type of Change

- [ ] Documentation update

## Breaking Change

## Checklist

## Additional Notes

### DIFF
--- a/chain-extensions/src/lib.rs
+++ b/chain-extensions/src/lib.rs
@@ -568,7 +568,7 @@ where
             .ext()
             .caller()
             .account_id()
-            .map(|id| id.clone())
+            .cloned()
             .map_err(|_| DispatchError::Other("Failed to get caller"))
     }
 }
--- a/chain-extensions/src/lib.rs
+++ b/chain-extensions/src/lib.rs
@@ -98,7 +98,7 @@ where
                     .map_err(|_| DispatchError::Other("Failed to decode input parameters"))?;
 
                 let call_result = pallet_subtensor::Pallet::<T>::add_stake(
-                    RawOrigin::Signed(env.caller()).into(),
+                    RawOrigin::Signed(env.caller()?).into(),
                     hotkey,
                     netuid,
                     amount_staked,
@@ -124,7 +124,7 @@ where
                     .map_err(|_| DispatchError::Other("Failed to decode input parameters"))?;
 
                 let call_result = pallet_subtensor::Pallet::<T>::remove_stake(
-                    RawOrigin::Signed(env.caller()).into(),
+                    RawOrigin::Signed(env.caller()?).into(),
                     hotkey,
                     netuid,
                     amount_unstaked,
@@ -150,7 +150,7 @@ where
                     .map_err(|_| DispatchError::Other("Failed to decode input parameters"))?;
 
                 let call_result = pallet_subtensor::Pallet::<T>::unstake_all(
-                    RawOrigin::Signed(env.caller()).into(),
+                    RawOrigin::Signed(env.caller()?).into(),
                     hotkey,
                 );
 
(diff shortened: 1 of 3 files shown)


camfairchild (approved):

silva-fj:
Hi @open-junius, I think this is not a good idea. Apart from the security concerns I shared in the private chat, this change would limit the utility of the contracts. Here is a concrete example: Imagine you are building a bridge for a subnet, users will transfer alpha to the contract, locking the funds on the bittensor side, a relayer submits the tx to mint the tokens on the subnet side. Then later the user wants to transfer back the tokens to bittensor, how can the contract transfer the stake to the user if the origin will be the caller (relayer)? I can think of other examples where this would be an issue.

bdmason:
@open-junius @camfairchild I don't think this is a good idea. In EVM and Substrate a contract is an entity in itself. It has it's own address, storage & balance. It's an autonomous entity. It needs to be able to act as itself.

The whole point of `pallet_contracts` is to bring EVM like functionality to Substrate, but without all the pain of dealing with dual wallets and the god awful address mapping. These contracts are not supposed to be stateless like on Solana.

masterdubs:
@open-junius 

Could you share more context on the motivation and intended outcomes of this change?

-  What specific issue does using the contract address as `Origin` create today (security, re-entrancy guards, UX)? Any incidents or reproducible edge cases?

- How should contracts that act on behalf of users (bridges, relayers, multisigs, DAOs) continue to function when the *contract itself* must be the actor of record? E.g. in a bridge flow where a relayer submits a burn/unlock on behalf of a user, how does the contract remain accountable for transfers/stake if `Origin` becomes the external caller (echoing @silva-fj)?

- Wouldn’t this break common DeFi patterns (Uniswap-style routers, aggregators, vaults) where:
  - a router/forwarder contract is expected to be the `Origin`,
  - allowances/escrow/state checks key off the contract as caller,
  - and downstream pallets rely on the contract-origin for accounting and guardrails?
  Any guidance on how those workflows should be rewritten if `Origin` is now the EOA?

camfairchild:
> Hi @open-junius, I think this is not a good idea. Apart from the security concerns I shared in the private chat, this change would limit the utility of the contracts. Here is a concrete example: Imagine you are building a bridge for a subnet, users will transfer alpha to the contract, locking the funds on the bittensor side, a relayer submits the tx to mint the tokens on the subnet side. Then later the user wants to transfer back the tokens to bittensor, how can the contract transfer the stake to the user if the origin will be the caller (relayer)? I can think of other examples where this would be an issue. 


As shared in the group chat, 

The question to you was if you could confirm that the origin here is the contract calling the chain extension or if it's something else.

https://docs.rs/pallet-contracts/latest/src/pallet_contracts/exec.rs.html#259

This doc suggests this is the case, but @open-junius 's concern was that this wasn't the case.

bdmason:
> > Hi @open-junius, I think this is not a good idea. Apart from the security concerns I shared in the private chat, this change would limit the utility of the contracts. Here is a concrete example: Imagine you are building a bridge for a subnet, users will transfer alpha to the contract, locking the funds on the bittensor side, a relayer submits the tx to mint the tokens on the subnet side. Then later the user wants to transfer back the tokens to bittensor, how can the contract transfer the stake to the user if the origin will be the caller (relayer)? I can think of other examples where this would be an issue.
> 
> As shared in the group chat,
> 
> The question to you was if you could confirm that the origin here is the contract calling the chain extension or if it's something else.
> 
> https://docs.rs/pallet-contracts/latest/src/pallet_contracts/exec.rs.html#259
> 
> This doc suggests this is the case, but @open-junius 's concern was that this wasn't the case.

Yes the contract is calling the chain extension.

camfairchild:
> > the origin here is the contract calling the chain extension or if it's something else.
> > 
> > https://docs.rs/pallet-contracts/latest/src/pallet_contracts/exec.rs.html#259
> > 
> 
> Yes the contract is calling the chain extension.

I'll take this to mean "the Origin is the contract calling the chain extension."

PR #2240, opened 2025-12-01
Voting Power EMA

konrad0960:
## Description
Voting power feature 
Purpose: Track validator voting power using an Exponential Moving Average (EMA) of their stake over time, preventing sudden stake changes from immediately affecting governance weight.

  Key Components:
  - EMA-based voting power calculation (updated each epoch)
  - Per-subnet enable/disable with 14-day grace period
  - Configurable alpha parameter for EMA smoothness
  - Voting power transfers on hotkey swap
  - EVM precompile for smart contract access

  Extrinsics: enable_voting_power_tracking, disable_voting_power_tracking, sudo_set_voting_power_ema_alpha

  Storage: VotingPower, VotingPowerTrackingEnabled, VotingPowerDisableAtBlock, VotingPowerEmaAlpha

## Type of Change

- [ ] Documentation update

## Checklist

## Additional Notes

I’ve attempted to run the VotingPower precompile EVM tests locally but encountered multiple Node.js compatibility issues with the polkadot-api dependencies.
Being unable to resolve them I want to give CI a try
That being said the unit tests written cover the core logic.


open-junius:
Enabled the CI process, to check if EVM test is runnable for PR.

konrad0960:
> Enabled the CI process, to check if EVM test is runnable for PR.

can we try again?

pallets/subtensor/src/epoch/run_epoch.rs
```diff
@@ -147,6 +147,9 @@ impl<T: Config> Pallet<T> {
         ValidatorTrust::<T>::insert(netuid, validator_trust);
         ValidatorPermit::<T>::insert(netuid, new_validator_permit);
         StakeWeight::<T>::insert(netuid, stake_weight);
+
+        // Update voting power EMA for all validators on this subnet
+        Self::update_voting_power_for_subnet(netuid);
```
gztensor: The 'persist_netuid_epoch_terms' method, as the name suggests, should only persist what's already calculated (subnet epoch terms). All calculations should happen in `epoch_mechanism` function. Because `update_voting_power_for_subnet` is using persisted terms from state, I suggest that it should be refactored to use the calculated terms in memory instead of re-reading them from state.

pallets/subtensor/src/utils/voting_power.rs
```diff
@@ -0,0 +1,286 @@
...
+        let previous_ema = VotingPower::<T>::get(netuid, hotkey);
+
+        // Calculate new EMA value
+        // new_ema = alpha * current_stake + (1 - alpha) * previous_ema
+        // All values use 18 decimal precision for alpha (alpha is in range [0, 10^18])
+        let new_ema = Self::calculate_voting_power_ema(current_stake, previous_ema, alpha);
+
+        // Only remove if they previously had voting power ABOVE threshold and it decayed below.
+        // This allows new validators to build up voting power from 0 without being removed.
+        if new_ema < min_stake && previous_ema >= min_stake {
```
gztensor, line 199: If the new validator has just built the previous_ema above min_stake and new_ema dropped below min_stake because of noise, the ema will be reset to 0. Is that desired behavior?

pallets/subtensor/src/epoch/run_epoch.rs
gztensor: Also, is that desired behavior to call this once per epoch? If a validator keeps moving stake around subnets to hit their epoch block, they may accumulate voting power in many subnets with one stake.

gztensor (changes requested):

pallets/subtensor/src/utils/voting_power.rs
konrad0960: Didn’t thought of this edge case
I guess we can then decide to reset to 0 when it gets below 90% of required threshold?

pallets/subtensor/src/epoch/run_epoch.rs
```diff
@@ -138,6 +139,9 @@ impl<T: Config> Pallet<T> {
         ValidatorTrust::<T>::insert(netuid, validator_trust);
         ValidatorPermit::<T>::insert(netuid, new_validator_permit);
         StakeWeight::<T>::insert(netuid, stake_weight);
+
+        // Update voting power EMA for all validators on this subnet
+        Self::update_voting_power_for_subnet(netuid, output);
```
gztensor: Great update! Please one little change here: Call this function before (or after) calling `persist_netuid_epoch_terms` in `epoch_with_mechanisms` to avoid confusion that `persist_netuid_epoch_terms` also updates voting power.

pallets/subtensor/src/utils/voting_power.rs
```diff
@@ -0,0 +1,289 @@
...
+        // Get previous EMA value
+        let previous_ema = VotingPower::<T>::get(netuid, hotkey);
+
+        // Calculate new EMA value
+        // new_ema = alpha * current_stake + (1 - alpha) * previous_ema
+        // All values use 18 decimal precision for alpha (alpha is in range [0, 10^18])
+        let new_ema = Self::calculate_voting_power_ema(current_stake, previous_ema, alpha);
+
+        // Use 90% of min_stake as removal threshold (hysteresis to prevent noise-triggered removal)
+        let removal_threshold = min_stake.saturating_mul(9) / 10;
```
gztensor: Please use `safe_div(10u64)` or CI will fail

pallets/subtensor/src/utils/voting_power.rs
```diff
@@ -0,0 +1,289 @@
...
+
+        // Only remove if they previously had voting power ABOVE threshold and decayed significantly below.
+        // This allows new validators to build up voting power from 0 without being removed.
+        if new_ema < removal_threshold && previous_ema >= min_stake {
```
gztensor: If alpha is slow enough, we'll probably never hit this condition.

pallets/subtensor/src/utils/voting_power.rs
```diff
@@ -0,0 +1,289 @@
...
+        // Get the EMA alpha value for this subnet
+        let alpha = Self::get_voting_power_ema_alpha(netuid);
+
+        // Get minimum stake threshold for validator permit
+        let min_stake = Self::get_stake_threshold();
+
+        // Iterate over epoch output using pre-calculated values
+        for (hotkey, terms) in epoch_output.iter() {
+            // Only validators (with vpermit) get voting power, not miners
+            if terms.new_validator_permit {
```
gztensor, line 143: What happens if a neuron gets deregistered? We don't have it in epoch terms, but we've got to garbage-collect their voting power at some point, probably not immediately.

gztensor (changes requested):

pallets/subtensor/src/utils/voting_power.rs
konrad0960: So far the requirements didn't specify how fast removal should happen in case stake is drastically dropped.
Hard rules on having voting power removed immediately if stake drops below threshold could be easily gamed by dropping stake just near the threshold.
Probably assymetric EMA could solve this? Introducing alpha_up and alpha_down whereas alpha_down should default to faster decay in example 0.5?

pallets/subtensor/src/utils/voting_power.rs
konrad0960: we got clear_voting_power_for_deregistered_hotkeys() just at the end of this function

pallets/subtensor/src/utils/voting_power.rs
shamil-gadelshin: Is it possible to trigger voting power removal by stake manipulation? 

Also, @ppolewicz suggests simplifying the logic of this method.

shamil-gadelshin (commented):
Do we have tests for voting power calculation methods?

precompiles/src/voting_power.rs
```diff
@@ -0,0 +1,131 @@
...
+/// This precompile allows smart contracts to query voting power for validators,
+/// enabling on-chain governance decisions like slashing and spending.
+pub struct VotingPowerPrecompile<R>(PhantomData<R>);
+
+impl<R> PrecompileExt<R::AccountId> for VotingPowerPrecompile<R>
+where
+    R: frame_system::Config + pallet_subtensor::Config,
+    R::AccountId: From<[u8; 32]> + ByteArray,
+{
+    const INDEX: u64 = 2060;
```
open-junius: The address is used by address_mapping precompile, the PR #2337 is merged. need to fix the conflict.

open-junius (changes requested):

precompiles/src/voting_power.rs
ppolewicz: Thank you, I just fixed that!

pallets/subtensor/src/utils/voting_power.rs
ppolewicz: simplified!

pallets/subtensor/src/utils/voting_power.rs
ppolewicz: I removed it, it was overcomplicated

pallets/subtensor/src/utils/voting_power.rs
ppolewicz: this branch was simplified

pallets/subtensor/src/epoch/run_epoch.rs
```diff
@@ -22,6 +22,7 @@ pub struct EpochTerms {
     pub validator_trust: u16,
     pub new_validator_permit: bool,
     pub bond: Vec<(u16, u16)>,
+    pub stake: u64,
```
gztensor, line 25: We should use type-safe method of expressing stake (alpha): AlphaCurrency.

gztensor (changes requested):

sam0x17 (approved):

sam0x17:
Accidentally merged and then reverted, @gztensor has some changes

PR #2247, opened 2025-12-03
Update max fee rate

shamil-gadelshin:
This PR sets the max fee rate constant from the swap pallet from 10000 (15.26%) to 1310 (2%). 

Additionally, it introduces a migration infrastructure for the swap pallet with `HasMigrationRun` storage, `on_runtime_upgrade` hook, and a related test.


shamil-gadelshin:
a version with removed migration: https://github.com/opentensor/subtensor/pull/2248

PR #2251, opened 2025-12-04
Rework coldkey swap

### LINKED ISSUE #2230: Rework cold key swap from scheduled to announce/swap
The idea is to rework the coldkey swap working as follow:
- you first announce your swap (with hash, as to not expose the destination)
- when the minimum delay has elapsed since announcement, you can call swap. 

This is similar to the `announce`/`proxy_announced`.

lorismoulin:
Closes #2230 

## Summary
This PR refactors the coldkey swap mechanism from a scheduler-based system to an announcement-based system with a configurable delay period. This change improves security, simplifies the codebase, and provides better transparency for coldkey swap operations.

### 1. **New Announcement-Based Flow**
- **Added `announce_coldkey_swap`**: Users must first announce their intention to swap coldkeys by providing a hash of the new coldkey
- **Added `swap_coldkey_announced`**: After the announcement delay period, users can execute the swap by providing the actual new coldkey
- **Announcement validation**: The system verifies that the new coldkey hash matches the announced hash and that sufficient time has passed
- **Sliding delay**: Each new announcement resets the delay period to 5 days from the announcement time, allowing users to change their plans without restriction

### 2. **Configuration Changes**
- **Renamed**: `InitialColdkeySwapScheduleDuration` → `InitialColdkeySwapAnnouncementDelay`
- **Removed**: `InitialColdkeySwapRescheduleDuration` (no longer needed)
- **Storage**: `ColdkeySwapScheduled` map → `ColdkeySwapAnnouncements` map
- **Default delay**: Maintained at 5 days (36,000 blocks)

### 3. **Refactored Core Swap Logic**
- **Simplified `do_swap_coldkey`**: Consolidated swap logic with clearer separation of concerns
- **New helper functions**: 
  - `transfer_subnet_ownership`
  - `transfer_auto_stake_destination`
  - `transfer_coldkey_stake`
  - `transfer_staking_hotkeys`
  - `transfer_hotkeys_ownership`
- **Identity preservation**: If new coldkey already has an identity, it's preserved instead of overwritten

### 4. **Updated `swap_coldkey` Extrinsic**
- Now root-only for arbitrary coldkey swaps (doesn't require announcement)
- Automatically removes any existing announcement when called
- Returns `DispatchResult` instead of `DispatchResultWithPostInfo`

### 5. **Admin Utilities**
- **Deprecated**: `sudo_set_coldkey_swap_schedule_duration` (call index 54)
- **Added**: `sudo_set_coldkey_swap_announcement_delay` (call index 84)
- **Added**: `remove_coldkey_swap_announcement` (call index 127) - root-only removal of announcements

### 6. **Events**
- **Removed**: `ColdkeySwapScheduled`
- **Added**: 
  - `ColdkeySwapAnnounced` - emitted when an announcement is made
  - `ColdkeySwapAnnouncementRemoved` - emitted when an announcement is removed
- **Updated**: `ColdkeySwapAnnouncementDelaySet` (renamed from `ColdkeySwapScheduleDurationSet`)

### 7. **Error Handling**
- **Removed**: `ColdkeyIsInArbitration`, `SwapAlreadyScheduled`, `FailedToSchedule`, `ColdkeySwapReannouncedTooEarly`
- **Added**:
  - `ColdkeySwapAnnouncementNotFound`
  - `ColdkeySwapTooEarly`
  - `AnnouncedColdkeyHashDoesNotMatch`
  - `Deprecated` (for the old `schedule_swap_coldkey` call)

### 8. **Transaction Extension**
- Updated `CustomTransactionError::ColdkeyInSwapSchedule` → `ColdkeySwapAnnounced`

### 9. **Migration**
- **Added**: `migrate_coldkey_swap_scheduled_to_announcements` to migrate existing scheduled swaps to the new announcement system
- Only migrates future scheduled swaps (past ones are ignored)
- Adjusts announcement time to be `scheduled_time - delay` so swaps can execute at the originally scheduled time

## Benefits

1. **Improved Security**: Announcement hash prevents front-running and ensures the exact coldkey is swapped
2. **Transparency**: All announcements are publicly visible on-chain before execution
3. **Simplified Code**: Removes complex scheduler integration and reduces storage requirements
4. **Flexibility**: Users can reannounce at any time with a fresh 5-day delay period, allowing them to change their coldkey swap plans without restriction
5. **Root Override**: Admins can still perform emergency swaps via the root-only `swap_coldkey` call

## Breaking Changes

⚠️ **BREAKING CHANGES**:
- The `schedule_swap_coldkey` extrinsic is deprecated and will return an error
- Configuration parameters renamed (migration handled automatically)
- Storage layout changed (migration included)

## Migration Path

For users with existing scheduled coldkey swaps:
- The migration will automatically convert scheduled swaps to announcements
- Your swap can still be executed at the originally scheduled time using `swap_coldkey_announced`
- You can reannounce at any time if you need to change your new coldkey, which will reset the delay to 5 days from the new announcement
- Ensure you have the new coldkey available when the delay period expires


pallets/subtensor/src/macros/dispatches.rs
```diff
@@ -2431,5 +2343,91 @@ mod dispatches {
...
+        )]
+        pub fn announce_coldkey_swap(
+            origin: OriginFor<T>,
+            new_coldkey_hash: T::Hash,
+        ) -> DispatchResult {
+            let who = ensure_signed(origin)?;
+            let now = <frame_system::Pallet<T>>::block_number();
+
+            if let Some(existing) = ColdkeySwapAnnouncements::<T>::get(who.clone()) {
+                let delay = ColdkeySwapAnnouncementDelay::<T>::get();
```
camfairchild: We had an extra delay of a day from this to reschedule.

Gives time to complain about something

pallets/subtensor/src/macros/dispatches.rs
```diff
@@ -2431,5 +2343,91 @@ mod dispatches {
...
+                let when = existing.0;
+                ensure!(
+                    now > when.saturating_add(delay),
+                    Error::<T>::ColdkeySwapReannouncedTooEarly
+                );
+            }
+
+            ColdkeySwapAnnouncements::<T>::insert(who.clone(), (now, new_coldkey_hash.clone()));
```
camfairchild: Should insert the block to swap at, incase we change the duration midway. They'd keep the original length.

Can possibly store the block *and* the length of the start and the end.

pallets/subtensor/src/macros/dispatches.rs
```diff
@@ -2431,5 +2343,91 @@ mod dispatches {
...
+
+            Self::deposit_event(Event::ColdkeySwapAnnounced {
+                who,
+                new_coldkey_hash,
+            });
+            Ok(())
+        }
+
+        /// Performs a coldkey swap iff an announcement has been made.
+        /// The provided new coldkey must match the announced coldkey hash.
```
camfairchild: Can we document which hash it is here

pallets/subtensor/src/macros/dispatches.rs
```diff
@@ -2431,5 +2343,91 @@ mod dispatches {
 
             Ok(())
         }
+
+        /// Announces a coldkey swap using coldkey hash.
+        /// This is required before the coldkey swap can be performed after the delay period.
```
camfairchild: Also document hash type here

pallets/subtensor/src/macros/dispatches.rs
```diff
@@ -2431,5 +2343,91 @@ mod dispatches {
...
+                let when = existing.0;
+                ensure!(
+                    now > when.saturating_add(delay),
+                    Error::<T>::ColdkeySwapReannouncedTooEarly
+                );
+            }
+
+            ColdkeySwapAnnouncements::<T>::insert(who.clone(), (now, new_coldkey_hash.clone()));
+
+            Self::deposit_event(Event::ColdkeySwapAnnounced {
```
camfairchild, line 2394: We should either deduct the swap cost here. Only the first announcement pays it, but we should deduct it now.

Then no need to keep it for later either

camfairchild (changes requested):

pallets/subtensor/src/macros/dispatches.rs
```diff
@@ -2432,5 +2348,109 @@ mod dispatches {
...
             Ok(())
         }
+
+        /// Announces a coldkey swap using BlakeTwo256 hash of the new coldkey.
+        ///
+        /// This is required before the coldkey swap can be performed
+        /// after the delay period.
+        ///
+        /// It can be reannounced after a delay of `ColdkeySwapReannouncementDelay` between the
+        /// original announcement and the reannouncement.
```
camfairchild: this would be since the first valid execution block, not the announcement

lorismoulin: @camfairchild right, I updated the comment

pallets/subtensor/src/macros/errors.rs
```diff
@@ -150,10 +148,14 @@ mod errors {
         TooManyChildren,
         /// Default transaction rate limit exceeded.
         TxRateLimitExceeded,
-        /// Swap already scheduled.
```
shamil-gadelshin, line 153: Shouldn't we deprecate enum items instead of deleting them?

pallets/subtensor/src/macros/events.rs
```diff
@@ -190,17 +200,6 @@ mod events {
                 <T as frame_system::Config>::AccountId,
             >>::Balance,
         },
-        /// A coldkey swap has been scheduled
-        ColdkeySwapScheduled {
```
shamil-gadelshin, line 196: Shouldn't we deprecate enum items instead of deleting them?

pallets/subtensor/src/lib.rs
```diff
@@ -2433,7 +2419,7 @@ pub mod pallet {
 
 #[derive(Debug, PartialEq)]
 pub enum CustomTransactionError {
-    ColdkeyInSwapSchedule,
+    ColdkeySwapAnnounced,
```
shamil-gadelshin: Shouldn't we deprecate enums instead of deleting them?

shamil-gadelshin (changes requested):
Looks good! I have a general question about enums and public API preservation.

pallets/subtensor/src/macros/errors.rs
camfairchild: imo this is fine because we should expect clients to use the metadata from the chain to handle decoding, etc.

camfairchild (approved):

pallets/subtensor/src/macros/errors.rs
lorismoulin: @shamil-gadelshin I was under the impression the enum are safe to add/remove elements when they are not used in long term storage:
- errors are not in storage and are decoded correctly using the latest stored runtime code.
- events are only stored in the actual block the event is emitted, this is safe because events storage is never read directly
- enum stored in any other storage item are potentially unsafe to modify

I realized this after updating the proxy pallet one day and [an event was inserted before end and it was fine](https://github.com/paritytech/polkadot-sdk/pull/7995/files#diff-c434cd259ed280c05a395c61be631e464b9454a5a331e2ce393e196d091d837d)

Only case it breaks is if clients use hardcoded metadata/index instead of last one as @camfairchild said

pallets/subtensor/src/macros/events.rs
lorismoulin: @shamil-gadelshin see my previous response

pallets/subtensor/src/macros/errors.rs
shamil-gadelshin: The current solution should work, as Cameron said, when clients use the latest metadata from the chain. Polkadot.js seems to reload the metadata on runtime upgrades. I'm not sure how the transition actually works, but the client reload should fix any possible issues.

pallets/subtensor/src/macros/dispatches.rs
```diff
@@ -2432,5 +2348,110 @@ mod dispatches {
...
+            new_coldkey: T::AccountId,
+        ) -> DispatchResult {
+            let who = ensure_signed(origin)?;
+
+            let (when, new_coldkey_hash) = ColdkeySwapAnnouncements::<T>::take(who.clone())
+                .ok_or(Error::<T>::ColdkeySwapAnnouncementNotFound)?;
+
+            ensure!(
+                new_coldkey_hash == T::Hashing::hash_of(&new_coldkey),
+                Error::<T>::AnnouncedColdkeyHashDoesNotMatch
```
gztensor, line 2426: @l0r1s , why do we need to hide the new coldkey behind a hash? Technically, this extrinsic discloses it before transaction is in the block.

lorismoulin: > @l0r1s , why do we need to hide the new coldkey behind a hash? Technically, this extrinsic discloses it before transaction is in the block.

The new coldkey is first announced using its hash (`announce_coldkey_swap`), and then swapped using `swap_coldkey_announced` (given the provided new coldkey map the hash provided for the announcement) which go through the MEV shield on the python side to prevent "front running"/blocking the swap. This come after an issue that happened few weeks ago, maybe @camfairchild can explain better than me the exact situation.

Hence, the new coldkey is known only when the swap has been executed.

pallets/subtensor/src/macros/dispatches.rs
```diff
@@ -1333,94 +1329,15 @@ mod dispatches {
...
-        ///
-        /// - Implement proper weight calculation based on the complexity of the operation.
-        /// - Consider adding checks to prevent scheduling too far into the future.
-        /// TODO: Benchmark this call
+        /// WARNING: This function is deprecated, please migrate to `announce_coldkey_swap`/`coldkey_swap`
         #[pallet::call_index(73)]
-        #[pallet::weight((Weight::from_parts(37_830_000, 0)
-		.saturating_add(T::DbWeight::get().reads(4))
-		.saturating_add(T::DbWeight::get().writes(2)), DispatchClass::Normal, Pays::Yes))]
+        #[pallet::weight(Weight::zero())]
```
shamil-gadelshin: Consider returning some weight - it could decrease spam probability.

pallets/subtensor/src/transaction_extension.rs
```diff
@@ -85,56 +85,62 @@ where
...
 
-        // Verify ColdkeySwapScheduled map for coldkey
-        match call.is_sub_type() {
-            // Whitelist
-            Some(Call::schedule_swap_coldkey { .. }) => {}
-            _ => {
-                if ColdkeySwapScheduled::<T>::contains_key(who) {
-                    return Err(CustomTransactionError::ColdkeyInSwapSchedule.into());
-                }
+        if ColdkeySwapAnnouncements::<T>::contains_key(who) {
```
shamil-gadelshin: How will it deal with proxies?

lorismoulin: @shamil-gadelshin good catch thank you, I updated it to handle the case where a transaction is a proxy call. Extracted this to a new tx extension but called it directly from the Subtensor one to avoid node update.

pallets/subtensor/src/extensions/subtensor.rs
```diff
@@ -85,56 +85,58 @@ where
...
-            // Whitelist
-            Some(Call::schedule_swap_coldkey { .. }) => {}
-            _ => {
-                if ColdkeySwapScheduled::<T>::contains_key(who) {
-                    return Err(CustomTransactionError::ColdkeyInSwapSchedule.into());
-                }
-            }
-        }
+        // TODO: move into tx extension pipeline but require node upgrade
+        CheckColdkeySwap::<T>::new().validate(
```
shamil-gadelshin, line 126: Why do we need a node upgraded for this? Isn't it a runtime configuration? 
Also, the current implementation works with EVM implicitly. It will stop working when adding to the tx extension pipeline without changing the EVM extensions.

pallets/subtensor/src/transaction_extension.rs
shamil-gadelshin: I wonder what causes the node upgrade.

shamil-gadelshin (approved):

pallets/subtensor/src/extensions/subtensor.rs
lorismoulin: Adding a new transaction extension require updating the MEV shield code because it uses the pipeline [here](https://github.com/opentensor/subtensor/blob/f74d69ed52e66c42476c94cdbeac8018f5c5567b/node/src/mev_shield/author.rs#L378)

shamil-gadelshin: Makes sense!

sam0x17 (approved):

PR #2262, opened 2025-12-05
MevShield Remove fees & Block external

JohnReedV:
## Description
This PR modifies the `announce_next_key` extrinsic to stop charging fees and instead block external calls to prevent spam.


ztsalexey:
Do not think this is needed anymore. Investigating now but there is no more mev

JohnReedV:
> Do not think this is needed anymore. Investigating now but there is no more mev

Don't know what you mean.

ztsalexey:
There is no frontrunning anymore. 99% confirmed

PR #2264, opened 2025-12-08
Filter for incorrect sudo calls

grbIzl:
## Description
Incorrect Sudo transactions are filtered out inside TransactionExtension to prevent adding them into the transactions pool.

## Related Issue(s)

- Closes #[issue number]

## Type of Change

- [ ] Documentation update

## Breaking Change

## Checklist

## Additional Notes


shamil-gadelshin (changes requested):
Why do we add a sudo dependency to subtensor? Can we add the new tx extension to the runtime code instead?

grbIzl:
> Why do we add a sudo dependency to subtensor? Can we add the new tx extension to the runtime code instead?

Good idea. It will be cleaner in this case.

shamil-gadelshin (dismissed):

runtime/src/sudo_wrapper.rs
```diff
@@ -0,0 +1,98 @@
...
+    fn prepare(
+        self,
+        _val: Self::Val,
+        _origin: &<T as Config>::RuntimeOrigin,
+        _call: &<T as Config>::RuntimeCall,
+        _info: &DispatchInfoOf<<T as Config>::RuntimeCall>,
+        _len: usize,
+    ) -> Result<Self::Pre, TransactionValidityError> {
+        Ok(())
+    }
```
lorismoulin:
```suggestion
impl_tx_ext_default!(<T as frame_system::Config>::RuntimeCall; weight prepare);
```

runtime/src/sudo_wrapper.rs
```diff
@@ -0,0 +1,98 @@
...
+{
+    const IDENTIFIER: &'static str = "SudoTransactionExtension";
+
+    type Implicit = ();
+    type Val = Option<T::AccountId>;
+    type Pre = ();
+
+    fn weight(&self, _call: &<T as Config>::RuntimeCall) -> Weight {
+        Weight::zero()
+    }
```
lorismoulin:
```suggestion
```

To combine with the one below

runtime/src/sudo_wrapper.rs
```diff
@@ -0,0 +1,98 @@
...
+    TransactionExtension<<T as Config>::RuntimeCall> for SudoTransactionExtension<T>
+where
+    <T as Config>::RuntimeCall: Dispatchable<Info = DispatchInfo, PostInfo = PostDispatchInfo>,
+    <T as Config>::RuntimeOrigin: AsSystemOriginSigner<T::AccountId> + Clone,
+    <T as Config>::RuntimeCall: IsSubType<SudoCall<T>>,
+{
+    const IDENTIFIER: &'static str = "SudoTransactionExtension";
+
+    type Implicit = ();
+    type Val = Option<T::AccountId>;
```
lorismoulin:
```suggestion
    type Val = ();
```

I get the impression the `Val` is never used? So may be better to just use unit type.

It is only used if you have a `prepare` implementation because it is passed as a parameter.

runtime/src/lib.rs
```diff
@@ -1608,6 +1611,7 @@ pub type TransactionExtensions = (
     frame_system::CheckEra<Runtime>,
     check_nonce::CheckNonce<Runtime>,
     frame_system::CheckWeight<Runtime>,
+    SudoTransactionExtension<Runtime>,
     ChargeTransactionPaymentWrapper<Runtime>,
```
lorismoulin: I'm not sure if this implies that sudo transaction without sudo origin won't be charged? Maybe it's safer to put this after the `ChargeTransactionPaymentWrapper` ?

l0r1s (changes requested):
small comments

runtime/src/sudo_wrapper.rs
grbIzl: I also use it in return type for validate.

lorismoulin: > I also use it in return type for validate.

Yes, but the [`validate` return value is only used in the `prepare`](https://paritytech.github.io/polkadot-sdk/master/polkadot_sdk_frame/traits/trait.TransactionExtension.html#associatedtype.Val) function and because there is no implementation here, it is not used at all so you can just remove it from `Val` and from return type of `validate`

l0r1s (approved):

l0r1s (approved):

sam0x17 (approved):

PR #2266, opened 2025-12-08
MevShield / Transaction Extension

JohnReedV:
## Description
This PR modifies the `MevShield` pallet and adds a transaction extension for validation checks of *signed* extrinsics.


sam0x17 (approved):

ztsalexey:
Why is this still being pushed? No clarity on what happened past week and encryption is not required anymore since there is no frontrunning

PR #2267, opened 2025-12-08
Revert "MevShield / Transaction Extension"

ztsalexey:
Reverts opentensor/subtensor#2266

According to my tests with relay service which was shielding transactions previously (just providing UX for scheduling) - during the week of Dec 1st, there was some change within consensus and as of my latest tests on Dec 5th - I've confirmed that relay has stopped working because something has changed. Also I've confirmed that there is no frontrunning anymore due to this change. Confirmed directly on mainnet through series of tests. 

1. Was there clarity on what has changed after December 2nd-3rd?
2. If there is no frontrunning present + lack of clarity - why is the encryption being pushed (solving non-existing issue and trust assumptions, which were in a way violated by silent changes past week)

### DIFF
--- a/pallets/shield/src/lib.rs
+++ b/pallets/shield/src/lib.rs
@@ -16,22 +16,20 @@ pub mod pallet {
     use super::*;
     use codec::Encode;
     use frame_support::{
-        dispatch::{DispatchInfo, GetDispatchInfo, PostDispatchInfo},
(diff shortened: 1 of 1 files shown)


sam0x17:
hi, there is still different kinds of MeV going on, additionally opt-in encrypted mempool is a very powerful feature that can be used for more than just MeV prevention

PR #2278, opened 2025-12-11
Add root claim maps.

shamil-gadelshin:
This PR introduces two "accounting maps" that are populated during the root claim:
```
    /// --- MAP ( subnet ) --> claimed root alpha (kept)
    #[pallet::storage]
    pub type SubnetRootClaimKeep<T: Config> =
        StorageMap<_, Blake2_128Concat, NetUid, AlphaCurrency, ValueQuery, DefaultZeroAlpha<T>>;

    /// --- MAP ( subnet ) --> claimed root alpha (swapped)
    #[pallet::storage]
    pub type SubnetRootClaimSwap<T: Config> =
        StorageMap<_, Blake2_128Concat, NetUid, AlphaCurrency, ValueQuery, DefaultZeroAlpha<T>>;
```

Maps will be cleared on the network deregistration.
Tests verify alpha addition during root claim for both swap and keep operations, and map clearing on network deregistration.

### DIFF
--- a/pallets/subtensor/src/coinbase/root.rs
+++ b/pallets/subtensor/src/coinbase/root.rs
@@ -210,7 +210,7 @@ impl<T: Config> Pallet<T> {
             Error::<T>::SubnetNotExists
         );
 
-        Self::finalize_all_subnet_root_dividends(netuid);
+        Self::clear_root_claim_root_data(netuid);
 
         // --- Perform the cleanup before removing the network.
         T::SwapInterface::dissolve_all_liquidity_providers(netuid)?;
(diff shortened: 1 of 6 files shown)


bdmason:
@shamil-gadelshin sorry I put this on the wrong PR before:

When we chatted I said these were fine as Alpha, but now I'm trying to work with it I've realised we need them as TAO. To use them practically we need to take diffs between blocks, but the alpha swapped/kept at different blocks isn't priced the same.

For example, if in block 10 the price of alpha is 0.5, then by block 20 the price is 0.8, and in between those blocks random amounts of alpha were sold each block, we would read the diffs of the root kept/swapped values at block 10 & 20 to find the flow during those 10 blocks, but we wouldn't know what alpha price to use without reading storage at every single block.

shamil-gadelshin:
Update: original maps were removed in favor of the temporary (in-block) storage and event from on_finalize().

PR #2280, opened 2025-12-11
Change default from Swap to Keep for delegates.

unconst:
## Description
Changes default Keep/Swap type to Keep. Caps the alpha in emission to tao emission.

### DIFF
--- a/pallets/subtensor/src/lib.rs
+++ b/pallets/subtensor/src/lib.rs
@@ -329,11 +329,11 @@ pub mod pallet {
     )]
     /// Enum for the per-coldkey root claim setting.
     pub enum RootClaimTypeEnum {
-        /// Swap any alpha emission for TAO.
-        #[default]
-        Swap,
         /// Keep all alpha emission.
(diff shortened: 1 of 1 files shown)


pallets/subtensor/src/lib.rs
```diff
@@ -329,11 +329,11 @@ pub mod pallet {
...
     /// Enum for the per-coldkey root claim setting.
     pub enum RootClaimTypeEnum {
-        /// Swap any alpha emission for TAO.
-        #[default]
-        Swap,
         /// Keep all alpha emission.
+		#[default]
         Keep,
+		/// Swap any alpha emission for TAO.
+        Swap,
```
shamil-gadelshin, line 336: We shouldn't change the order because of the underlying representation (numbers).

shamil-gadelshin:
Here is the updated version.

shamil-gadelshin:
https://github.com/opentensor/subtensor/pull/2282

PR #2286, opened 2025-12-12
Add tao inflow recording.

shamil-gadelshin:


gztensor (approved):

shamil-gadelshin:
cc @0xcacti @camfairchild

bdmason:
When we chatted I said these were fine as Alpha, but now I'm trying to work with it I've realised we need them as TAO. To use them practically we need to take diffs between blocks, but the alpha swapped/kept at different blocks isn't priced the same.

For example, if in block 10 the price of alpha is 0.5, then by block 20 the price is 0.8, and in between those blocks random amounts of alpha were sold each block, we would read the diffs of the root kept/swapped values at block 10 & 20 to find the flow during those 10 blocks, but we wouldn't know what alpha price to use without reading storage at every single block.

shamil-gadelshin:
> ... these were fine as Alpha, but now I'm trying to work with it I've realised we need them as TAO...

Not sure I follow - this "sub-PR" is about recording TAO in `SubnetTaoFlow` . 

> For example, if in block 10 the price of alpha is 0.5, then by block 20 the price is 0.8, and in between those blocks random amounts of alpha were sold each block, we would read the diffs of the root kept/swapped values at block 10 & 20 to find the flow during those 10 blocks, but we wouldn't know what alpha price to use without reading storage at every single block.

 The PR is using a moving average price for TAO in flow recording for `Keep` operation. Those values are recorded each block and could be retrieved for specific blocks (`SubnetMovingPrice`).

bdmason:
> When we chatted I said these were fine as Alpha, but now I'm trying to work with it I've realised we need them as TAO. To use them practically we need to take diffs between blocks, but the alpha swapped/kept at different blocks isn't priced the same.
> 
> For example, if in block 10 the price of alpha is 0.5, then by block 20 the price is 0.8, and in between those blocks random amounts of alpha were sold each block, we would read the diffs of the root kept/swapped values at block 10 & 20 to find the flow during those 10 blocks, but we wouldn't know what alpha price to use without reading storage at every single block.

Sorry Shamil I commented on the wrong PR, I just looked at the other and wondered where my comment had gone!

PR #2290, opened 2025-12-15
Balancer swap

gztensor:
## Description

Implement PalSwap (similar to balancer swap) to replace Uniswap V3.

Unlike uniswap v2 or v3, it allows adding liquidity unproportionally to price. This is achieved by introducing the weights w1 and w2 so that w1 + w2 = 1. In these formulas x means base currency (alpha) and y means quote currency (tao). The w1 weight in the code is referred as weight_base, and w2 as weight_quote. Because of the w1 + w2 = 1 constraint, only weight_quote is stored, and weight_base is always calculated.

The formulas used for pool operation are following:

**Price**

p = (w1 * y) / (w2 * x)

**Reserve deltas** 

(or -1 * payouts) in swaps are computed by

if ∆x is given (sell): ∆y = y * ((x / (x+∆x))^(w1/w2) - 1)
if ∆y is given (buy):  ∆x = x * ((y / (y+∆y))^(w2/w1) - 1)

**Limit orders**

When swaps are executing the orders with slippage control, we need to know what amount we can swap before the price reaches the limit value of p':

If p' < p (sell): ∆x = x * ((p / p')^w2 - 1)
If p' < p (buy):  ∆y = y * ((p' / p)^w1 - 1)

**Swap initialization**

In order to initialize weights with existing reserve values and price (for existing subnets):

w1 = px / (px + y)
w2 = y / (px + y)

**Injection**

Weights are adjusted when some amounts are added to the reserves. This prevents price from changing.

new_w1 = p * (x + ∆x) / (p * (x + ∆x) + y + ∆y)
new_w2 = (y + ∆y) / (p * (x + ∆x) + y + ∆y)

**User Liquidity**

The user liquidity stays disabled for now. The last commit # where all comments about user liquidity and tests (commented out) are present: fb0017fa3ba2f31dba3bc2d4917d8873ba1e759f , diff it with 4e21da29372b0f04d66a5866e043d7f03851693e to see the user liquidity code.

## Type of Change
- [ ] Documentation update

## Breaking Change

This PR changes the following externally accessible entities:

- `add_liqudiity` extrinsic: Remove tick_low and tick_high parameters.
- `Positions` map is replaced with `PositionsV2` map
- Following state maps removed:
  - AlphaSqrtPrice
  - CurrentLiquidity
  - CurrentTick
  - FeeGlobalTao
  - FeeGlobalAlpha
  - ScrapReservoirTao (applied before removing)
  - ScrapReservoirAlpha (applied before removing)
  - Ticks
  - SwapV3Initialized
  - TickIndexBitmapWords

## Checklist


Cargo.toml
gztensor, line 1: We're using Sam's bigmath crate now to handle high precision exponents.

chain-extensions/src/tests.rs
gztensor, line 1: I have changed U96F32 to U64F64 for representation of price. F64 has more fractional bits and prices are virtually never greater than 1000 anyway, which is way below u64::MAX.

pallets/subtensor/src/lib.rs
gztensor, line 1: Removing storage for provided reserves. From now on, we can calculate provided reserves using balancer very precisely, no need to store them separately. Anyway, this is only relevant when/if we decide to go for user liquidity with balancer.

pallets/subtensor/src/migrations/migrate_cleanup_swap_v3.rs
gztensor, line 1: If there are any left-overs in provided reserves, add them to reserves before removing.

pallets/swap-interface/src/lib.rs
gztensor, line 1: adjust_protocol_liquidity needs to report how much it actually added to reserves because it is adding fees and this addition needs to be considered when we modify SubnetTAO and SubnetAlphaIn in run_coinbase.

pallets/swap-interface/src/order.rs
gztensor, line 1: We are not operating square root of prices anymore, so renamed this.

pallets/swap/src/pallet/balancer.rs
gztensor, line 1: This file contains the heart of balancer math. All swaps come through this logic. Please read the comments to understand what we're doing.

pallets/swap/src/pallet/impls.rs
gztensor, line 1: This file is where uniswap v3 logic used to be. Reworked to use balancer.

pallets/swap/src/pallet/migrations/migrate_swapv3_to_balancer.rs
gztensor, line 1: Now we can remove all swap v3 maps.

pallets/swap/src/pallet/mod.rs
gztensor, line 1: Removed swap v3 maps and replaced with balancer maps.

pallets/swap/src/pallet/swap_step.rs
gztensor, line 1: Unline uniswap v3, balancer swap is always a single step, but yet, swap_step provides cleaner separation of swap logic, and I decided to preserve the structure. See how much simpler it is now.

pallets/subtensor/src/coinbase/run_coinbase.rs
gztensor, line 1: adjust_protocol_liquidity returns the actual adjustment that needs to be made to reserves, which includes previously collected tao and alpha fees.

When tao fees are charged, the tao is coming from user balance to FeesTao map, not accounted for in SubnetTAO, so it needs to be added here.

When alpha fees are charged, the alpha is coming from user's stake, and is also not accounted for in SubnetAlphaIn.

pallets/subtensor/src/staking/move_stake.rs
gztensor, line 1: This is still an approximation.

pallets/subtensor/src/staking/stake_utils.rs
gztensor, line 1: Continue removing storage for provided reserves here

pallets/swap/src/position.rs
gztensor, line 1: Positions are unused atm, but I preserved the struct for later use, when we have user liquidity.

pallets/swap/src/tick.rs
gztensor, line 1: This is part of uniswap v3 logic. Not needed anymore.

pallets/swap/src/pallet/balancer.rs
```diff
@@ -0,0 +1,1085 @@
...
+        // Calculate new reserve weights
+        let new_reserve_weight = if q_sum != 0 {
+            // Both TAO and Alpha are non-zero, normal case
+            Perquintill::from_rational(quantity_2, q_sum)
+        } else {
+            // Either TAO or Alpha reserve were and/or remain zero => Initialize weights to 0.5
+            Perquintill::from_rational(1u128, 2u128)
+        };
+
+        self.set_quote_weight(new_reserve_weight)
```
0xcacti, line 289: looks like you forgot to set the base weight here?  we only set the quote weight.

0xcacti: how could we have possibly gotten through testing? are we missing something?

0xcacti: nevermind, base is always looked up as 1 - quote

pallets/subtensor/src/coinbase/block_emission.rs
0xcacti, line 1: We should just delete this whole function in which we have changes.  The function isn't used anywhere other than in a test, and the logic for emissions is wrong.

pallets/subtensor/src/coinbase/root.rs
0xcacti, line 1: I love cleaning.

pallets/swap-interface/src/lib.rs
0xcacti: Yeah, good point. correc.t

pallets/swap/src/pallet/impls.rs
```diff
@@ -40,176 +39,143 @@ pub struct RemoveLiquidityResult {
...
-                Positions::<T>::insert(
-                    (netuid, protocol_account_id, position.id),
-                    position.clone(),
-                );
-
-                // Update position ticks
-                Self::add_liquidity_at_index(netuid, position.tick_low, liquidity_delta, false);
-                Self::add_liquidity_at_index(netuid, position.tick_high, liquidity_delta, true);
-            }
+    ) -> (TaoCurrency, AlphaCurrency) {
```
0xcacti, line 95: This is high level correct.

of course, you know that I think it's really weird the way that the reserves are in another pallet so we update them in the calling context rather than here, but I suppose my view there doesn't really matter.

As is, it's correct.

pallets/subtensor/src/coinbase/run_coinbase.rs
```diff
@@ -67,7 +67,8 @@ impl<T: Config> Pallet<T> {
             let tao_to_swap_with: TaoCurrency =
                 tou64!(excess_tao.get(netuid_i).unwrap_or(&asfloat!(0))).into();
 
-            T::SwapInterface::adjust_protocol_liquidity(*netuid_i, tao_in_i, alpha_in_i);
+            let (actal_injected_tao, actal_injected_alpha) =
```
0xcacti: typo

says "actal" instead of "actual"

pallets/subtensor/src/coinbase/run_coinbase.rs
```diff
@@ -87,15 +88,17 @@ impl<T: Config> Pallet<T> {
...
             });
 
             // Inject TAO in.
             let injected_tao: TaoCurrency =
                 tou64!(*tao_in.get(netuid_i).unwrap_or(&asfloat!(0))).into();
             SubnetTaoInEmission::<T>::insert(*netuid_i, injected_tao);
             SubnetTAO::<T>::mutate(*netuid_i, |total| {
-                *total = total.saturating_add(injected_tao);
+                // Reserves also received fees in addition to injected_tao
+                *total = total.saturating_add(actal_injected_tao);
```
0xcacti: actal typo.

pallets/subtensor/src/coinbase/run_coinbase.rs
```diff
@@ -87,15 +88,17 @@ impl<T: Config> Pallet<T> {
                 AlphaCurrency::from(tou64!(*alpha_in.get(netuid_i).unwrap_or(&asfloat!(0))));
             SubnetAlphaInEmission::<T>::insert(*netuid_i, alpha_in_i);
             SubnetAlphaIn::<T>::mutate(*netuid_i, |total| {
-                *total = total.saturating_add(alpha_in_i);
+                // Reserves also received fees in addition to alpha_in_i
+                *total = total.saturating_add(actal_injected_alpha);
```
0xcacti: actal typo.

pallets/subtensor/src/coinbase/run_coinbase.rs
```diff
@@ -140,7 +143,8 @@ impl<T: Config> Pallet<T> {
             log::debug!("alpha_emission_i: {alpha_emission_i:?}");
 
             // Get subnet price.
-            let price_i: U96F32 = T::SwapInterface::current_alpha_price(netuid_i.into());
+            let price_i: U96F32 =
+                U96F32::saturating_from_num(T::SwapInterface::current_alpha_price(netuid_i.into()));
```
0xcacti, line 147: Any risk in this cast?  I assume not because Greg is the type master.

pallets/subtensor/src/staking/move_stake.rs
0xcacti: No formula for this exists, but we could do newtons method.  We have an implementation if you want it. 

@gztensor see discord.

pallets/subtensor/src/staking/stake_utils.rs
0xcacti: yay.

pallets/subtensor/src/lib.rs
0xcacti: : )

pallets/swap/src/pallet/hooks.rs
0xcacti, line 1: I don't entirely understand this file

pallets/swap/src/pallet/mod.rs
0xcacti: This is so much nicer.

pallets/swap/src/pallet/balancer.rs
```diff
@@ -0,0 +1,1090 @@
...
+            Ok(Balancer { quote })
+        } else {
+            Err(BalancerError::InvalidValue)
+        }
+    }
+
+    /// Constraints limit balancer weights within certain range of values:
+    ///   - Both weights are above minimum
+    ///   - Sum of weights is equal to 1.0
+    fn check_constraints(quote: Perquintill) -> bool {
```
0xcacti, line 105: We are working on inventing a more advanced theory here for when we reach the MIN_WEIGHT what we should do.  All good for now, but these constraints may change in the future.

pallets/swap/src/pallet/balancer.rs
```diff
@@ -0,0 +1,1090 @@
...
+    ///
+    /// Note that weights must remain within the proper range (both be above MIN_WEIGHT),
+    /// so only reasonably small disproportions of updates are appropriate.
+    pub fn update_weights_for_added_liquidity(
+        &mut self,
+        tao_reserve: u64,
+        alpha_reserve: u64,
+        tao_delta: u64,
+        alpha_delta: u64,
+    ) -> Result<(), BalancerError> {
```
0xcacti, line 260: Just want to make sure that here we are passing in the actual_delta, 
I.e. the quantity injected + fees

0xcacti: I checked and it is because it's only called by adjust_protocol_liqudity which handles this.

pallets/swap/src/pallet/balancer.rs
```diff
@@ -0,0 +1,1090 @@
...
+    }
+
+    /// Calculates current liquidity from alpha and tao reserves using the formula:
+    ///   L = x^w1 * y^w2
+    /// where
+    ///   x - alpha reserve
+    ///   y - tao reserve
+    ///   w1 - base weight
+    ///   w2 - quote weight
+    pub fn calculate_current_liquidity(&self, tao_reserve: u64, alpha_reserve: u64) -> u64 {
```
0xcacti: Do we even need this function for anything?  could probably get rid of it?

0xcacti: I guess if we want user liquidity at some point in the future.

0xcacti (approved):
We will provide one more review, looking at `pallets/swap/src/pallet/swap_step.rs` and also `pallets/swap/src/pallet/impls.rs`

it got late and we don't want to lose our progress.

pallets/swap/src/pallet/impls.rs
```diff
@@ -40,176 +39,143 @@ pub struct RemoveLiquidityResult {
...
         let tao_reserve = T::TaoReserve::reserve(netuid.into());
         let alpha_reserve = T::AlphaReserve::reserve(netuid.into());
 
-        // Set price
-        let price = U64F64::saturating_from_num(tao_reserve)
-            .safe_div(U64F64::saturating_from_num(alpha_reserve));
-
-        let epsilon = U64F64::saturating_from_num(0.000000000001);
+        // Create balancer based on price
+        let balancer = Balancer::new(if let Some(price) = maybe_price {
```
0xcacti, line 58: nice handles existing subnets just in case.  excellent.

0xcacti: just to be clear, when we pass in maybe_price 

is it the case that new subnets won't have a price at the time of initialization?

pallets/swap/src/pallet/impls.rs
```diff
@@ -40,176 +39,143 @@ pub struct RemoveLiquidityResult {
...
+            // Price is given, calculate weights:
+            // w_quote = y / (px + y)
+            let px_high = (price.saturating_to_num::<u64>() as u128)
```
0xcacti, line 61: We should have a test just to make sure this is correct.  I mean it makes sense, mult the integer part and the reserve then the fractional part and the reserve and do some fun type casting to get to the end type, but just want to be sure because it's nontrivial.

pallets/swap/src/pallet/impls.rs
```diff
@@ -289,70 +255,36 @@ impl<T: Config> Pallet<T> {
             Error::<T>::ReservesTooLow
         );
 
-        Self::maybe_initialize_v3(netuid)?;
+        Self::maybe_initialize_palswap(netuid, None)?;
```
0xcacti, line 213: Man should we just pass in a price of None here?  what if it's a subnet with a price?

pallets/swap/src/pallet/balancer.rs
gztensor: Yes, the only weight that is stored and needs to be set is quote. Base weight is always calculated from quote weight.

pallets/subtensor/src/coinbase/block_emission.rs
gztensor: Agree. I did it in a separate PR because it is not related to balancer: 
https://github.com/opentensor/subtensor/pull/2346

pallets/subtensor/src/coinbase/run_coinbase.rs
gztensor: The U64F64 -> U96F32 conversion only exists here because this PR changes the price from being U96F32 to U64F64: Less digits in its integer part and more digits in its fractional part. Nonetheless, for the purpose on run_coinbase it stays as it was before: U96F32. The risk is loss of lower 32 binary digits of fractional part of the price. The part of price below 0.000000000232831 is thrown away for calculating emissions. I don't think it's a big deal, tbh. But this is not introduced by this PR. It was always like this: We just never maintained the prices so precisely.

pallets/swap/src/pallet/hooks.rs
gztensor: Hooks in substrate execute without user interaction on certain events. In this case, I am `using on_runtime_upgrade` to run the migration. It is executed in the beginning of the block following after the runtime upgrade.

pallets/swap/src/pallet/balancer.rs
gztensor: Yes. I am leaving a few unused things behind. In case we decide to have user liquidity, they will help us have much smoother transition and less or no migrations:

- Positions
- Current liquidity
- Tests for user liquidity (all commented out)

pallets/swap/src/pallet/impls.rs
gztensor: New subnets get their price right here before the first swap. This price is based on whatever reserves were assigned at creation and 0.5 weights.

pallets/swap/src/pallet/impls.rs
gztensor: There is a test for this that involves both integer and fractional part of price:

https://github.com/opentensor/subtensor/blob/a3756d9b601b5a6522be330160c2bda62744d6d1/pallets/swap/src/pallet/tests.rs#L2723

Explanation: Converting U64F64 to u64 is the same as rounding to the floor integer. I am doing this to avoid both overflow and precision loss of u64 by U64F64 multiplication. This is very hypothetical, but if alpha_reserve is 21M and the price is 1000., it overflows U64F64.

pallets/swap/src/pallet/impls.rs
gztensor: If it's a subnet with a price, it would be handled in the migration right after the balancer upgrade and `maybe_initialize_palswap` will be a noop in this case. Here it is a non-noop only for new subnets.

pallets/subtensor/src/tests/claim_root.rs
```diff
@@ -1082,6 +1098,103 @@ fn test_claim_root_coinbase_distribution() {
     });
 }
```
shamil-gadelshin, line 1100: Is there a reason you don't restore this test?

pallets/subtensor/src/tests/staking.rs
```diff
@@ -6,9 +6,10 @@ use frame_support::dispatch::{DispatchClass, DispatchInfo, GetDispatchInfo, Pays
 use frame_support::sp_runtime::DispatchError;
 use frame_support::{assert_err, assert_noop, assert_ok, traits::Currency};
 use frame_system::RawOrigin;
-use pallet_subtensor_swap::tick::TickIndex;
+// use pallet_subtensor_swap::Call as SwapCall;
```
shamil-gadelshin: Leftover comments.

pallets/swap/src/pallet/impls.rs
```diff
@@ -40,176 +39,143 @@ pub struct RemoveLiquidityResult {
...
 
-        // Set current tick
-        let current_tick = TickIndex::from_sqrt_price_bounded(current_sqrt_price);
-        CurrentTick::<T>::set(netuid, current_tick);
+        // TODO: Review when/if we have user liquidity
+        // Initialize the pal-swap:
 
         // Set initial (protocol owned) liquidity and positions
-        // Protocol liquidity makes one position from TickIndex::MIN to TickIndex::MAX
+        // Protocol liquidity makes one position
```
shamil-gadelshin: Why do we keep the obsolete code here?

pallets/swap/src/position.rs
```diff
@@ -1,38 +1,29 @@
...
-    pub tick_low: TickIndex,
-    /// Tick index for higher boundary of price
-    pub tick_high: TickIndex,
-    /// Position liquidity
-    pub liquidity: u64,
-    /// Fees accrued by the position in quote currency (TAO) relative to global fees
-    pub fees_tao: I64F64,
-    /// Fees accrued by the position in base currency (Alpha) relative to global fees
-    pub fees_alpha: I64F64,
+    // /// Position liquidity
```
shamil-gadelshin: A lot of commented code. Do we need these types?

shamil-gadelshin (dismissed):
Looks good! Are we sure that the next PR will restore user liquidity? We left a lot of code comments that make the code harder to read.

basfroman:
Hey @gztensor, we are already preparing changes for `SDK` and `btcli` related to this PR.

But, also, we would like to clarify whether it is possible to implement two additional related functionalities for clients within current PR:
- `SwapRuntimeApi.get_subnet_prices` - to return the prices of all subnets (similar to `SwapRuntimeApi.get_subnet_price`)
- A way through which we can calculate slippage without doing a `sim_swap` call to the chain. Something in `SubnetInfoRuntimeApi.get_dynamic_info` and `SubnetInfoRuntimeApi.get_all_dynamic_info` to replace existing Uniswap logic

If you have any questions, please feel free to let us know and we will be happy to discuss them at the meeting.

pallets/subtensor/src/tests/claim_root.rs
gztensor: It's a copy of the test above it, forgot to cleanup, thank you.

pallets/subtensor/src/tests/staking.rs
gztensor: Addressed

sam0x17 (dismissed):

sam0x17 (dismissed):

tutacrypto:
@gregzaitsev can you detail the new method to fetch the alpha price from the chain with balancer pools?

PR #2299, opened 2025-12-16
Feat/transfer stake destination hotkey

### LINKED ISSUE #1377: transfer_stake should be able to transfer to a new hotkey
### Describe the bug

when using transfer_stake, it doesn't make any sense to keep the hotkey the same because the new coldkey will never own that old hotkey. This means that the hotkey won't even show up in any of the btcli commands when checking your stake if the stake has been transferred to you on some random hotkey owned by someone else. There should be a destination hotkey as well as destination coldkey

### To Reproduce

use transfer_stake and find out that you can't change the hotkey

### Expected behavior

you should be able to set a destination hotkey

### Screenshots

_No response_

### Environment

Linux

### Additional context

_No response_

MkDev11:
## Description

When using `transfer_stake`, users previously couldn't specify a destination hotkey - the stake would transfer to a new coldkey but remain on the original hotkey. This was problematic because the new coldkey doesn't own that hotkey, so the stake wouldn't show up in btcli commands for the recipient.

This PR adds a `destination_hotkey` parameter to `transfer_stake`, allowing users to transfer stake to both a different coldkey AND a different hotkey in a single operation. This makes the transfer actually useful - recipients can now receive stake on their own hotkey.

### What changed

The `transfer_stake` extrinsic now accepts:
- `destination_coldkey` - who receives the stake
- `origin_hotkey` - the hotkey to take stake from
- `destination_hotkey` - the hotkey to put stake on (NEW)
- `origin_netuid` / `destination_netuid` - subnet IDs
- `alpha_amount` - amount to transfer

Previously, the single `hotkey` parameter was used for both origin and destination.

## Related Issue(s)

- Closes #1377

## Type of Change

- [ ] Documentation update

## Breaking Change

This is a breaking change for existing callers of `transfer_stake`:

**Before:**
```rust
transfer_stake(destination_coldkey, hotkey, origin_netuid, destination_netuid, alpha_amount)
```

**After:**
```rust
transfer_stake(destination_coldkey, origin_hotkey, destination_hotkey, origin_netuid, destination_netuid, alpha_amount)
```

**Migration:** Existing callers should pass the same hotkey for both `origin_hotkey` and `destination_hotkey` to maintain the previous behavior.

**Affected interfaces:**
- Substrate extrinsic (call_index 86)
- EVM precompile: `transferStake(bytes32,bytes32,bytes32,uint256,uint256,uint256)`
- Chain extension (function ID 6)
- ink! contract interface

## Checklist

Screenshot 2025-12-16 095544

### Files changed

| File | Change |
|------|--------|
| `pallets/subtensor/src/macros/dispatches.rs` | Added `destination_hotkey` parameter to dispatch |
| `pallets/subtensor/src/staking/move_stake.rs` | Updated `do_transfer_stake` to use both hotkeys |
| `pallets/subtensor/src/macros/events.rs` | Updated `StakeTransferred` event to include both hotkeys |
| `precompiles/src/staking.rs` | Updated EVM precompile signature |
| `precompiles/src/solidity/stakingV2.sol` | Updated Solidity interface |
| `chain-extensions/src/lib.rs` | Updated chain extension |
| `contract-tests/bittensor/lib.rs` | Updated ink! contract interface |
| `pallets/transaction-fee/src/lib.rs` | Updated pattern matching |
| `docs/wasm-contracts.md` | Updated documentation |
| Test files | Updated all tests + added new test |

### Why this matters

Before this change, if Alice transferred stake to Bob, the stake would end up on Alice's hotkey but owned by Bob's coldkey. Bob couldn't see or manage this stake easily because his btcli commands filter by his own hotkeys.

Now, Alice can transfer stake directly to Bob's hotkey, making the transfer actually useful for the recipient.

### DIFF
--- a/chain-extensions/src/lib.rs
+++ b/chain-extensions/src/lib.rs
@@ -226,12 +226,13 @@ where
             }
             FunctionId::TransferStakeV1 => {
                 let weight = Weight::from_parts(160_300_000, 0)
-                    .saturating_add(T::DbWeight::get().reads(13_u64))
+                    .saturating_add(T::DbWeight::get().reads(14_u64))
                     .saturating_add(T::DbWeight::get().writes(6_u64));
 
                 env.charge_weight(weight)?;
 
-                let (destination_coldkey, hotkey, origin_netuid, destination_netuid, alpha_amount): (
+                let (destination_coldkey, origin_hotkey, destination_hotkey, origin_netuid, destination_netuid, alpha_amount): (
+                    T::AccountId,
                     T::AccountId,
                     T::AccountId,
                     NetUid,
@@ -244,7 +245,8 @@ where
                 let call_result = pallet_subtensor::Pallet::<T>::transfer_stake(
                     RawOrigin::Signed(env.caller()).into(),
                     destination_coldkey,
-                    hotkey,
+                    origin_hotkey,
+                    destination_hotkey,
                     origin_netuid,
                     destination_netuid,
                     alpha_amount,
--- a/chain-extensions/src/lib.rs
+++ b/chain-extensions/src/lib.rs
@@ -224,6 +224,7 @@ where
                     }
                 }
             }
+            // Backward-compatible TransferStakeV1 with 5 parameters (same hotkey for origin and destination)
             FunctionId::TransferStakeV1 => {
                 let weight = Weight::from_parts(160_300_000, 0)
                     .saturating_add(T::DbWeight::get().reads(14_u64))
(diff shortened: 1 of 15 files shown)


MkDev11:
@sam0x17 Please check the PR and let me know the result. Thanks.

camfairchild:
We should avoid updating EVM interfaces because they are breaking
Instead can add a new function and support the old behaviour

MkDev11:
> We should avoid updating EVM interfaces because they are breaking Instead can add a new function and support the old behaviour

@sam0x17 What do you think of that?

sam0x17:
I would defer to vune on this one, unless there is a way you can make this non-breaking

MkDev11:
> I would defer to vune on this one, unless there is a way you can make this non-breaking

Just updated again. This has been made non-breaking. The original `transferStake` (5 params) is preserved for backward compatibility, and a new `transferStakeV2` (6 params) was added for the new hotkey-to-hotkey transfer feature. Existing contracts will continue to work unchanged.

PR #2306, opened 2025-12-18
Revert root claim default type back to Swap

shamil-gadelshin:
This PR reverts the default root claim type change back to `Swap` while preserving the added test.


sam0x17:
context: research team is still contemplating how to properly do adjustments to this, so we are removing this change from today's release

sam0x17 (approved):

PR #2311, opened 2025-12-19
Set default weights for system pallets.

shamil-gadelshin:
This PR sets default weights for system pallets:
- frame_system
- pallet_timestamp

The default implementation `impl WeightInfo for ()` provides the same values, so the tx fee won't change.

### DIFF
--- a/runtime/src/lib.rs
+++ b/runtime/src/lib.rs
@@ -353,7 +353,7 @@ impl frame_system::Config for Runtime {
     type PreInherents = ();
     type PostInherents = ();
     type PostTransactions = ();
-    type ExtensionsWeightInfo = (); // frame_system exports only default extension weights
+    type ExtensionsWeightInfo = frame_system::SubstrateExtensionsWeight<Runtime>;
 }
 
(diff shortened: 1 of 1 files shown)


runtime/src/lib.rs
```diff
@@ -369,7 +369,7 @@ impl pallet_grandpa::Config for Runtime {
 
     type KeyOwnerProof = sp_core::Void;
 
-    type WeightInfo = ();
+    type WeightInfo = (); // pallet_grandpa exports only default implementation
```
shamil-gadelshin, line 374: The containing module seems to be private.

l0r1s (dismissed):

runtime/src/lib.rs
lorismoulin: > The containing module seems to be private.

Sorry, deleted my message after I checked the code!

l0r1s (dismissed):

open-junius (dismissed):

sam0x17 (approved):

PR #2313, opened 2025-12-20
feat: implement on-chain invariant enforcement with recovery mechanism

Dairus01:
# Subtensor Invariant Enforcement: Problem, Analysis & Fix

## 1. The Problem: Silent Economic Corruption
In a complex Delegated Proof of Stake (DPoS) system like Subtensor, accounting bugs can be catastrophic. If the total stake stored in `SubnetAlphaOut` drifts from the actual sum of individual neuron stakes (`TotalHotkeyAlpha`), or if emission injection exceeds the protocol's defined schedule ([BlockEmission](cci:1://file:///c:/Users/user/Desktop/subtensor/pallets/subtensor/src/lib.rs:464:4-468:5)), the network suffers from **silent economic corruption**.

Prior to this fix, such inconsistencies could persist undetected, potentially allowing malicious actors to exploit infinite money glitches or causing catastrophic inflation.

### **Q: Where should we check for consistency?**
**Decision:** [on_finalize](cci:1://file:///c:/Users/user/Desktop/subtensor/pallets/subtensor/src/macros/hooks.rs:40:8-50:9) hook.
**Rationale:** Invariants must hold true at the end of every block after all extrinsics and internal logic (like [run_coinbase](cci:1://file:///c:/Users/user/Desktop/subtensor/pallets/subtensor/src/coinbase/run_coinbase.rs:21:4-53:5)) have executed. Checking at [on_initialize](cci:1://file:///c:/Users/user/Desktop/subtensor/pallets/subtensor/src/macros/hooks.rs:11:8-38:9) is too late (corruption happened in previous block), and checking inside every extrinsic is too heavyweight.

### **Q: How do we balance safety vs. liveness?**
**Decision:** **Pause, Don't Panic (in Production).**
**Rationale:**
- **Panic:** If we panic in [on_finalize](cci:1://file:///c:/Users/user/Desktop/subtensor/pallets/subtensor/src/macros/hooks.rs:40:8-50:9), the block is discarded. If this happens deterministically, the chain **halts**. This is acceptable for testnets but fatal for mainnet.
- **event + Pause:** In production, we detect the violation, emit a **Critical** event, and **pause emissions** for the affected subnet. This "stops the bleeding" without stopping the blockchain.
- **Panic (Test):** In `cfg(test)`, we want immediate failure to catch bugs during development.

### **Q: How do we handle "dust" and non-injected sinks?**
**Decision:** **Inequality Check (`<=`).**
**Rationale:** Strict equality (`sum == total`) is dangerous for emissions because some TAO might be burned or added to total issuance directly (excess) without being injected into a subnet pool. The invariant `total_injected <= block_emission` ensures we never *create* more money than allowed, which is the critical safety property.

## 3. The Solution Implementation

We implemented a robust invariant enforcement system with the following components:

### A. Dedicated Invariant Module ([src/invariants.rs](cci:7://file:///c:/Users/user/Desktop/subtensor/pallets/subtensor/src/invariants.rs:0:0-0:0))
A new module specifically for consistency checks, separating audit logic from business logic.

- **[check_stake_invariant()](cci:1://file:///c:/Users/user/Desktop/subtensor/pallets/subtensor/src/invariants.rs:48:4-80:5)**:
    - Iterates all hotkeys in a subnet.
    - Sums `TotalHotkeyAlpha`.
    - Compares strictly with `SubnetAlphaOut`.
    - **Optimization:** Runs only at `BlocksSinceLastStep == 0` (epoch boundary) to minimize computational weight.

- **[check_emission_invariant()](cci:1://file:///c:/Users/user/Desktop/subtensor/pallets/subtensor/src/invariants.rs:23:4-46:5)**:
    - Sums `SubnetTaoInEmission` across all active subnets.
    - asserts `sum(injected) <= BlockEmission`.
    - Runs **every block** to catch inflation bugs immediately.

### B. Safety Mechanisms ([src/macros/hooks.rs](cci:7://file:///c:/Users/user/Desktop/subtensor/pallets/subtensor/src/macros/hooks.rs:0:0-0:0) & [events.rs](cci:7://file:///c:/Users/user/Desktop/subtensor/pallets/subtensor/src/macros/events.rs:0:0-0:0))
- **Hook:** checks are called at the very end of [on_finalize](cci:1://file:///c:/Users/user/Desktop/subtensor/pallets/subtensor/src/macros/hooks.rs:40:8-50:9).
- **Event:** `CriticalInvariantViolation(NetUid, Bytes)` alerts the world.
- **Storage:** `SubnetEmissionPaused` map prevents broken subnets from receiving further funds.

### C. Recovery Path ([src/macros/dispatches.rs](cci:7://file:///c:/Users/user/Desktop/subtensor/pallets/subtensor/src/macros/dispatches.rs:0:0-0:0))
- **Extrinsic:** [unpause_subnet_emission(netuid)](cci:1://file:///c:/Users/user/Desktop/subtensor/pallets/subtensor/src/macros/dispatches.rs:2434:8-2454:9).
- **Permission:** Root (Sudo) only.
- **Purpose:** Allows governance to fix the underlying state corruption and then resume normal operations manually.

### D. Comprehensive Testing ([src/tests/invariants.rs](cci:7://file:///c:/Users/user/Desktop/subtensor/pallets/subtensor/src/tests/invariants.rs:0:0-0:0))
- **Unit Tests:** Verify that:
    1. Normal operation passes.
    2. Stake mismatch triggers panic (in tests).
    3. Emission violation triggers panic (in tests).
    4. Paused subnets are skipped (preventing infinite error loops).
    5. Governance can unpause the subnet.

## 4. Verification
The changes have been pushed to the `fix/invariant-enforcement` branch.
- **Git Remote:** `https://github.com/Dairus01/subtensor`
- **Branch:** `fix/invariant-enforcement`

You can verify the fix by running:
```bash
cargo test --package pallet-subtensor --lib -- tests::invariants

### DIFF
--- a/pallets/subtensor/src/coinbase/run_coinbase.rs
+++ b/pallets/subtensor/src/coinbase/run_coinbase.rs
@@ -2,8 +2,16 @@ use super::*;
 use alloc::collections::BTreeMap;
 use safe_math::*;
 use substrate_fixed::types::U96F32;
+use frame_support::traits::{
+    Imbalance,
+    tokens::{
+        Precision, Preservation,
+        fungible::{Balanced, Inspect, Mutate},
+    },
+};
 use subtensor_runtime_common::{AlphaCurrency, Currency, NetUid, TaoCurrency};
 use subtensor_swap_interface::SwapHandler;
+use frame_support::traits::tokens::imbalance::Credit;
 
 // Distribute dividends to each hotkey
 macro_rules! asfloat {
@@ -43,8 +51,11 @@ impl<T: Config> Pallet<T> {
         let root_sell_flag = Self::get_network_root_sell_flag(&subnets_to_emit_to);
         log::debug!("Root sell flag: {root_sell_flag:?}");
 
-        // --- 4. Emit to subnets for this block.
-        Self::emit_to_subnets(&subnets_to_emit_to, &subnet_emissions, root_sell_flag);
+        // --- 4. Mint emission and emit to subnets for this block.
+        // We mint the emission here using the Imbalance trait to ensure conservation.
+        let emission_u64: u64 = block_emission.saturating_to_num::<u64>();
+        let imbalance = T::Currency::issue(emission_u64);
+        Self::emit_to_subnets(&subnets_to_emit_to, &subnet_emissions, root_sell_flag, imbalance);
 
         // --- 5. Drain pending emissions.
         let emissions_to_distribute = Self::drain_pending(&subnets, current_block);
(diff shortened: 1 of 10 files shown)


Dairus01:
@open-junius @camfairchild @sam0x17 @gztensor 

Please, I would love your review on this PR

camfairchild:
It would make much more sense to implement something using Imbalances.

Also, running this logic every epoch doesn't seem like a good idea at all.

Dairus01:
> It would make much more sense to implement something using Imbalances.
> 
> Also, running this logic every epoch doesn't seem like a good idea at all.

Thanks for the feedback. I’ve refactored emission logic to use Substrate Imbalances throughout. Block emission now mints a PositiveImbalance, which is transactionally consumed during subnet distribution, making conservation enforced by construction. The epoch-wide emission invariant has been removed accordingly.

open-junius (changes requested):
Need fix the clippy issue at first

Dairus01:
> Need fix the clippy issue at first

I have reworked on it

PR #2317, opened 2025-12-22
fix: DefaultPendingChildKeyCooldown should use prod_or_fast value

### LINKED ISSUE #2107: `DefaultPendingChildKeyCooldown` should use `prod_or_fast!(7_200, 15)` instead of `0`
### Context: 
1. function `set_children` executes `do_schedule_children`
2. function `do_schedule_children` uses `PendingChildKeyCooldown` storage to get `cooldown_block`
3. storage `PendingChildKeyCooldown` uses `DefaultPendingChildKeyCooldown` which is `0` by default.
Theoretically, if we compile a chain from scratch without history, we'll get `PendingChildKeyCooldown` equal to `0` instead of `7200/15`. This happens with localnet now.

### Fix should include:
1. `DefaultPendingCooldown` is `prod_or_fast!(7_200, 15)`
2. `DefaultPendingChildKeyCooldown` gets from `DefaultPendingCooldown`.

DanielDerefaka:
## Summary
`DefaultPendingChildKeyCooldown` was returning `0` instead of using the same value as `DefaultPendingCooldown`. This causes `PendingChildKeyCooldown` to be `0` on fresh chains (like localnet) instead of the expected `7200` (prod) or `15` (fast/test).

## Changes
- Changed `DefaultPendingChildKeyCooldown` to call `DefaultPendingCooldown::<T>::get()` for consistency

## Before
```rust
pub fn DefaultPendingChildKeyCooldown<T: Config>() -> u64 {
    0
}
```

## After
```rust
pub fn DefaultPendingChildKeyCooldown<T: Config>() -> u64 {
    DefaultPendingCooldown::<T>::get()
}
```

Fixes #2107

### DIFF
--- a/pallets/subtensor/src/lib.rs
+++ b/pallets/subtensor/src/lib.rs
@@ -2359,10 +2359,11 @@ pub mod pallet {
     #[pallet::storage]
     pub type HasMigrationRun<T: Config> = StorageMap<_, Identity, Vec<u8>, bool, ValueQuery>;
 
-    /// Default value for pending childkey cooldown (settable by root, default 0)
+    /// Default value for pending childkey cooldown (settable by root).
+    /// Uses the same value as DefaultPendingCooldown for consistency.
     #[pallet::type_value]
     pub fn DefaultPendingChildKeyCooldown<T: Config>() -> u64 {
-        0
+        DefaultPendingCooldown::<T>::get()
     }
 
     /// Storage value for pending childkey cooldown, settable by root.


DanielDerefaka:
@sam0x17 Would you mind reviewing this fix? It addresses issue #2107 where DefaultPendingChildKeyCooldown was returning 0 instead of using the proper prod_or_fast value.

JohnReedV (approved):

sam0x17 (approved):

PR #2327, opened 2025-12-29
Chain buys go to owners of other subnets

unconst:
## Description
Bittensor for the subnets.

Chain buys are a novel update to the Bittensor dynamic TAO mechanism through which subnets attain direct chain purchases when their price determined alpha_in emission exceeds 0.5 per block. The effect is exciting, transitioning Bittensor's emission away from a purely price neutral injection to a more price-positive position.

There are a number of downsides to this change however, for one, chain buys open up potential for abuse where subnets attempt to game the emission computation (notably TaoFlow) in the short term, attain direct chain buys over a short window, and then extract the TAO used to purchase the subnet. This can be done without any outside interest from the community, and with no regard to the subnet producing value -- indeed, subnets that attempt this almost universally burn 100% of their miner emission since it inconveniences the exploit. 

The solution lies in distribution: there must be individuals in the ecosystem that can "shit test" (sell) the distribution of Bittensor emission through market dynamics. However, without any individuals owning the subnet in the first place -- there exists no miner emission, all root dividends are automatically sold, and no outside buyers exist -- there is correspondingly no one to sell. No sell pressure makes gaming TAO flow cheap, low risk and an obvious attack vector. 

The change in this PR is justified simply: if we distribute the chain bought alpha to subnet owners at large, then this group of individuals can police the worst aspects of Bittensor's subnets, attain that value to fund their own projects, and collectively coalesce more global integration. In the worst case, a subnet excluded from chain buys does not experience any downward price consequences from other owners selling their token (since it is matched with chain buys).

### DIFF
--- a/pallets/subtensor/src/coinbase/run_coinbase.rs
+++ b/pallets/subtensor/src/coinbase/run_coinbase.rs
@@ -78,7 +78,7 @@ impl<T: Config> Pallet<T> {
                 );
                 if let Ok(buy_swap_result_ok) = buy_swap_result {
                     let bought_alpha: AlphaCurrency = buy_swap_result_ok.amount_paid_out.into();
-                    Self::recycle_subnet_alpha(*netuid_i, bought_alpha);
+                    Self::distribute_alpha_to_subnet_owners(*netuid_i, bought_alpha);
                 }
             }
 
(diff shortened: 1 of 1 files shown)


pallets/subtensor/src/coinbase/run_coinbase.rs
```diff
@@ -111,6 +111,84 @@ impl<T: Config> Pallet<T> {
...
+                );
+
+                // Stake the alpha share to the owner coldkey on the owner hotkey for this subnet
+                Self::increase_stake_for_hotkey_and_coldkey_on_subnet(
+                    &owner_hotkey,
+                    &owner_coldkey,
+                    netuid,
+                    alpha_share,
+                );
+            }
```
ppolewicz, line 188: and if we cannot get the subnet owner coldkey/hotkey for some reason, then what happens? Shouldn't we accumulate the amount and recycle it at the end of the function?

pallets/subtensor/src/coinbase/run_coinbase.rs
```diff
@@ -78,7 +78,7 @@ impl<T: Config> Pallet<T> {
                 );
                 if let Ok(buy_swap_result_ok) = buy_swap_result {
                     let bought_alpha: AlphaCurrency = buy_swap_result_ok.amount_paid_out.into();
-                    Self::recycle_subnet_alpha(*netuid_i, bought_alpha);
+                    Self::distribute_alpha_to_subnet_owners(*netuid_i, bought_alpha);
```
ppolewicz, line 81: I suggest a global hyperparameter to allow root to choose between `0` for full recycle and `65535` for full distribution. We don't know yet how powerful this is going to be.

bdmason:
I suspect a lot of subnet owners will see the new alpha they are given as a bit of extra free income and sell it, which will mean chain buys go from price positive to negative sell pressure.

ppolewicz:
> I suspect a lot of subnet owners will see the new alpha they are given as a bit of extra free income and sell it, which will mean chain buys go from price positive to negative sell pressure.

Unless keeping these alphas will be better because the subnet emitting it is actually good and has a future. Then they will hold.

PR #2331, opened 2026-01-06
taoflow hotfix

sam0x17:
Root subnet stake moves now adjust root-claimed tracking: on ROOT transfers we remove root claim from the origin pair and add it to the destination pair to keep accounting aligned

### DIFF
--- a/pallets/subtensor/src/staking/stake_utils.rs
+++ b/pallets/subtensor/src/staking/stake_utils.rs
@@ -862,6 +862,9 @@ impl<T: Config> Pallet<T> {
             netuid,
             alpha,
         );
+		if netuid == NetUid::ROOT {
+			Self::remove_stake_adjust_root_claimed_for_hotkey_and_coldkey(origin_hotkey, origin_coldkey, alpha);
+		}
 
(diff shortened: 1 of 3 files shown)


ales-otf (approved):

pallets/subtensor/src/staking/stake_utils.rs
```diff
@@ -862,6 +862,13 @@ impl<T: Config> Pallet<T> {
             netuid,
             alpha,
         );
+        if netuid == NetUid::ROOT {
```
ales-otf, line 865: there is `NetUid::is_root` btw (so just `netuid.is_root()`)

camfairchild: This is what's live now, good point though

gztensor (approved):

JohnReedV (approved):

l0r1s:
try-runtime panics on finney?

sam0x17:
> try-runtime panics on finney?

yup just because spec version is unchanged

shamil-gadelshin (approved):

PR #2340, opened 2026-01-12
MevShield Emit Failure Event on Pre-Dispatch Failures

JohnReedV:
## Description
Currently the `MevShield` pallet will emit an event in the case that decryption of the inner extrinsic fails or if the inner extrinsic results in a runtime error. 

This PR adds logic to emit the failure event in the case that the inner extrinsic fails on a pre-dispatch check such as the inability to pay the fee.

**Requires** node upgrades to work.

### DIFF
--- a/node/src/mev_shield/proposer.rs
+++ b/node/src/mev_shield/proposer.rs
@@ -13,6 +13,18 @@ use std::{
     sync::{Arc, Mutex},
 };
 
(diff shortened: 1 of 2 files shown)


l0r1s (approved):

basfroman:
we care about btcli e2e tests already

ales-otf (approved):

PR #2353, opened 2026-01-16
Reenable alpha fees

### LINKED ISSUE #2336: Shared pool precision affects alpha tx fees
The shared pool precision is limited in a way a following example explains in detail. Simply put, if there are two coldkeys and one of them is trying to unstake partially, which is going to push beyond of what alpha shared pool can handle in terms of precision, the shared pool requests unstaking the full amount.

This is not a huge issue when removing stake: We just unstake all instead of unstaking partially, no alpha or tao is lost. But when paying transaction fees in alpha, we have no utility to return unused portion of unstaked tao mainly due to two reasons:

1. This operation is unstaking outside of transaction. Very hard to index/track, it will harm indexers (such as taostats), and will be confusing for the community
2. This kind of unstaking does not update tao flow and will be immediately exploitable

The fix is to use bigmath in alpha share pool to make it's precision virtually infinite.

**Precision example**

| Action | Shared pool numerator | Stake | Shared pool denominator | Total hotkey alpha |
|---|---|---|---|---|
| 1. Alice stakes 1 | Alice: 1 | Alice: 1 | 1 | 1 |
| 2. Bob stakes 1 | Alice: 1, Bob: 1 | Alice: 1, Bob: 1 | 2 | 2 |
| 3. Emission of 100 | Alice: 1, Bob: 1 | Alice: 51, Bob: 51 | 2 | 102 |
| 4. Alice unstakes 50 | Alice: 0, Bob: 1 | Alice: 0, Bob: 52 | 1 | 52 |

The problem appears on step 4, if we don't fully unstake Alice. One may ask why don't we use higher precision denominator to start with. Here is another example that shows that denominator value will never be enough:

| Action | Shared pool numerator | Stake | Shared pool denominator | Total hotkey alpha |
|---|---|---|---|---|
| 1. Alice stakes 1 | Alice: 1000 | Alice: 1 | 1000 | 1 |
| 2. Emission of 100 | Alice: 100 | Alice: 101 | 1000 | 101 |
| 3. Alice unstakes 100 | Alice: 1 | Alice: 1 | 10 | 1 |

Emissions are "eating" the denominator digits and in order to increase it back, we need to re-scale the complete share pool (all coldkeys in it), which is a O(n) operation.

The migration from Alpha to AlphaV2 map will be lazy due to the size of Alpha map. Related issue https://github.com/opentensor/subtensor/issues/2367 is to finish the migration and do the final cleanup.

**Action Plan**

- [x] Re-enable alpha fees
- [x] Implement real alpha -> tao swapping when charging fees in alpha
- [x] Add SafeFloat type for high precision floating point calculations
- [x] Refactor and re-implement share pool to use correct types (u64 for shared_value and SafeFloat for precision ops)
- [x] Test SafeFloat thoroughly
  - [x] adjust_precision
  - [x] add
  - [x] div
  - [x] mul_div
- [x] Add specific unit test for loss of 7 TAO in alpha fees
- [x] Lazy-Migrate Alpha -> AlphaV2 (with SafeFloat)
- [x] Lazy-Migrate TotalHotkeyShares -> TotalHotkeySharesV2 (with SafeFloat)
- [x] Handle overflows in SafeFloat (quietly ignore with error log: Expected to happen in 400B years in the worst case)
- [x] Remove `PrecisionLoss` error and related tests (the tests should fail before removal)
- [x] Initialize AlphaV2 and THSV2 in genesis
- [x] Use Alpha/THS V2 in
  - [x] deletage_info
  - [x] get_total_stake_for_coldkey
  - [x] get_total_stake_for_coldkey_on_subnet
  - [x] clear_small_nominations
  - [x] populate_root_coldkey_staking_maps
  - [x] destroy_alpha_in_out_stakes
  - [x] perform_swap_coldkey
  - [x] perform_hotkey_swap_on_all_subnets
  - [x] perform_hotkey_swap_on_one_subnet
- [x] Tests for share pool data operations (v1-v2 logic is non-trivial and needs to be tested)
- [ ] Address TODO: When balancer is merged...
- [ ] Deferred PR with residual Alpha and THS migration, with hard limit for the count of keys for safety

gztensor:
### Summary 

Allow paying transaction fees in alpha when no tao balance is available.
See issue #2336 for details.

### Implementation

The share pool is re-implemented using `SafeFloat` type. The `SafeFloat` type is a structure that stores `U256` mantissa and an `i64` exponent in a normalized form (`SAFE_FLOAT_MAX / 10 <= mantissa < SAFE_FLOAT_MAX`). It implements basic arithmetic that is needed for share pool operations: Comparisons (gt), addition, division, and mul_div operation that calculates `self * a / b` without loss of precision.

This PR allows the alpha to be unstaked precisely, no matter what the state of share pool is. There is no need to unstake the full amount in case of low precision anymore.

### User visible changes

- Unstaking with zero TAO balance becomes possible. 
- Issue #2336 is fixed.
- Users will experience higher precision staking operations.

## Related Issue(s)

- Closes #2336 

## Type of Change
- [ ] Documentation update

## Breaking Change

The `Alpha` and `TotalHotkeyShares` maps will be deprecated in favor of high precision `AlphaV2` and `TotalHotkeySharesV2`. The migration is lazy, so if any client must use `Alpha` and/or `TotalHotkeyShares` maps directly, it should read both maps and use a non-zero value, whichever is non-zero at the moment. `Alpha` and `TotalHotkeyShares` migrations are not synched and may occur separately for the same hotkey.

If a client uses `stakeInfoRuntimeApi` to get stake values for coldkey or hotkey-coldkey pairs, it will not break, this interface is supported and is not changed by this PR.

## Checklist


pallets/subtensor/src/staking/helpers.rs
```diff
@@ -413,7 +414,121 @@ impl<T: Config> Pallet<T> {
...
+            (key, sf)
+        });
+
+        // New Alpha shares format: SafeFloatSerializable -> SafeFloat
+        let v2 = AlphaV2::<T>::iter().map(|(key, val_sf_ser)| {
+            let sf: SafeFloat = SafeFloat::from(&val_sf_ser);
+            (key, sf)
+        });
+
+        legacy.chain(v2)
```
shamil-gadelshin: Deduplication, maybe?

pallets/subtensor/src/staking/helpers.rs
```diff
@@ -413,7 +414,121 @@ impl<T: Config> Pallet<T> {
         }
     }
 
+    // Same thing as populate_root_coldkey_staking_maps, but for AlphaV2
+    pub fn populate_root_coldkey_staking_maps_v2() {
+        // Get starting key for the batch. Get the first key if we restart the process.
```
shamil-gadelshin, line 421: Why do we need this?

pallets/subtensor/src/swap/swap_hotkey.rs
```diff
@@ -513,9 +534,42 @@ impl<T: Config> Pallet<T> {
...
+                    SafeFloatSerializable::from(&SafeFloat::from(alpha.saturating_add(new_alpha))),
+                );
+                weight.saturating_accrue(T::DbWeight::get().reads_writes(1, 2));
+
+                // Swap StakingHotkeys.
+                // StakingHotkeys( coldkey ) --> Vec<hotkey> -- the hotkeys that the coldkey stakes.
+                let mut staking_hotkeys = StakingHotkeys::<T>::get(&coldkey);
+                weight.saturating_accrue(T::DbWeight::get().reads(1));
+                if staking_hotkeys.contains(old_hotkey) && !staking_hotkeys.contains(new_hotkey) {
+                    staking_hotkeys.push(new_hotkey.clone());
```
shamil-gadelshin: Shouldn't we remove the old hotkey here?

shamil-gadelshin (changes requested):
Great work!

A couple of comments:
- What is the reason we copy zstd-* libs in our repo? I would move zstd-libs into separate crates and reference them like our `bls` patch.
- We need to merge #2251 to integrate "swap coldkey changes" first
- Consider hiding Alpha map access behind some "temporary interface". It'll allow implementing Alphav2 logic easily. We would remove the interface after the migration period.  
- I don't see effects on the root claim (it should work as expected)

pallets/subtensor/src/swap/swap_hotkey.rs
gztensor: Not here because coldkey may stake on the same hotkey in some other subnet. This removal is done in `perform_hotkey_swap_on_all_subnets`.

pallets/subtensor/src/staking/helpers.rs
gztensor: Added deduplication. Returning v2 on duplicates, not a sum, because normal (by design) logic of lazy migration should not allow the keys with non-zero values to simultaneously exist in both maps, so preferring one of them is safer.

sam0x17:
> * What is the reason we copy zstd-* libs in our repo? I would move zstd-libs into separate crates and reference them like our `bls` patch.

Yes, I ended up doing this because the upstream zstd crate is broken with respect to no_std when using the particular target we use. I agree we should fork it

pallets/subtensor/src/staking/helpers.rs
gztensor: In case if slow migration doesn't finish by the time this PR is on mainnet, this function may get used. Leaving in for now, but will mark it with a "TODO" to remove later.

JohnReedV (approved):

basfroman:
Hey @gztensor, update from Cortex:
- SDK staging is up to date with this PR
- btcli will be updated on Monday

sam0x17 (approved):

pallets/subtensor/src/macros/events.rs
```diff
@@ -528,5 +528,18 @@ mod events {
             /// Alpha burned
             alpha: AlphaBalance,
         },
+
+        /// Transaction fee was paid in Alpha.
+        ///
+        /// Emitted in addition to `TransactionFeePaid` when the fee payment path is Alpha.
+        /// `alpha_fee` is the exact Alpha amount deducted.
+        TransactionFeePaidWithAlpha {
```
camfairchild, line 536: needs netuid

PR #2354, opened 2026-01-16
Check node compat workflow

lorismoulin:
A simple workflow that will run a small network of 3 nodes with 2 validators on the "check-node-compat" label and ensure the network progress and finalize for 50 blocks, ensuring devnet-ready node and the branch the label is added on have compatible nodes.


shamil-gadelshin (approved):

open-junius (approved):
Look good, will try the label when runtime update.

PR #2320, opened 2025-12-24
enable multi block per slot

open-junius:
## Description
When use the forge to deploy contract, I noticed there is a panic in runtime. the source code in pallet aura.
the error message is "Slot must increase".  source code as follows:

error message
```bash
user@user-X870-EAGLE-WIFI7:~/github/tmp/collateral-contracts$ ./deploy.sh 
[⠊] Compiling...
No files changed, compilation skipped
Error: server returned an error response: error code -32603: Create pending runtime api error: Failed to call runtime API, Execution failed: Execution aborted due to trap: wasm trap: wasm `unreachable` instruction executed
WASM backtrace:
error while executing at wasm backtrace:
    0: 0x70a480 - node_subtensor_runtime.wasm!__rustc[5224e6b81cd82a8f]::rust_begin_unwind
    1: 0x29b6c - node_subtensor_runtime.wasm!core::panicking::panic_fmt::h4f1a54382be6d50f
    2: 0x2ea43f - node_subtensor_runtime.wasm!<pallet_aura::pallet::Pallet<T> as frame_support::traits::hooks::OnInitialize<<<<T as frame_system::pallet::Config>::Block as sp_runtime::traits::Block>::Header as sp_runtime::traits::Header>::Number>>::on_initialize::hb1215f7411744c05
    3: 0x583a3d - node_subtensor_runtime.wasm!<(TupleElement0,TupleElement1,TupleElement2,TupleElement3,TupleElement4,TupleElement5,TupleElement6,TupleElement7,TupleElement8,TupleElement9,TupleElement10,TupleElement11,TupleElement12,TupleElement13,TupleElement14,TupleElement15,TupleElement16,TupleElement17,TupleElement18,TupleElement19,TupleElement20,TupleElement21,TupleElement22,TupleElement23,TupleElement24,TupleElement25,TupleElement26) as frame_support::traits::hooks::OnInitialize<BlockNumber>>::on_initialize::h13ce7ae7c448e363
    4: 0x61154f - node_subtensor_runtime.wasm!frame_executive::Executive<System,Block,Context,UnsignedValidator,AllPalletsWithSystem,COnRuntimeUpgrade>::initialize_block::h02a0e55757ce5f00
    5: 0x417fc2 - node_subtensor_runtime.wasm!Core_initialize_block
```

pallet aura function
```rust
fn on_initialize(_: BlockNumberFor<T>) -> Weight {
			if let Some(new_slot) = Self::current_slot_from_digests() {
[...]

PR #2368, opened 2026-01-23
Restore pending_create_data_providers and fix EVM pending block issues

shamil-gadelshin:
This PR fixes Frontier  "pending block" feature: eth_getBalance request for pending block, estimate gas feature, etc.

After merging the "NPOS" preparation PR (#1941), we united `create_inherent_data_providers` and `pending_create_data_providers` (which produce different results - one slot difference for pending data provider), and chose `pending_create_data_providers` for both aura consensus and frontier EthDeps parameter. Which, in turn, advanced our chain for one block slot in the future. The future slot is not an issue when a shift occurs every block, however, when we need to create a pending block (one slot in the future), we use the same slot, and our Aura consensus breaks with "slot must increase" assertion. 

This PR restores the correct inherent providers for both consensus cases (aura and frontier), it also adds providers for the future BABE consensus.

Relates to #2320

I tested it manually on local machine and a baedeker clone. This PR must be tested on the testnet before deployment because it changes the consensus logic.

### DIFF
--- a/node/src/consensus/aura_consensus.rs
+++ b/node/src/consensus/aura_consensus.rs
@@ -102,6 +102,18 @@ impl ConsensusMechanism for AuraConsensus {
     fn create_inherent_data_providers(
         slot_duration: SlotDuration,
     ) -> Result<Self::InherentDataProviders, Box<dyn Error + Send + Sync>> {
+        let timestamp = sp_timestamp::InherentDataProvider::from_system_time();
+        let slot =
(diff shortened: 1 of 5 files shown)


open-junius (dismissed):
It looks good to me. I tested it in local environment, querying with "pending" parameter is OK. 
I tried the Foundry tool, the contract deployment is OK now.

liamaharon (dismissed):
Nice catch!

sam0x17 (approved):

sam0x17 (approved):

PR #2381, opened 2026-01-29
Subnet buyback

shamil-gadelshin:
This PR introduces a new extrinsic - `subnet_buyback`. It's a combination of `add_stake` or `add_stake_limit` (depending on the provided price limit) and `burn_alpha` extrinsics. Only the subnet owner can call this extrinsic and only once per tempo. This extrinsic has a higher priority than regular extrinsics but a lower priority than system(sudo) extrinsics or drand pulses.


pallets/subtensor/src/staking/recycle_alpha.rs
```diff
@@ -134,6 +134,36 @@ impl<T: Config> Pallet<T> {
...
+        let alpha = if let Some(limit) = limit {
+            Self::do_add_stake_limit(origin.clone(), hotkey.clone(), netuid, amount, limit, false)?
+        } else {
+            Self::do_add_stake(origin.clone(), hotkey.clone(), netuid, amount)?
+        };
+
+        Self::do_burn_alpha(origin, hotkey, alpha, netuid)?;
+
+        Self::set_rate_limited_last_block(&RateLimitKey::SubnetBuyback(netuid), current_block);
+
```
lorismoulin, line 166: Would it make sense to have an additional event to the stake/burn events to indicate a buyback occurred? Also, how about storing how much buyback has been done by a SN owner as an additional metric?

pallets/subtensor/src/macros/dispatches.rs
```diff
@@ -2487,5 +2488,24 @@ mod dispatches {
             Self::deposit_event(Event::ColdkeySwapReset { who: coldkey });
             Ok(())
         }
+
+        /// --- Subnet buyback
```
lorismoulin: Doc could be a bit more detailed?

l0r1s (dismissed):

basfroman:
Let's be consistent with the naming conventions. Currently, this is a truly controversial name.
The name of the extrinsic function should intuitively indicate its purpose and which group it belongs to. Currently, this is not the case.
Otherwise, Cortex will soon be ready for this functionality.

pallets/subtensor/src/staking/recycle_alpha.rs
shamil-gadelshin: I added the event for now. If an event index won't be enough, we'll add the metric.

l0r1s (dismissed):

gztensor (approved):

sam0x17 (approved):

PR #2383, opened 2026-01-29
Feat: Proxy aware chain extensions

LandynDev:
## Description
 
Adds 12 proxy-delegated chain extension variants that allow a smart contract with proxy rights to manage on behalf of a proxied coldkey. Each handler accepts an `on_behalf_of: AccountId` parameter, verifies the ProxyType and dispatches the operation as the proxied account. `ProxyType::Any` is accepted as a superset of all types.

## Related Issue
Closes https://github.com/opentensor/subtensor/issues/2566

## Type of Change

- [x] Documentation update

## Breaking Change

This PR does not create a breaking change.
This introduces new proxy chain extensions to wrap runtime calls with correct proxy perms.

## Checklist

## Screenshots
**Unit Tests:**
image

## Additional Notes

### DIFF
--- a/chain-extensions/src/lib.rs
+++ b/chain-extensions/src/lib.rs
@@ -16,6 +16,7 @@ use pallet_contracts::chain_extension::{
 };
 use pallet_subtensor_proxy as pallet_proxy;
 use pallet_subtensor_proxy::WeightInfo;
+use sp_runtime::traits::Zero;
 use sp_runtime::{DispatchError, Weight, traits::StaticLookup};
 use sp_std::marker::PhantomData;
 use substrate_fixed::types::U96F32;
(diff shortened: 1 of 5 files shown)


chain-extensions/src/lib.rs
```diff
@@ -530,6 +530,529 @@ where
...
+                ) = env
+                    .read_as()
+                    .map_err(|_| DispatchError::Other("Failed to decode input parameters"))?;
+
+                let caller = env.caller();
+                if real_coldkey != caller {
+                    pallet_proxy::Pallet::<T>::find_proxy(
+                        &real_coldkey,
+                        &caller,
+                        Some(ProxyType::Staking),
```
open-junius: Should we check other ProxyType like Any also enable the add_stake call.

open-junius:
All V2 extensions are not the upgrade or replace of V1. Could use the name like proxy_xxx.
Do you consider use the unique extension to implement them all. like proxy_execute(real_coldkey: AccountId, call: RuntimeCall). like in EVM precompile https://github.com/opentensor/subtensor/blob/main/precompiles/src/proxy.rs#L146.

LandynDev:
> All V2 extensions are not the upgrade or replace of V1. Could use the name like proxy_xxx. Do you consider use the unique extension to implement them all. like proxy_execute(real_coldkey: AccountId, call: RuntimeCall). like in EVM precompile https://github.com/opentensor/subtensor/blob/main/precompiles/src/proxy.rs#L146.

Went ahead and went this direction, was simplest.
Updated PR descr. with added testing from these simplified changes
Thanks for the review

LandynDev:
> All V2 extensions are not the upgrade or replace of V1. Could use the name like proxy_xxx. Do you consider use the unique extension to implement them all. like proxy_execute(real_coldkey: AccountId, call: RuntimeCall). like in EVM precompile https://github.com/opentensor/subtensor/blob/main/precompiles/src/proxy.rs#L146.

I went ahead and implemented this with `Proxy*` variants (e.g. `ProxyAddStakeV1`, `ProxyRemoveStakeV1`, etc.).

Regarding a single proxy_execute(real_coldkey, call: RuntimeCall) — I actually started with that approach (a generic proxy_call dispatcher that took SCALE-encoded RuntimeCall bytes). The problem is that ink! contract developers would have to manually SCALE-encode the target function selector and all arguments into raw bytes. Found it to be error-prone and a poor developer experience. 

I pretty much just implemented it how the `Caller*` variants were done in #2550.

PR #2385, opened 2026-01-31
Taoflow2

ppolewicz:
## Description

This PR introduces mechanisms to control which subnets receive TAO emission and how root dividends are distributed based on validator behavior.

### New metric: EffectiveRootProp

Each subnet distributes dividends to two groups: alpha stakers (staked directly on subnet) and root stakers (staked on root network). `EffectiveRootProp` measures what fraction flows to root stakers, scaled by how efficiently root validators are actually earning dividends:

```py
raw_root_prop = root_dividends / (alpha_dividends + root_dividends)
EffectiveRootProp[netuid] = raw_root_prop * utilization
```

### Root stake utilization (dividend efficiency)

Utilization measures how well root validators are earning their expected share of dividends on each subnet. For each root-staked validator registered on the subnet:

```py
expected_share = root_stake_i / total_root_stake_on_subnet
actual_share   = root_dividends_i / total_root_dividends
efficiency_i   = min(actual_share / expected_share, 1.0)
utilization    = sum(root_stake_i * efficiency_i) / total_root_stake_on_subnet
```

Only root stake of validators with UIDs registered on the subnet counts — TAO which is staked to a validator which is not registered on the subnet as well as unstaked TAO are irrelevant.

### Hard cap at 50% utilization

If a subnet's utilization drops below 50%, **all** root dividends for that subnet are recycled (returned to the subnet's alpha pool) and `EffectiveRootProp` is set to 0. This is a binary gate: subnets at or above 50% utilization receive full root dividends, subnets below 50% receive none.

### Emission filters

Three emission filters are applied sequentially in `get_subnet_block_emissions()`:

1. **EffectiveRootProp scaling** (disabled by default): Multiplies each subnet's emission share by `min(EffectiveRootProp, RootProp)` and re-normalizes. Using the minimum prevents exploitation by disabling alpha validators to artificially inflate `EffectiveRootProp` above `RootProp`.
2. **Top subnet proportion**: Keeps only the top X% of subnets by share (default 100% = all subnets). Uses `ceil()` rounding so a single subnet always counts as in top 50%.
3. **Absolute subnet limit**: Hard cap on number of subnets receiving emission (default: disabled).

After each filter, shares are re-normalized to sum to 1.0. Subnets that get filtered out receive zero emission.

### But why?

We had a couple of problems since Taoflow emission schedule was deployed:
- New subnets get very little injection of Tao when they desperately need it
- Old subnets which have a lot of liquidity get even more
- Since dtao subnets are generally incentivized to fight root prop (see sn89 using tricks to reduce root prop artificially) to reduce sell pressure
- sn104 managed to cut off root validators from receiving any dividends at all by using the time gap between CHK application and immediate effectiveness of `btcli stake move` coupled with randomly chosen "miner" (they are self-mining, the subnet doesn't really do anything). We think they can safely inflow to increase emissions and we want to mitigate this risk.
- sometimes subnet owners say "oh, the validator code doesn't work for you? Then CHK stake to me, I don't want to debug your environment problems" and it is unacceptable, so it'd be good to incentivize them to care that the validator code actually works

The dividend-efficiency utilization and hard cap directly address these: subnets where root validators can't earn dividends (like sn104) get their root emission recycled, while subnets with healthy validator participation receive full emission.

## Type of Change

## Checklist

## AI reviewers
- Claude 4.6 Opus
- Codex 5.3-high
- Devin


pallets/subtensor/src/coinbase/subnet_emissions.rs
```diff
@@ -22,14 +22,157 @@ impl<T: Config> Pallet<T> {
...
+
+        // ceil(total * proportion / 10000)
+        let top_k = ((total as u64) * (proportion as u64)).div_ceil(10000);
+        let top_k = top_k.max(1) as usize; // At least 1 subnet
+
+        log::debug!(
+            "EmissionTopSubnetProportion: keeping top {top_k} of {total} subnets (proportion: {proportion}/10000)"
+        );
+
+        Self::zero_and_redistribute_bottom_shares(shares, top_k);
```
ppolewicz: We don't want to call it multiple times when `apply_effective_root_prop_scaling`, `apply_top_subnet_proportion_filter`, `apply_top_subnet_absolute_limit` are all enabled. These functions should return a boolean and then in `get_subnet_block_emissions` we should do `if effective_root_prop_scaling_was_applied || top_subnet_absolute_limit_was_applied || top_subnet_proportion_filter_was_applied` - and then we call `zero_and_redistribute_bottom_shares` one time.

pallets/admin-utils/src/lib.rs
```diff
@@ -2259,6 +2259,67 @@ pub mod pallet {
...
+            DispatchClass::Operational,
+            Pays::Yes
+        ))]
+        pub fn sudo_set_effective_root_prop_emission_scaling(
+            origin: OriginFor<T>,
+            enabled: bool,
+        ) -> DispatchResult {
+            ensure_root(origin)?;
+            pallet_subtensor::Pallet::<T>::set_effective_root_prop_emission_scaling(enabled);
+            log::debug!("set_effective_root_prop_emission_scaling( {enabled:?} )");
```
shamil-gadelshin: This is noise. Please, remove unnecessary logs.

pallets/admin-utils/src/lib.rs
```diff
@@ -2259,6 +2259,67 @@ pub mod pallet {
...
+                .saturating_add(<T as frame_system::Config>::DbWeight::get().writes(1)),
+            DispatchClass::Operational,
+            Pays::Yes
+        ))]
+        pub fn sudo_set_emission_top_subnet_proportion(
+            origin: OriginFor<T>,
+            proportion: u16,
+        ) -> DispatchResult {
+            ensure_root(origin)?;
+            ensure!(
```
shamil-gadelshin: Why do we put these limits for the proportion?

pallets/admin-utils/src/lib.rs
```diff
@@ -2259,6 +2259,67 @@ pub mod pallet {
...
+                proportion > 0 && proportion <= 10000,
```
shamil-gadelshin: The proportion type is counterintuitive. Consider using U64F64 and limit "0-1" or "0-100".

pallets/subtensor/src/coinbase/run_coinbase.rs
```diff
@@ -638,6 +645,41 @@ impl<T: Config> Pallet<T> {
...
+            .fold(zero, |acc, v| acc.saturating_add(*v));
+
+        let total_root_divs: U96F32 = root_alpha_dividends
+            .values()
+            .fold(zero, |acc, v| acc.saturating_add(*v));
+
+        let total = total_alpha_divs.saturating_add(total_root_divs);
+
+        let effective_root_prop = if total > zero {
+            U64F64::saturating_from_num(total_root_divs.checked_div(total).unwrap_or(zero))
```
shamil-gadelshin: Do you need these conversions between U64F64 and u96F32? Can we use a single format?

pallets/subtensor/src/coinbase/subnet_emissions.rs
```diff
@@ -22,14 +22,139 @@ impl<T: Config> Pallet<T> {
...
+    }
+
+    /// Limits the number of subnets receiving emission to an absolute number.
+    /// When limit is 0, no filtering occurs (disabled).
+    /// When limit > 0 and less than the number of subnets with nonzero shares,
+    /// zeros shares beyond the top `limit` subnets and re-normalizes.
+    pub(crate) fn apply_top_subnet_absolute_limit(shares: &mut BTreeMap<NetUid, U64F64>) {
+        let limit = EmissionTopSubnetAbsoluteLimit::<T>::get();
+        if limit == 0 {
```
shamil-gadelshin: Please, use rust idiomatic None for this case.

pallets/subtensor/src/coinbase/subnet_emissions.rs
```diff
@@ -22,14 +22,139 @@ impl<T: Config> Pallet<T> {
...
+    }
+
+    /// Zeros shares outside top_k (by descending share value) and re-normalizes the rest.
+    /// Subnets with equal shares at the boundary are included if they tie with the k-th position.
+    pub(crate) fn zero_and_redistribute_bottom_shares(
+        shares: &mut BTreeMap<NetUid, U64F64>,
+        top_k: usize,
+    ) {
+        let zero = U64F64::saturating_from_num(0);
+        if top_k == 0 || shares.is_empty() {
```
shamil-gadelshin: why do we iterate through the empty array?

pallets/subtensor/src/coinbase/subnet_emissions.rs
```diff
@@ -22,14 +22,139 @@ impl<T: Config> Pallet<T> {
...
+    /// Filters subnets so only the top proportion (by share) receive emission.
+    /// Uses ceil(count * proportion / 10000) to determine how many subnets to keep.
+    /// A single subnet always counts as in top 50%.
+    pub(crate) fn apply_top_subnet_proportion_filter(shares: &mut BTreeMap<NetUid, U64F64>) {
+        let proportion = EmissionTopSubnetProportion::<T>::get();
+        if proportion >= 10000 {
+            return; // 100% means all subnets get emission
+        }
+
+        let total = shares.len() as u32;
```
shamil-gadelshin: Incorrect type conversion - there is a different conversion below.

pallets/subtensor/src/coinbase/run_coinbase.rs
```diff
@@ -508,6 +508,13 @@ impl<T: Config> Pallet<T> {
         alpha_dividends: BTreeMap<T::AccountId, U96F32>,
         root_alpha_dividends: BTreeMap<T::AccountId, U96F32>,
     ) {
+        // Compute and store EffectiveRootProp before distributing (uses raw dividend values).
+        Self::compute_and_store_effective_root_prop(
```
shamil-gadelshin: It seems we calculate the effective root prop for the next round. Is it intentional?

pallets/subtensor/src/coinbase/subnet_emissions.rs
```diff
@@ -22,14 +22,139 @@ impl<T: Config> Pallet<T> {
...
         subnets_to_emit_to: &[NetUid],
         block_emission: U96F32,
     ) -> BTreeMap<NetUid, U96F32> {
         // Get subnet TAO emissions.
-        let shares = Self::get_shares(subnets_to_emit_to);
+        let mut shares = Self::get_shares(subnets_to_emit_to);
         log::debug!("Subnet emission shares = {shares:?}");
 
+        // Apply EffectiveRootProp scaling if enabled.
+        Self::apply_effective_root_prop_scaling(&mut shares);
```
shamil-gadelshin, line 158: The first filter affects the other two. Do we want this exact order?

pallets/subtensor/src/tests/subnet_emissions.rs
```diff
@@ -488,3 +488,599 @@ fn seed_price_and_flow(n1: NetUid, n2: NetUid, price1: f64, price2: f64, flow1:
 //         assert_abs_diff_eq!(s1 + s2 + s3, 1.0, epsilon = 1e-9);
 //     });
 // }
+
```
shamil-gadelshin, line 504: There are no tests to verify that the new taoflow actually works as a whole.

pallets/subtensor/src/coinbase/run_coinbase.rs
shamil-gadelshin: Also, we calculate this value once per epoch. Could this fact be exploited?

shamil-gadelshin (changes requested):
The tests don't cover the whole `run_coinbase`. There are some concerns about the calculation order and other minor issues.

pallets/subtensor/src/coinbase/run_coinbase.rs
shamil-gadelshin: We could probably place a proper stake before we calculate the value and then remove it.

pallets/subtensor/src/coinbase/subnet_emissions.rs
ppolewicz: Yes, this is very much intended

ppolewicz:
Closing: rebasing taoflow2 onto origin/testnet instead of devnet-ready.

PR #2412, opened 2026-02-10
Improve dissolve network process

### LINKED ISSUE #2411: finalize_all_subnet_root_dividends has unbounded iteration over all hotkeys
In `pallets/subtensor/src/staking/claim_root.rs`, the `finalize_all_subnet_root_dividends` function collects ALL hotkeys from `RootClaimable` storage into memory, then mutates each one:

```rust
pub fn finalize_all_subnet_root_dividends(netuid: NetUid) {
    let hotkeys = RootClaimable::<T>::iter_keys().collect::<Vec<_>>();
    for hotkey in hotkeys.iter() {
        RootClaimable::<T>::mutate(hotkey, |claimable| {
            claimable.remove(&netuid);
        });
    }
    let _ = RootClaimed::<T>::clear_prefix((netuid,), u32::MAX, None);
}
```

This is O(N) reads + O(N) writes where N is the total number of hotkeys with any root claimable balance. This function is called from `do_dissolve_network`, which is an extrinsic. With thousands of hotkeys, this could:
- Exceed block weight limits
- Cause high memory usage from the `collect::<Vec<_>>()`

The weight annotation on `do_dissolve_network` may not account for this unbounded iteration.

**Suggested fix:** Consider using `RootClaimable::iter_keys()` without collecting, or implementing a bounded/paginated cleanup, or using a different storage layout that allows per-netuid cleanup without iterating all hotkeys.

open-junius:
## Description
The PR will fix the #2411, and optimize the process of dissolve a network.
Instead of remove all storages related to the netuid once, it will use the on_idle hook to remove these step by step.
and take the weights into consideration. 

- Too avoid repeated code, introduce 2 macro to handle the weight for single value and double storage.
- calculate weight for each iteration, if weight is not enough to continue, clean up will stop there. The checkpoint is recorded as prefix, then cleanup can continue in on_idle next block.
- Now, the cleanup can handle any size of data. No risk of hanging at dissolve network extrinsic.
- It includes some indent fix, not relevant to the functionality. but worth to fix them.

## Related Issue(s)

- Closes #2411 

## Type of Change

- [ ] Documentation update

## Breaking Change

## Checklist

## Additional Notes


pallets/subtensor/src/macros/hooks.rs
```diff
@@ -216,5 +221,43 @@ mod hooks {
...
+                    T::SwapInterface::clear_protocol_liquidity(*netuid, remaining_weight);
+                remaining_weight = remaining_weight.saturating_sub(weight_used);
+
+                let weight_used = T::CommitmentsInterface::purge_netuid(*netuid, remaining_weight);
+                remaining_weight = remaining_weight.saturating_sub(weight_used);
+
+                let weight_used = Self::remove_network(*netuid, remaining_weight);
+                remaining_weight = remaining_weight.saturating_sub(weight_used);
+
+                DissolvedNetworks::<T>::mutate(|networks| networks.retain(|n| *n != *netuid));
```
evgeny-s, line 279: Since we use `WeightMeterWrapper` that does early return, we might have a situation where there is not enough weight to finish some particular operation. As a result, we will have partially cleaned data, but the removal will be marked as finalized. It would be great to simulate it in the test to check if this scenario is possible.

pallets/subtensor/src/macros/hooks.rs
```diff
@@ -216,5 +221,43 @@ mod hooks {
...
+        // # Returns:
+        // 	* 'Weight': The weight remaining after the function.
+        //
+        fn remove_data_for_dissolved_networks(remaining_weight: Weight) -> Weight {
+            let mut remaining_weight = remaining_weight;
+            let dissolved_networks = DissolvedNetworks::<T>::get();
+
+            for netuid in dissolved_networks.iter() {
+                let weight_used =
+                    Self::finalize_all_subnet_root_dividends(*netuid, remaining_weight);
```
evgeny-s, line 245: Seems like we missed `dissolve_all_liquidity_providers`, is it intentional?

pallets/subtensor/src/staking/claim_root.rs
```diff
@@ -131,6 +132,10 @@ impl<T: Config> Pallet<T> {
         root_claim_type: RootClaimTypeEnum,
         ignore_minimum_condition: bool,
     ) {
+        if DissolvedNetworks::<T>::get().contains(&netuid) {
```
evgeny-s, line 135: Since we have a gap between the network that is marked for deletion and the actual deletion, shouldn't we add this check at all the other places? staking, etc.

open-junius: Usually we use the NetworksAdded or if_subnet_exist to check if the subnet is deleted. Maybe use the if_subnet_exist is better, I will check.

pallets/subtensor/src/macros/hooks.rs
open-junius: I will add such test case.

pallets/subtensor/src/macros/hooks.rs
open-junius: It should be included. I mistake it is only called by disable_lp.

sam0x17:
some merge conflicts

pallets/subtensor/src/staking/remove_stake.rs
```diff
@@ -501,33 +579,137 @@ impl<T: Config> Pallet<T> {
...
+                continue;
+            }
+            hotkeys_in_subnet.push(hot.clone());
+
+            let mut iterate_all = true;
+
+            for (cold, this_netuid, share_u64f64) in Self::alpha_iter_single_prefix(&hot) {
+                if !weight_meter.can_consume(r.saturating_mul(2_u64)) {
+                    iterate_all = false;
+                    LastKeptRawKey::<T>::set(Some(TotalHotkeyAlpha::<T>::hashed_key_for(
```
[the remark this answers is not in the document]
open-junius: Current solution is that the function handle all coldkeys for one hotkey if enough weights. otherwise, skip the hotkey, wait for next time with much more remaining weights. Some coldkeys could be scanned twice, but not added into stakers.

pallets/subtensor/src/lib.rs
```diff
@@ -2106,6 +2163,30 @@ pub mod pallet {
...
+    /// --- ITEM ( dissolved_subnet_total_alpha_value ) Total alpha value for the dissolved subnet.
+    /// It is only used during clean the data for dissolved networks.
+    #[pallet::storage]
+    pub type DissolvedSubnetTotalAlphaValue<T> = StorageValue<_, u128, OptionQuery>;
+
+    /// --- ITEM ( dissolved_subnet_settled_alpha_value ) Settled alpha value for the dissolved subnet.
+    /// It is only used during clean the data for dissolved networks.
+    #[pallet::storage]
+    pub type DissolvedSubnetDistributedTao<T> = StorageValue<_, u128, OptionQuery>;
+
```
lorismoulin: The naming of `DissolvedNetworks` feels a bit misleading here. At this point the network has been removed from the active set, but dissolution cleanup is still in progress, so a name like `DissolveCleanupQueue` would make the intent clearer.

Also, this introduces several global storage values that seem to represent one logical cleanup state machine. Could these be collapsed into a single `DissolveCleanupState` storage value?

For example:

```rust
pub struct PendingDissolveCleanupState {
    netuid: NetUid,
    phase: DissolveCleanupPhase,
}
```

Where each phase owns only the state it needs:

```rust
pub enum DissolveCleanupPhase {
    CleanRootClaimable {
        last_key: Option<Vec<u8>>,
    },
    CleanRootClaimed,
    ComputeTotalAlpha {
        last_key: Option<Vec<u8>>,
        total_alpha_value: u128,
    },
    SettleStakes {
        last_key: Option<Vec<u8>>,
        total_alpha_value: u128,
        distributed_tao: u128,
    },
    CleanAlpha {
        last_key: Option<Vec<u8>>,
    },
    // ...
}
```

That would avoid having global storage whose meaning depends on the current phase and the head of the queue.

pallets/subtensor/Cargo.toml
```diff
@@ -58,6 +58,7 @@ pallet-crowdloan.workspace = true
 pallet-subtensor-proxy.workspace = true
 pallet-shield.workspace = true
 pallet-scheduler.workspace = true
+sp-debug-derive = {workspace = true, features = ["force-debug"]}
```
lorismoulin: I don't think this should be present, only for debug purpose locally because it will prevent the `<wasm:stripped>` that are needed for production runtime to reduce WASM blob size.

pallets/subtensor/src/macros/dispatches.rs
```diff
@@ -1231,9 +1231,9 @@ mod dispatches {
         /// Remove a user's subnetwork
         /// The caller must be the owner of the network
         #[pallet::call_index(61)]
-        #[pallet::weight(Weight::from_parts(119_000_000, 0)
-		.saturating_add(T::DbWeight::get().reads(6))
-		.saturating_add(T::DbWeight::get().writes(31)))]
+        #[pallet::weight(Weight::from_parts(28_560_000, 0)
+		.saturating_add(T::DbWeight::get().reads(3))
+		.saturating_add(T::DbWeight::get().writes(3)))]
```
lorismoulin: This should be using the `<T as crate::pallet::Config>::WeightInfo::dissolve_network())` directly instead of inlining the weights, this is the correct way to do that and what we now use in the CI.

You can add the `dissolve_network()` function to the `pallets/subtensor/src/weights.rs` manually with Weight 0 and it should be overwritten by CI.

pallets/subtensor/src/macros/events.rs
```diff
@@ -532,6 +532,12 @@ mod events {
             alpha: AlphaBalance,
         },
 
+        /// data for a dissolved network has been cleaned up.
+        DissolvedNetworkDataCleaned {
+            /// The subnet ID
+            netuid: NetUid,
+        },
```
lorismoulin, line 541: Could we rename this event to something like `DissolveCleanupCompleted` or `NetworkDissolveCleanupCompleted`? `DissolvedNetworkDataCleaned` is a bit confusing, and what is completing here is the async cleanup phase after the network has already been removed from the active set.

pallets/subtensor/src/macros/hooks.rs
```diff
@@ -1,5 +1,6 @@
 use frame_support::pallet_macros::pallet_section;
-
+// use subtensor_commitments_interface::CommitmentsHandler;
+// use subtensor_swap_interface::SwapHandler;
```
lorismoulin:
```suggestion
```

pallets/subtensor/src/macros/hooks.rs
```diff
@@ -220,5 +229,327 @@ mod hooks {
...
+        // # Args:
+        // 	* 'remaining_weight': (Weight):
+        // 		- The weight available for this cleanup step.
+        // 	* 'netuid': (&NetUid):
+        // 		- The subnet to clean dissolved-network data for.
+        //
+        // # Returns:
+        // 	* 'Weight': The weight used for the cleanup step.
+        //
+        fn remove_data_for_dissolved_networks(remaining_weight: Weight, netuid: &NetUid) -> Weight {
```
lorismoulin: Could we move `remove_data_for_dissolved_networks` out of `hooks.rs` into a dissolution focused module, e.g. `subnets/dissolution.rs`?

The hook itself only needs to call the cleanup runner, while this function is the main async dissolution state machine. Keeping the state/phase types and cleanup transition logic together would make this much easier to review and maintain.

pallets/subtensor/src/coinbase/root.rs
```diff
@@ -210,72 +209,195 @@ impl<T: Config> Pallet<T> {
...
         );
 
-        Self::finalize_all_subnet_root_dividends(netuid);
+        let mut dissolved_networks = DissolvedNetworks::<T>::get();
+        ensure!(
+            !dissolved_networks.contains(&netuid),
+            Error::<T>::NetworkAlreadyDissolved
+        );
+
+        // TODO Most of data cleanup is done in the block hook, should we charge the user for this?
```
lorismoulin: I don’t think this TODO is needed. Since the heavy cleanup is now performed asynchronously from `on_idle`, this feels like protocol/chain maintenance rather than work that should be directly charged to the dissolve caller. The extrinsic should pay for marking the subnet inactive / queueing cleanup; the later bounded cleanup is part of block production.

common/src/lib.rs
```diff
@@ -523,12 +525,126 @@ impl TypeInfo for NetUidStorageIndex {
...
 
+#[macro_export]
+macro_rules! WeightMeterWrapper {
+    ( $meter:expr, $weight:expr ) => {{
+        if !$meter.can_consume($weight.clone()) {
+            return false;
+        }
+        $meter.consume($weight.clone());
+    }};
+}
```
lorismoulin: I don’t think this abstraction is buying much. At the call sites, spelling out `can_consume` / `consume` would be clearer than a macro/helper, especially because the important behavior is the early `return false` when the cleanup runs out of weight. This is only two lines and keeping it explicit makes the phase logic easier to audit.

common/src/lib.rs
```diff
@@ -523,12 +525,126 @@ impl TypeInfo for NetUidStorageIndex {
...
+        }
+
+        let result: $crate::MultiRemovalResults = <$storage>::clear_prefix($netuid, limit, None);
+        $meter.consume(weight.saturating_mul(result.backend.into()));
+
+        if !result.maybe_cursor.is_none() {
+            return false;
+        }
+    }};
+}
```
lorismoulin: I’d prefer not to introduce this as an exported macro. It hides ordinary control flow (`return false`) inside the macro, which makes the cleanup paths harder to audit.

It also looks more generic than the current use case needs. Every call site passes `T::DbWeight::get().writes(1)`, so the macro does not need an arbitrary `$weight` parameter or a zero-weight branch that maps to `u32::MAX`. For this cleanup path, a prefix entry deletion is always accounted as one DB write.

This could be a small local helper using a closure instead:

```rust
fn clear_prefix_with_meter<T: frame_system::Config>(
    meter: &mut WeightMeter,
    clear_prefix: impl FnOnce(u32) -> MultiRemovalResults,
) -> bool {
    let per_item = T::DbWeight::get().writes(1);
    let Some(limit) = meter
        .remaining()
        .checked_div_per_component(&per_item)
        .and_then(|limit| u32::try_from(limit).ok())
    else {
        return false;
    };

    if limit == 0 {
        return false;
    }

    let result = clear_prefix(limit);
    meter.consume(per_item.saturating_mul(result.unique.max(result.loops).into()));

    result.maybe_cursor.is_none()
}
```

pallets/swap/src/pallet/mod.rs
```diff
@@ -118,10 +130,6 @@ mod pallet {
     pub type HasMigrationRun<T: Config> =
         StorageMap<_, Identity, BoundedVec<u8, MigrationKeyMaxLen>, bool, ValueQuery>;
 
-    /// Alpha reservoir for scraps of protocol claimed fees.
-    #[pallet::storage]
-    pub type ScrapReservoirAlpha<T> = StorageMap<_, Twox64Concat, NetUid, AlphaBalance, ValueQuery>;
-
     #[pallet::event]
```
lorismoulin, line 125: Why are we removing `ScrapReservoirAlpha` here?

pallets/swap/src/pallet/mod.rs
```diff
@@ -26,8 +26,20 @@ type MigrationKeyMaxLen = ConstU32<128>;
...
+    #[derive(Encode, Decode, Default, TypeInfo, Clone, PartialEq, Eq, Debug, MaxEncodedLen)]
+    pub enum CleanUpPhaseEnum {
+        #[default]
+        /// Phase 1: Clear protocol liquidity remove liquidity.
+        ClearProtocolLiquidityRemoveLiquidity,
+        /// Phase 2: Clear protocol liquidity tick index bitmap words.
+        ClearProtocolLiquidityTickIndexBitmapWords,
+        /// Phase 3: Clear protocol liquidity parameters.
+        ClearProtocolLiquidityParameters,
+    }
```
lorismoulin: This is not used anywhere

pallets/subtensor/src/tests/requested_functions_tests.rs
```diff
@@ -0,0 +1 @@
+
```
lorismoulin: Empty file?

pallets/subtensor/src/staking/claim_root.rs
lorismoulin: Using an ensure would make this more consistent with the rest of the codebase given that the first thing we do when dissolving a netwwork is removing it from NetworksAdded

```rust
ensure!(
    Self::if_subnet_exist(netuid),
    Error::<T>::SubnetNotExists
);
```

pallets/subtensor/src/staking/claim_root.rs
```diff
@@ -453,15 +459,66 @@ impl<T: Config> Pallet<T> {
...
+            Some(raw_key) => RootClaimable::<T>::iter_from(raw_key),
+            None => RootClaimable::<T>::iter(),
+        };
+
+        // Iterate directly without collecting to avoid unnecessary allocation
+        for (hotkey, _) in iter {
+            let can_consume = weight_meter.can_consume(T::DbWeight::get().reads(2));
+            if !can_consume {
+                read_all = false;
+                LastKeptRawKey::<T>::set(Some(RootClaimable::<T>::hashed_key_for(&hotkey)));
```
lorismoulin: Write not accounted for

pallets/subtensor/src/staking/claim_root.rs
```diff
@@ -453,15 +459,66 @@ impl<T: Config> Pallet<T> {
...
+                break;
+            }
+            weight_meter.consume(T::DbWeight::get().reads(2));
+
+            let mut claimable = RootClaimable::<T>::get(&hotkey);
+            if claimable.contains_key(&netuid) {
+                let can_consume = weight_meter.can_consume(T::DbWeight::get().writes(1));
+                if !can_consume {
+                    read_all = false;
+                    LastKeptRawKey::<T>::set(Some(RootClaimable::<T>::hashed_key_for(&hotkey)));
```
lorismoulin: Write not accounted for.

pallets/subtensor/src/staking/claim_root.rs
```diff
@@ -453,15 +459,66 @@ impl<T: Config> Pallet<T> {
...
-        for hotkey in hotkeys.iter() {
-            RootClaimable::<T>::mutate(hotkey, |claimable| {
                 claimable.remove(&netuid);
-            });
+                to_remove_map.insert(hotkey.clone(), claimable);
+            }
+        }
+
+        if read_all {
+            LastKeptRawKey::<T>::set(None);
```
lorismoulin: Write not accounted for.

pallets/subtensor/src/staking/claim_root.rs
```diff
@@ -453,15 +459,66 @@ impl<T: Config> Pallet<T> {
...
+        }
+
+        // write weight already consumed in advance
+        for (hotkey, claimable) in to_remove_map.iter() {
+            RootClaimable::<T>::insert(hotkey, claimable);
         }
```
lorismoulin: I assume this writes after the loop to avoid mutating `RootClaimable` while iterating over it. If so, could we make that explicit in the comment? Right now “write weight already consumed in advance” explains the accounting, but not why the writes are delayed.

Also, when removing the subnet entry leaves the claimable map empty, should this remove `RootClaimable::<T>::remove(hotkey)` instead of inserting an empty map back into storage?

pallets/subtensor/src/staking/lock.rs
```diff
@@ -1811,59 +1816,95 @@ impl<T: Config> Pallet<T> {
...
+    /// Removes `Lock` entries for `netuid`, resuming from `LastKeptRawKey` when weight is limited.
+    pub fn remove_network_lock(netuid: NetUid, weight_meter: &mut WeightMeter) -> bool {
+        let r = T::DbWeight::get().reads(1);
+        let w = T::DbWeight::get().writes(1);
+        let mut keys_to_remove: sp_std::vec::Vec<(T::AccountId, T::AccountId)> =
+            sp_std::vec::Vec::new();
+        let mut read_all = true;
+
+        let iter = match LastKeptRawKey::<T>::get() {
+            Some(key) => Lock::<T>::iter_from(key),
```
lorismoulin, line 1963: Using `iter_from(LastKeptRawKey)` assumes the underlying storage ordering remains stable between cleanup passes. Is that guaranteed here? If new `Lock` entries can be inserted or existing entries removed between passes, resuming from a raw key may skip entries that sort before the saved key.

Maybe the intended invariant is that once a subnet is queued for dissolve cleanup, no new state for that `netuid` can be written. If so, I think that should be documented/enforced near the cleanup state machine, especially for global maps like `Lock` where the cursor is not scoped to the subnet prefix.

pallets/subtensor/src/tests/mock.rs
```diff
@@ -31,9 +33,51 @@ use sp_runtime::{
...
+}
+impl TypeInfo for TestMaxFields {
+    type Identity = Self;
+    fn type_info() -> scale_info::Type {
+        scale_info::Type::builder()
+            .path(scale_info::Path::new("TestMaxFields", module_path!()))
+            .composite(scale_info::build::Fields::unit())
+    }
+}
+
```
lorismoulin: Is this really needed? Can't we just just use a `parameter_types!`?

pallets/subtensor/src/tests/mock.rs
```diff
@@ -31,9 +33,51 @@ use sp_runtime::{
...
+        block_with_offset.checked_div(tempo_plus_one).unwrap_or(0)
+    }
+}
+
+// Implement OnMetadataCommitment for U256 by creating a local wrapper type
+pub struct TestOnMetadataCommitment;
+impl pallet_commitments::OnMetadataCommitment<U256> for TestOnMetadataCommitment {
+    fn on_metadata_commitment(_netuid: NetUid, _who: &U256) {}
+}
+
```
lorismoulin: I think it's implemented for `()`

pallets/subtensor/src/macros/hooks.rs
```diff
@@ -220,5 +229,327 @@ mod hooks {
...
+                            let done =
+                                Self::clean_up_root_claimed_for_subnet(*netuid, &mut weight_meter);
+
+                            if done {
+                                DissolvedNetworksCleanupPhase::<T>::set(Some(
+                                    DissolvedNetworksCleanupPhaseEnum::DestroyAlphaInOutStakesGetTotalAlphaValue,
+                                ));
+                            }
+                            done
+                        }
```
lorismoulin: It can be collapsed by returning Option<Phase> on each branch and setting it at the end.

pallets/subtensor/src/macros/errors.rs
```diff
@@ -277,6 +277,10 @@ mod errors {
         InvalidVotingPowerEmaAlpha,
         /// Deprecated call.
         Deprecated,
+        /// Subnet buyback exceeded the operation rate limit
+        SubnetBuybackRateLimitExceeded,
+        /// Network already dissolved
+        NetworkAlreadyDissolved,
```
lorismoulin: Could we rename this error to something like `NetworkDissolveInProgress` or `NetworkDissolveAlreadyQueued`? The check is for membership in the async cleanup queue, not necessarily for a subnet whose dissolve lifecycle has fully completed, so `NetworkAlreadyDissolved` sounds a bit too final.

pallets/subtensor/src/tests/remove_data_tests.rs
```diff
@@ -0,0 +1,862 @@
+#![allow(clippy::expect_used, clippy::indexing_slicing, clippy::unwrap_used)]
+
```
lorismoulin, line 2: This file and other tests files you added have `_tests` at the end which is not the case for other files.

pallets/subtensor/src/benchmarks.rs
```diff
@@ -2010,6 +2010,22 @@ mod pallet_benchmarks {
...
+
+        // Initialize network
+        Subtensor::<T>::init_new_network(netuid, tempo);
+
+        // Set network owner
+        SubnetOwner::<T>::insert(netuid, coldkey.clone());
+
+        #[extrinsic_call]
+        _(RawOrigin::Root, coldkey.clone(), netuid);
+    }
```
lorismoulin, line 2112: Please add a simple assertion after the call so we can ensure it did what we expected.

pallets/subtensor/src/staking/remove_stake.rs
```diff
@@ -420,20 +424,42 @@ impl<T: Config> Pallet<T> {
...
-        // 1) Ensure the subnet exists.
-        ensure!(Self::if_subnet_exist(netuid), Error::<T>::SubnetNotExists);
+    pub fn destroy_alpha_in_out_stakes(netuid: NetUid, weight_meter: &mut WeightMeter) -> bool {
+        let total_alpha_value_u128: u128 = match DissolvedSubnetTotalAlphaValue::<T>::get() {
+            Some(value) => value,
+            None => {
+                log::warn!("DissolvedSubnetTotalAlphaValue not set");
+                return false;
+            }
+        };
```
lorismoulin: let Some can be used here

pallets/subtensor/src/staking/remove_stake.rs
```diff
@@ -420,20 +424,42 @@ impl<T: Config> Pallet<T> {
...
 
-        // 2) Owner / lock cost.
+        let mut distributed_tao_value_u128: u128 = match DissolvedSubnetDistributedTao::<T>::get() {
+            Some(value) => value,
+            None => {
+                log::warn!("DissolvedSubnetDistributedTao not set");
+                return false;
+            }
+        };
```
lorismoulin: let Some can be used here

pallets/subtensor/src/staking/remove_stake.rs
```diff
@@ -512,35 +641,153 @@ impl<T: Config> Pallet<T> {
...
-        // 6) Pro‑rata distribution of the pot by α value (largest‑remainder),
-        //    **credited directly to each staker's COLDKEY free balance**.
-        if pot_u64 > 0 && total_alpha_value_u128 > 0 {
-            struct Portion<A, C> {
-                _hot: A,
-                cold: C,
-                is_protocol: bool,
-                share: u64, // TAO to credit to coldkey balance
-                rem: u128,  // remainder for largest‑remainder method
+        read_all
```
lorismoulin: My impression is that this block of code is reused a lot of time and we should extract a function, eventually taking a closure to avoid a macro. It would make the whole thing much easier to audit.

pallets/subtensor/src/coinbase/root.rs
```diff
@@ -286,9 +408,6 @@ impl<T: Config> Pallet<T> {
         RegistrationsThisInterval::<T>::remove(netuid);
         POWRegistrationsThisInterval::<T>::remove(netuid);
         BurnRegistrationsThisInterval::<T>::remove(netuid);
```
lorismoulin: Maybe we can extract small functions instead of having comment? Way easier to read.

pallets/subtensor/src/coinbase/root.rs
```diff
@@ -351,128 +459,364 @@ impl<T: Config> Pallet<T> {
...
+        }
+        if read_all {
+            LastKeptRawKey::<T>::set(None);
+        }
+
+        for hot in to_rm {
+            IsNetworkMember::<T>::remove(&hot, netuid);
+        }
+        read_all
+    }
```
lorismoulin: A lot of the cleanup phases repeat the same weighted scan/checkpoint/remove pattern. Could we factor that into a local helper that takes closures for iteration, cursor key construction, and removal? For example, a helper could own the read/write weight checks and `LastKeptRawKey` handling, while each phase only provides `iter`, `hashed_key_for`, and `remove`. That would make the state-machine code much easier to audit.

Rough idea of a possible signature:

```rust
fn remove_account_entries_for_netuid<I>(
    weight_meter: &mut WeightMeter,
    netuid: NetUid,
    iter: I,
    raw_key_for: impl Fn(&T::AccountId, NetUid) -> Vec<u8>,
    remove: impl Fn(T::AccountId, NetUid),
) -> bool
where
    I: Iterator<Item = (T::AccountId, NetUid)>,
{
    ...
}
```

lorismoulin (changes requested):
Some of this is a structural/style suggestion rather than a hard blocker, but I think it would make the async dissolve lifecycle much easier to reason about and maintain.

___

This PR seems like a good opportunity to move `do_dissolve_network` out of `coinbase/root.rs`. With the new async cleanup flow, dissolution is really subnet lifecycle logic rather than root/coinbase logic. Could we move it into a dedicated module such as `subnets/dissolution.rs`, alongside the cleanup queue/state/phase code?

___

One broader cleanup-state concern: many of the weighted cleanup loops implement their own resume cursor using `LastKeptRawKey` plus `StorageMap::hashed_key_for(...)`.

Since `clear_prefix` already returns `MultiRemovalResults { maybe_cursor, ... }`, could we use that cursor directly for prefix-clearing phases instead of maintaining a separate raw-key checkpoint? That would align with FRAME’s intended continuation API and avoid hand-rolling cursor keys per storage type.

For the phases that iterate manually with `iter_from`, it may still be worth encapsulating the cursor as part of the phase state rather than a single global `LastKeptRawKey`. Right now the meaning of `LastKeptRawKey` depends on the active phase and storage map, which makes stale cursor bugs harder to reason about.

common/src/lib.rs
open-junius: Agreed. I will remove it. not simplify too much.

common/src/lib.rs
open-junius: Always try to use the macro to wrap the code logic. Actually the simple function can do the same. It is much easier to check. will remove the macro also.

pallets/subtensor/src/macros/errors.rs
open-junius: will rename it.

pallets/subtensor/src/macros/events.rs
open-junius: will use "NetworkDissolveCleanupCompleted"

pallets/subtensor/src/macros/hooks.rs
open-junius: Make sense, will move to new dissolution.rs file.

pallets/subtensor/src/staking/lock.rs
open-junius: Yes, after a subnet is dissolved and in the cleanup queue. We can't insert any data for the netuid anymore. otherwise, the cleanup never could be completed. I checked all extrinsics, should forbid any data storage related to the dissolved subnet, also in the hook functions like coinbase.
Since some new thing added like Lock, conviction. I will check them again.

pallets/swap/src/pallet/mod.rs
open-junius: will remove it. a lot of data storage already cleaned.

pallets/swap/src/pallet/mod.rs
open-junius: Add data is already removed. maybe the storage definition is still useful, let me put it back.

pallets/subtensor/src/benchmarks.rs
open-junius: will check if netuid in the cleanup queue.

pallets/subtensor/src/tests/mock.rs
open-junius: just use the same implementation in the commitment pallet. will try parameter_types! macro

open-junius: Just tried it. We need TypeInfo impl for TestMaxFields. but parameter_types! doesn't auto implement it. Derive(TypeInfo) is ok. I will use it.

pallets/subtensor/src/tests/requested_functions_tests.rs
open-junius: will remove it.

JohnReedV (dismissed):

pallets/subtensor/src/utils/cleanup.rs
```diff
@@ -0,0 +1,45 @@
...
+            }
+        }
+
+        for hot in to_rm {
+            ops_based_on_key(&hot);
+        }
+
+        (read_all, last_item)
+    }
+}
```
lorismoulin, line 46: 👍

pallets/subtensor/src/subnets/dissolution.rs
```diff
@@ -0,0 +1,968 @@
...
+                        if done {
+                            DissolvedNetworksCleanupPhase::<T>::set(Some(
+                                DissolveCleanupPhase::CleanSubnetRootDividendsRootClaimed,
+                            ));
+                        } else {
+                            DissolvedNetworksCleanupPhase::<T>::set(Some(
+                                DissolveCleanupPhase::CleanSubnetRootDividendsRootClaimable {
+                                    last_key: new_key,
+                                },
+                            ));
```
lorismoulin: Phase transition weight is never accounted for.

pallets/subtensor/src/lib.rs
```diff
@@ -320,6 +323,26 @@ pub mod pallet {
         pub additional: Vec<u8>,
     }
 
+    /// Data structure for a pending network registration in the execution queue.
+    #[crate::freeze_struct("93f81374b91abeff")]
+    #[derive(Encode, Decode, Default, TypeInfo, Clone, PartialEq, Eq, Debug)]
+    pub struct NetworkRegistrationInfo<AccountId> {
```
lorismoulin: Please move this in `subnets.rs` close to the code that is is being used.

pallets/subtensor/src/lib.rs
```diff
@@ -349,6 +372,113 @@ pub mod pallet {
             subnets: BTreeSet<NetUid>,
         },
     }
+    /// Enum for the dissolve cleanup phase.
+    #[derive(Encode, Decode, TypeInfo, Clone, PartialEq, Eq, Debug, DecodeWithMemTracking)]
+    pub enum DissolveCleanupPhase {
```
lorismoulin: Please move this to `dissolution.rs` close to the code it is being used.

Also I think you can remove the Clean/Destroy from the name, just keep the name of where we are in the cleanup, like `RootDividendsRootClaimed`, `ProtocolLiquidity`, etc...

pallets/subtensor/src/lib.rs
```diff
@@ -2119,6 +2249,33 @@ pub mod pallet {
...
+    pub type DissolvedSubnetTotalAlphaValue<T> = StorageValue<_, u128, OptionQuery>;
+
+    /// --- ITEM ( dissolved_subnet_settled_alpha_value ) Settled alpha value for the dissolved subnet.
+    /// It is only used during clean the data for dissolved networks.
+    #[pallet::storage]
+    pub type DissolvedSubnetDistributedTao<T> = StorageValue<_, u128, OptionQuery>;
+
+    /// --- ITEM ( last_kept_raw_key ) Resume key for weight-limited cleanup within a phase.
+    #[pallet::storage]
+    pub type LastKeptRawKey<T> = StorageValue<_, Vec<u8>, OptionQuery>;
```
lorismoulin: The resume key is kept in the phase cleanup enum, this is not used anywhere.

pallets/subtensor/src/subnets/dissolution.rs
```diff
@@ -0,0 +1,968 @@
...
+    // 	* 'remaining_weight': (Weight):
+    // 		- The weight available for this cleanup step.
+    // 	* 'netuid': (&NetUid):
+    // 		- The subnet to clean dissolved-network data for.
+    //
+    // # Returns:
+    // 	* 'Weight': The weight used for the cleanup step.
+    //
+    pub fn remove_data_for_dissolved_networks(remaining_weight: Weight, netuid: &NetUid) -> Weight {
+        let mut weight_meter = frame_support::weights::WeightMeter::with_limit(remaining_weight);
```
lorismoulin: This function should not be passed the netuid but directly fetch the active netuid being processed directly given it is accessing the global `DissolvedNetworksCleanupPhase`

pallets/subtensor/src/subnets/dissolution.rs
```diff
@@ -0,0 +1,968 @@
...
+                true,
+            );
+            // just complete one registration at a time since on_idle just complete one network dissolve cleanup
+            // if one registration fails, then try next one. it could be not align with the order of registration in the queue
+            if result.is_ok() {
+                NetworkRegistrationQueue::<T>::mutate(|queue| queue.remove(index));
+                break;
+            }
+        }
+    }
```
lorismoulin, line 996: An error for an item in the queue will keep the item in the queue indefinitely and lock user funds.

The file now holds dissolve queue, cleanup phase execution, map-cleanup helpers, and
registration queue processing.

Registration draining belongs with subnet registration, or in
a small coordinator that explicitly reacts to cleanup completion. Using a final phase eventually.

pallets/subtensor/src/lib.rs
```diff
@@ -2119,6 +2249,33 @@ pub mod pallet {
...
+
+    /// --- ITEM ( dissolved_subnet_total_alpha_value ) Total alpha value for the dissolved subnet.
+    /// It is only used during clean the data for dissolved networks.
+    #[pallet::storage]
+    pub type DissolvedSubnetTotalAlphaValue<T> = StorageValue<_, u128, OptionQuery>;
+
+    /// --- ITEM ( dissolved_subnet_settled_alpha_value ) Settled alpha value for the dissolved subnet.
+    /// It is only used during clean the data for dissolved networks.
+    #[pallet::storage]
+    pub type DissolvedSubnetDistributedTao<T> = StorageValue<_, u128, OptionQuery>;
```
lorismoulin: `DissolvedNetworksCleanupPhase` should be called something like `ActiveNetworkBeingDissolved` and holds both the netuid and the phase because having to pop the first in the cleanup queue everytime is hacky and error prone.

The `DissolveCleanupQueue` should only hold the networks waiting for dissolution, not the one being dissolved.


`TotalAlphaValue` and `DistributedTao` can b merged into the `DissolveCleanupPhase` given this is something accumulated during the cleanup, no need for additional storage for this.

pallets/subtensor/src/macros/hooks.rs
```diff
@@ -185,12 +185,23 @@ mod hooks {
...
+        fn on_idle(_block: BlockNumberFor<T>, limit: Weight) -> Weight {
+            let dissolved_networks = DissolveCleanupQueue::<T>::get();
+            let weight = match dissolved_networks.get(0) {
+                Some(netuid) => Self::remove_data_for_dissolved_networks(limit, netuid),
+                None => Weight::from_parts(0, 0),
+            };
+            Self::process_network_registration_queue();
+
+            weight
+        }
```
lorismoulin, line 201: `process_network_registration_queue` weight should be accounted for and only called if we cleanup up a network recently, eventually we want an end phase that helps us detect we just finished a cleanup instead of blindly trying to process from the queue.

pallets/subtensor/src/macros/hooks.rs
```diff
@@ -185,12 +185,23 @@ mod hooks {
             // Self::check_total_stake()?;
             Ok(())
         }
+
+        fn on_idle(_block: BlockNumberFor<T>, limit: Weight) -> Weight {
+            let dissolved_networks = DissolveCleanupQueue::<T>::get();
+            let weight = match dissolved_networks.get(0) {
```
lorismoulin: Also this part is hacky. We should not be taking the network being dissolved from the first item in the queue.

We should pop an item from the queue when the active network being dissolved is none.

lorismoulin (changes requested):
Critical issue with the use of [`iter_from`](https://paritytech.github.io/polkadot-sdk/master/frame_support/storage/trait.IterableStorageMap.html#tymethod.iter_from) because it resumes from the key **AFTER** the one passed as argument, which will cause partial cleanup and stale storage, for example alpha stake won't be fully cleaned up if the fully hotkey could be processed and we will just jump to the next hotkey.

pallets/subtensor/src/utils/cleanup.rs
open-junius: >Will fix it.

pallets/subtensor/src/subnets/dissolution.rs
open-junius: will add the weight for phase update

pallets/subtensor/src/lib.rs
open-junius: moved to subnets.rs

pallets/subtensor/src/lib.rs
open-junius: updated according to your comment.

pallets/subtensor/src/lib.rs
open-junius: removed.

pallets/subtensor/src/subnets/dissolution.rs
open-junius: updated.

pallets/subtensor/src/subnets/dissolution.rs
open-junius: Yes, set_new_network_state could be failed and we can not avoid the failure. I was thinking if we need a new extrinsic to cancel the network registration for registration in the queue. then we can return the locked fund to the user.

pallets/subtensor/src/lib.rs
open-junius: For iteration raw key, we can store it in the DissolveCleanupPhase, because it is updated in the same phase and clean it up after the phase is done. The dissolved network already removed from the DissolveCleanupQueue

pallets/subtensor/src/macros/hooks.rs
open-junius: will compute the weight for process_network_registration_queue, and take it into whole weight in on_idle.

pallets/subtensor/src/macros/hooks.rs
open-junius: will fix this part.

pallets/subtensor/src/lib.rs
open-junius: refactor the code, use a new storage to store all status data used for one dissolved network.

pallets/subtensor/src/macros/hooks.rs
open-junius: For process_network_registration_queue, it could be failed to complete one network registration. That is why I call it even there is no dissolved network cleanup done.

open-junius: `if current_count.saturating_add(cleanup_queue_len) >= subnet_limit`, always checked. It is used to avoid there is no netuid available, no based on if a dissolved network cleanup is completed.

Cargo.toml
```diff
@@ -321,5 +321,7 @@ pow-faucet = []
 
 [patch.crates-io]
 w3f-bls = { git = "https://github.com/opentensor/bls", branch = "fix-no-std" }
+# core2 was yanked from crates.io; keep the last published sources available via git.
+core2 = { git = "https://github.com/bbqsrc/core2", rev = "545e84bcb0f235b12e21351e0c69767958efe2a7" }
 zstd-sys = { git = "https://github.com/gztensor/zstd-sys" }
```
lorismoulin: Why do we need this?

pallets/subtensor/src/macros/hooks.rs
```diff
@@ -187,12 +187,20 @@ mod hooks {
             // Self::check_total_stake()?;
             Ok(())
         }
+
+        fn on_idle(_block: BlockNumberFor<T>, limit: Weight) -> Weight {
+            let mut weight = Self::remove_data_for_dissolved_networks(limit);
+
+            weight.saturating_accrue(Self::process_network_registration_queue());
+
```
lorismoulin, line 199: Make sense to check that `weight < limit` before calling the `Self::process_network_registration_queue()`, wdyt?

evgeny-s (commented):
Hey, @open-junius . Found 2 high-severity issues. Please check them:

[PR2412-HIGH-1-alpha-cleanup-offbyone.md](https://github.com/user-attachments/files/29341834/PR2412-HIGH-1-alpha-cleanup-offbyone.md)
[PR2412-HIGH-2-netuid-reuse.md](https://github.com/user-attachments/files/29341835/PR2412-HIGH-2-netuid-reuse.md)

pallets/subtensor/src/subnets/dissolution.rs
```diff
@@ -0,0 +1,981 @@
...
+        );
+
+        // Just remove the network from the added networks, it is used to check if the network is existed.
+        NetworksAdded::<T>::remove(netuid);
+        // Reduce the total networks count.
+        TotalNetworks::<T>::mutate(|n: &mut u16| *n = n.saturating_sub(1));
+
+        TotalStake::<T>::mutate(|total| *total = total.saturating_sub(SubnetTAO::<T>::get(netuid)));
+
+        dissolved_networks.push(netuid);
```
lorismoulin: This remark from the AI may be wise to look at, we don't want `TotalStake` to drift.

lorismoulin (changes requested):
Overall looks correct to me even if there is a lot of changes and I'm not able to internalize all of them.

My only concern is about the `TotalStake` drift reported by AI and also the use of `iter_from` which uses the **next** key and can create stale state if a key is not fully cleaned up.

pallets/subtensor/src/macros/hooks.rs
open-junius: I will add it, remove_data_for_dissolved_networks can guarantee limit not overused.

Cargo.toml
open-junius: There is dependency conflict without choose its rev. I will try if it is ok in CI
```
error: failed to select a version for the requirement `core2 = "^0.4"`
  version 0.4.0 is yanked
location searched: crates.io index
required by package `cid v0.9.0`
    ... which satisfies dependency `cid = "^0.9.0"` of package `sc-network v0.51.1 (https://github.com/opentensor/polkadot-sdk.git?rev=7cc54bf2d50ae3921d718736dfeb0de9468539c7#7cc54bf2)`
    ... which satisfies git depend
```

pallets/subtensor/src/subnets/dissolution.rs
open-junius: just wonder how to handle the storage. update all in the do_dissolve_network or change little by little in the cleanup process. If we don't update it immediately, it could impact emission computation. Do you have any advice?

open-junius:
> Hey, @open-junius . Found 2 high-severity issues. Please check them:
> 
> [PR2412-HIGH-1-alpha-cleanup-offbyone.md](https://github.com/user-attachments/files/29341834/PR2412-HIGH-1-alpha-cleanup-offbyone.md) [PR2412-HIGH-2-netuid-reuse.md](https://github.com/user-attachments/files/29341835/PR2412-HIGH-2-netuid-reuse.md)

I will fix it. Since the solution as "remove an item from queue then put it into CurrentDissolveCleanupStatus" is easily causing errors. I change it. When start the cleanup, just get an item from queue, then put it in CurrentDissolveCleanupStatus. After cleanup is done, we remove it from the queue.

JohnReedV (approved):

UnArbosSix (dismissed):

UnArbosSix (approved):

lorismoulin (approved):

PR #2414, opened 2026-02-10
Add storage leak detection test for subnet dissolution

ppolewicz:
## Summary
- Adds `test_dissolve_network_no_storage_leak` which automatically detects per-subnet storage that is not cleaned up when a subnet is dissolved
- Fixes `Swap::ScrapReservoirTao` and `Swap::ScrapReservoirAlpha` not being cleaned up in `do_clear_protocol_liquidity` (caught by the new test)

## How the test works
1. Snapshots **all** raw storage keys (via `sp_io::storage::next_key`) before subnet creation
2. Creates a subnet and exercises it fully: registers neurons, stakes, serves axon/prometheus, sets childkeys, sets weights, runs 2 epochs (for vpermit/bonds/dividends)
3. Dissolves the subnet via root
4. Snapshots all storage keys after dissolution
5. Any key present in "after" but not in "before" that belongs to `SubtensorModule` or `Swap` pallets (excluding a known global allowlist) is reported as a leak

## Why this is useful
This test is **future-proof**: when a developer adds a new per-netuid `StorageMap` but forgets to clean it up in `remove_network`, this test fails automatically with a clear error message naming the leaked storage item — no need to update the test itself.

## Test plan
- [x] `cargo test -p pallet-subtensor --lib` — 998 tests pass
- [x] `cargo clippy --workspace --all-features --all-targets` — clean

### DIFF
--- a/pallets/subtensor/src/coinbase/root.rs
+++ b/pallets/subtensor/src/coinbase/root.rs
@@ -362,12 +362,17 @@ impl<T: Config> Pallet<T> {
         StakeWeight::<T>::remove(netuid);
         LoadedEmission::<T>::remove(netuid);
 
+        // --- 18b. Root prop.
+        RootProp::<T>::remove(netuid);
+        RootClaimableThreshold::<T>::remove(netuid);
+
         // --- 19. DMAPs where netuid is the FIRST key: clear by prefix.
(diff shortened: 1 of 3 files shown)


ppolewicz:
continued in https://github.com/opentensor/subtensor/pull/2579

PR #2420, opened 2026-02-11
Emission suppression via root validator voting

ppolewicz:
## Summary

Adds a mechanism for root validators to vote to suppress TAO emissions from subnets they believe are not operating in the network's interest. When more than half of the stake-weighted root validators vote to suppress a subnet, its emission share is zeroed and redistributed proportionally to the remaining subnets.

### Key features

- **Stake-weighted voting**: Each root validator's vote is weighted by their root stake. Suppression activates when >50% of total registered root stake votes to suppress.
- **Root override**: A sudo extrinsic (`sudo_set_emission_suppression_override`) allows root to force-suppress or force-unsuppress any subnet, bypassing the vote outcome.
- **Root sell pressure control**: A global flag (`KeepRootSellPressureOnSuppressedSubnets`, default `true`) controls whether root validators still receive alpha dividends from suppressed subnets. When disabled, all alpha goes to subnet validators instead — removing root sell pressure on the suppressed subnet's token.
- **Coldkey-keyed votes**: Votes are keyed by coldkey (not hotkey) so they can be migrated during coldkey swaps. Swaps fail if the destination coldkey already has votes to prevent silent overwrites.
- **Per-epoch collection**: Votes are collected and suppression ratios are recalculated once per epoch per subnet, keeping on-chain computation bounded.

### New extrinsics

| Call index | Name | Access |
|---|---|---|
| 133 | `sudo_set_emission_suppression_override` | Root |
| 134 | `vote_emission_suppression` | Signed (coldkey) |
| 135 | `sudo_set_keep_root_sell_pressure_on_suppressed_subnets` | Root |

### New storage items

- `EmissionSuppression<NetUid → U64F64>` — current suppression ratio per subnet
- `EmissionSuppressionOverride<NetUid → Option<bool>>` — root override per subnet
- `EmissionSuppressionVote<(NetUid, AccountId) → Option<bool>>` — per-coldkey vote per subnet
- `KeepRootSellPressureOnSuppressedSubnets<bool>` — global flag (default true)

### Tests

24 test scenarios covering:
- Share zeroing and renormalization with majority/minority votes
- Override force-suppress and force-unsuppress
- Vote eligibility (root registration, minimum stake)
- Vote clearing and epoch-only collection
- Coldkey swap migration and conflict detection
- Subnet dissolution cleanup
- Root sell pressure flag behavior
- Root subnet vote rejection
- Multi-hotkey vote weight aggregation
- Sudo event emission

## Test plan

- [x] All 24 emission suppression tests pass (`cargo test -p pallet-subtensor --lib`)
- [x] Full pallet test suite passes
- [x] `cargo clippy --workspace --all-features --all-targets` clean
- [x] `scripts/fix_rust.sh` clean


ppolewicz:
## Code Review Fixes (136c1a49f)

Addressed findings from automated code review after merging `devnet-ready`:

### R-1: Disable mode now actually recycles root alpha to validators
Previously, Disable mode zeroed `root_alpha` before computation — alpha "stayed" with validators by never being separated. Now it computes `root_alpha` normally and explicitly adds it to `PendingValidatorEmission`, matching the documented "recycled to subnet validators" behavior.

### R-7: Extract suppression threshold constant
Replaced magic number `0.5` with `EMISSION_SUPPRESSION_THRESHOLD` constant in `subnet_emissions.rs`.

### R-9: Clarify variable naming
Renamed `suppression_mode` → `root_sell_pressure_mode` in `run_coinbase.rs` to avoid confusion with the suppression fraction.

### R-12: Fix weight estimation
`sudo_set_emission_suppression_override` now accounts for 2 reads (subnet existence check + storage op) instead of 1.

### New test
`test_disable_mode_recycles_root_alpha_to_validators` — runs emission with Enable mode then Disable mode on the same suppressed subnet and asserts that Disable validators receive more alpha by exactly the root alpha amount from Enable mode.

ppolewicz:
This is not going to work, I'm afraid, because root validators won't vote due to liability

pallets/subtensor/src/coinbase/subnet_emissions.rs
```diff
@@ -246,4 +251,78 @@ impl<T: Config> Pallet<T> {
...
+    }
+
+    /// Check if a subnet is currently emission-suppressed, considering override first.
+    pub(crate) fn is_subnet_emission_suppressed(netuid: NetUid) -> bool {
+        match EmissionSuppressionOverride::<T>::get(netuid) {
+            Some(true) => true,
+            Some(false) => false,
+            None => {
+                EmissionSuppression::<T>::get(netuid)
+                    > U64F64::saturating_from_num(EMISSION_SUPPRESSION_THRESHOLD)
```
gztensor, line 277: The f64 is non-deterministic, we don't use it in runtime. This needs to be done like:

```rust
U64F64::saturating_from_num(1)
    .safe_div(U64F64::saturating_from_num(2))
```

pallets/subtensor/src/coinbase/subnet_emissions.rs
```diff
@@ -246,4 +251,78 @@ impl<T: Config> Pallet<T> {
...
+    /// Called once per subnet per epoch. No-op for root subnet.
+    pub(crate) fn collect_emission_suppression_votes(netuid: NetUid) {
+        if netuid.is_root() {
+            return;
+        }
+        let root_n = SubnetworkN::<T>::get(NetUid::ROOT);
+        let mut suppress_stake = U64F64::saturating_from_num(0u64);
+        let mut total_root_stake = U64F64::saturating_from_num(0u64);
+
+        for uid in 0..root_n {
```
gztensor, line 309: You can iter_prefix on Keys to avoid this additional reading: `SubnetworkN::<T>::get(NetUid::ROOT)`

pallets/subtensor/src/lib.rs
```diff
@@ -2420,12 +2426,40 @@ pub mod pallet {
...
+    #[pallet::storage]
+    pub type EmissionSuppressionVote<T: Config> =
+        StorageDoubleMap<_, Identity, NetUid, Blake2_128Concat, T::AccountId, bool, OptionQuery>;
+
+    /// Controls how root alpha dividends are handled on emission-suppressed subnets.
+    /// - Disable (0x00): root gets no alpha; root alpha recycled to subnet validators.
+    /// - Enable  (0x01): root still accumulates alpha (old behaviour).
+    /// - Recycle (0x02, default): root alpha swapped to TAO and TAO burned.
+    #[pallet::storage]
+    pub type KeepRootSellPressureOnSuppressedSubnets<T: Config> =
```
gztensor, line 2454: This map is not removed when subnet is removed.

gztensor (changes requested):

bittoby:
HI, @ppolewicz Why you closed this PR? If possible, Can I complete this issue? I'd love to contribute! I can do

PR #2453, opened 2026-02-23
Disable shield on testnet

lorismoulin:


JohnReedV (approved):

open-junius (approved):
need bump the runtime version

PR #2459, opened 2026-02-25
feat: add origin AccountId to StakeAdded/StakeRemoved events

### LINKED ISSUE #311: Add the origin AccountId to StakeAdded and StakeRemoved events

Currently the StakeAdded/Removed events do not list the origin that initiated the stake add or removal action. This makes it hard to track staking action based on events and requires a link to the actual extrinsic triggering this event.

Add the origin's AccountId to these events. The information is already available in the functions emitting the events, so it should be as simple as adding another field to the Events tuple. 

_No response_

### Additional context

_No response_

altis0725:
Fixes #311

## Summary
- Add `origin` (`T::AccountId`) as the first field of `StakeAdded` and `StakeRemoved` events to identify which account initiated the staking operation
- The origin is the coldkey for direct extrinsic calls and the lease coldkey for automated distributions
- Update all emit sites in `stake_utils.rs` (`stake_into_subnet`, `unstake_from_subnet`, `transfer_stake_within_subnet`) to propagate the origin parameter
- Bump `spec_version` 380 → 381 for the event metadata change

## Test plan
- [x] Added `test_add_stake_event_has_origin` asserting `StakeAdded` origin == coldkey
- [x] Added `test_remove_stake_event_has_origin` asserting `StakeRemoved` origin == coldkey
- [x] Updated `test_do_move_event_emission` to verify origin in both `StakeRemoved` and `StakeAdded` during move operations
- [ ] CI compilation and existing test suite pass

### DIFF
--- a/pallets/subtensor/src/macros/events.rs
+++ b/pallets/subtensor/src/macros/events.rs
@@ -14,7 +14,9 @@ mod events {
         /// a network is removed.
         NetworkRemoved(NetUid),
         /// stake has been transferred from the a coldkey account onto the hotkey staking account.
+        /// (origin, coldkey, hotkey, tao_amount, alpha_amount, netuid, fee)
         StakeAdded(
+            T::AccountId,
             T::AccountId,
             T::AccountId,
             TaoCurrency,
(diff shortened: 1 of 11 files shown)


bdmason:
These event parameters come through as an array, adding the new param to the start of the array is a breaking change. It would be much less disruptive if it was added to the end of the params

altis0725:
Good catch, moved origin to the end of the params. Should be backwards-compatible now.

PR #2468, opened 2026-02-26
Coldkey swap dispatch guard

lorismoulin:
### Motivation

The previous `CheckColdkeySwap` implementation was a `TransactionExtension` that manually unwrapped one level of proxy calls to find the real account. This was fragile — nested proxies (`proxy(proxy(real, call))`) could bypass the check entirely.

By moving to a `DispatchGuard`, the check fires at every `call.dispatch(origin)` site, including inside the proxy pallet's `do_proxy()`. This automatically handles any nesting depth with the real resolved origin, without any manual proxy unwrapping.

### Changes

- Replaced the `CheckColdkeySwap` `TransactionExtension` with a `DispatchGuard` implementation in a new `guards/` module
- Added `ColdkeySwapActive` and `ColdkeySwapDisputed` pallet errors.
- Cleaned up `SubtensorTransactionExtension` by removing the inlined guard call and its extra trait bounds
- Deprecated the `ColdkeyInSwapSchedule` `CustomTransactionError` variant (coldkey swap now uses announcements and the check moved to the DispatchGuard) and removed the unused `ColdkeySwapDisputed` and `InvalidRealAccount` variants

### Test coverage

- No active swap allows calls through
- `None` and `Root` origins bypass the guard
- Active swap blocks forbidden calls (`remark`, `add_stake`, `remove_stake`, `set_weights`, `register_network`)
- Active swap allows authorized calls (`announce_coldkey_swap`, `swap_coldkey_announced`, `dispute_coldkey_swap`, `submit_encrypted`)
- Disputed swap blocks all calls (forbidden + authorized)
- Proxied forbidden call is blocked at the inner dispatch level
- Nested proxy (2 levels deep) is blocked at the innermost dispatch level


l0r1s:
@ales-otf you may be interested in the implementation of a guard for the rate limiting

sam0x17 (approved):

open-junius (approved):
Looks good to me.

PR #2474, opened 2026-02-27
Use mortal eras for all transactions

gztensor:
## Description

Subtensor used to have immortal eras which can cause unwanted side effects. Making eras mortal. Currently the transaction life time will be 256 blocks (51.2 minutes) from the block of signing.

## Type of Change
- [ ] Documentation update

## Checklist

### DIFF
--- a/runtime/src/lib.rs
+++ b/runtime/src/lib.rs
@@ -164,13 +164,20 @@ impl frame_system::offchain::CreateSignedTransaction<pallet_drand::Call<Runtime>
     ) -> Option<Self::Extrinsic> {
         use sp_runtime::traits::StaticLookup;
 
+        // Transaction validity period from block hash included in signature
+        // Has to a power of 2. 256 blocks yields to 3072 s = 51.2 minutes.
+        let validity_period = 256_u64;
+        let current_block = frame_system::Pallet::<Runtime>::block_number();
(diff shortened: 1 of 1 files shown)


open-junius:
Just curious why it is 2048 blocks in UI, the value is 256 in code.

l0r1s (approved):

runtime/src/lib.rs
l0r1s, line 157: This code is never called and we can safely remove it.

Drand only send unsigned extrinsic (for the pulse) and sign the payload itself. We never send signed extrinsic directly from the Drand pallet and this code only provides the required bound to use the [`send_signed_transaction`](https://paritytech.github.io/polkadot-sdk/master/frame_system/offchain/trait.SendSignedTransaction.html) from the offchain worker which is never used because it uses [`send_unsigned_transaction`](https://paritytech.github.io/polkadot-sdk/master/frame_system/offchain/trait.SendUnsignedTransaction.html).

The only way to override the Era would be to wrap the CheckEra (CheckMortality) to override the Era the same way we did for the TransactionPaymentWrapper.

Era is set from the client side and not from inside the runtime, the runtime only expose the transaction extension pipeline encoding and how the data should be organized in the extrinsic.

l0r1s (changes requested):

PR #2478, opened 2026-03-02
Add `transferStakeFrom` and `approve` functions to staking precompile

jeremyp-tao:
## Description
Adds new functions to the staking precompile similar to ERC20 `transferFrom` and `approve`, which will allow smart contracts to atomically receive alpha tokens and react to the transfer, which is not possible today (if an user transfers alpha token to a smart contract, the latter is not aware of this transfer). Allowance is managed per netuid for finer control.

## Related Issue(s)

None

## Type of Change

- [ ] Documentation update

## Breaking Change

None

## Checklist

Not applicable

## Additional Notes

None


precompiles/src/staking.rs
```diff
@@ -447,6 +473,177 @@ where
...
+
+        let source = R::AccountId::from(source_coldkey.0);
+        let spender = R::AccountId::from(spender_coldkey.0);
+        let netuid = try_u16_from_u256(origin_netuid)?;
+
+        Ok(AllowancesStorage::<R>::get(source, (spender, netuid)))
+    }
+
+    #[precompile::public("increaseAllowance(bytes32,uint256,uint256)")]
+    fn increase_allowance(
```
open-junius, line 527: We can use approve interface to do increase/decrease allowance, and they are not in the standard ERC20 definition.

open-junius:
It is better to implement both function in an Alpha wrapper contract, instead of in precompile. @ppolewicz @camfairchild any update for ERC20-alpha-wrapper

precompiles/src/staking.rs
```diff
@@ -27,22 +27,48 @@
...
+}
+
+pub type AllowancesStorage<R> = StorageDoubleMap<
+    AllowancesPrefix,
+    // For each approver
+    Blake2_128Concat,
+    <R as frame_system::Config>::AccountId,
+    // For each pair of (spender, netuid)
+    Blake2_128Concat,
+    (<R as frame_system::Config>::AccountId, u16),
```
camfairchild: What happens on subnet de-reg?

camfairchild (changes requested):

precompiles/src/staking.rs
jeremyp-tao: Contracts can easily get the current allowance and increase it atomically (they must check for re-entrancy attacks), however users cannot do that. That's why those functions are very common in tokens implementing ERC20.

precompiles/src/staking.rs
jeremyp-tao: I'm not sure a de-reg would cause any issue. There approve would still be here but the user can update the allowance at anytime, and is suppose to trust/verify that the contract will do appropriate things with the approved funds. So I think it is not worth it to complicate the design.

open-junius:
please merge devnet-ready, will fix the failed CI tasks.

precompiles/src/staking.rs
open-junius: After de-reg, we need clean up the data, the transfer alpha doesn't work for de-reg netuid.
And for approved allowance, there are two options.
1. just clean up the storage
2. transfer alpha according to each allowance with the same netuid.

jeremyp-tao: Clearing the storing with the netuid prefix is a bad idea because there can be a lot of entries, and it is not safe in general to have unbounded storage storage writes.

If a netuid is de-reg, calling  `transferStakeFrom` will revert like `transferStake` even if there is allowance, because under the hood they call the same dispatchable. Once the netuid is registered again however, the transfers will work again (if user have funds in the new alpha), and allowance will be preserved.

For the last point (allowance preservation), I think this is not the issue as stated in my previous comment. If the user trust (or verify code) the contract with a long-live high amount allowance to not be malicious and transfer the funds in unwanted way, there is no reason the contract will behave differently after the de-reg/re-reg. If one think a contract may go rogue in the future, they should either not interact with it or make a short-live exact amount approval and consume it quickly.

In any case at the protocol level I don't see any risk of keeping the allowances across netuid de-regs, as they don't interact directly with the staking pallet, and instead ends up being a normal `transferStake` call (on the behalf of the approver) only when the spender call `transferStakeFrom`.

Do you see a concrete issue I missed?

open-junius: If a subnet is de-registered, its netuid will be locked and no one can use it again before all data are cleaned up. But it could be reused again after all old data removed from the storage. So the new registered subnet is different one even the netuid is the same, means the alpha is not the same.

jeremyp-tao: Clearing big storage maps is risky in Substrate in general :/

For the approvals as said above it's not really a problem imo that they stay across netuid de-reg/re-reg. Users trust the smart contract to not consume the approval without a valid reason.

Approvals are just numbers which allows transfering stake on the behalf of the approver. After de-reg/re-reg, the approver will have no stake and the transferFrom will fail because the transfer part itself will fail in the pallet (like if you called yourself transfer). If they stake for this new network with same netuid, their previous approval will still be there and transfer could be made indeed, but they already trusted the contract so it should not be an issue. If they don't trust the contract long-term, they should like with ERC20 only approve the required amount to be consumed in a short timeframe, returning the allowance back to 0.

So I really so no issue of keeping the allowance data in storage.

sam0x17 (approved):

precompiles/src/staking.rs
```diff
@@ -448,6 +482,177 @@ where
...
+        }
+
+        Ok(())
+    }
+
+    #[precompile::public("transferStakeFrom(address,bytes32,bytes32,uint256,uint256,uint256)")]
+    fn transfer_stake_from(
+        handle: &mut impl PrecompileHandle,
+        source_address: Address,
+        destination_coldkey: H256,
```
camfairchild, line 623: I think this should also be Address

jeremyp-tao: It is a legitimate use to transfer the stake to a non-EVM account.

camfairchild: I think the usecase is more likely to be with an EVM address as the destination given it's a smart contract doing the call. 

The point of using `Address` would be to avoid having to convert any EVM address into the right format.

Potentially you could add another function to allow using the non-EVM form of the address.

PR #2487, opened 2026-03-05
Fix the ability to revert hotkey swap

evgeny-s:
## Description
This PR fixes the inability to revert a hotkey swap by disabling `is_hotkey_registered_on_any_network` for cases where netuid is provided. 

**More changes**:
- Updated `HotkeySwapOnSubnetInterval` from 5 days to 1 day.
- Added hotkey swap revert tests for: `Parent keys`, `Child keys`, `Pending child keys`, `Auto stake destination`, `Subnet ownership`, `Dividend records`, `Voting Power`, `Root Reclaimable`.
- Removed test `test_swap_hotkey_registered_on_other_subnet`, since it tests the condition that is no longer valid.

## Type of Change
- [ ] Documentation update

## Breaking Change

## Checklist

## Additional Notes


pallets/subtensor/src/swap/swap_hotkey.rs
```diff
@@ -57,11 +57,13 @@ impl<T: Config> Pallet<T> {
...
-            !Self::is_hotkey_registered_on_any_network(new_hotkey),
-            Error::<T>::HotKeyAlreadyRegisteredInSubNet
-        );
+        // 7. Ensure the new hotkey is not already registered on any network, only if netuid is none
+        if netuid.is_none() {
+            ensure!(
+                !Self::is_hotkey_registered_on_any_network(new_hotkey),
+                Error::<T>::HotKeyAlreadyRegisteredInSubNet
+            );
+        }
```
open-junius, line 77: It is better to add is_hotkey_registered_on_specific_network check if netuid == Some(netuid).

evgeny-s: Added additional check, like you suggested here: https://github.com/opentensor/subtensor/pull/2487/changes/893bfb0009d40c2112188da41b52909767008b59

open-junius:
please cargo fmt the code and bump runtime version.

pallets/subtensor/src/swap/swap_hotkey.rs
```diff
@@ -57,11 +57,21 @@ impl<T: Config> Pallet<T> {
 
         weight.saturating_accrue(T::DbWeight::get().reads(2));
 
-        // 7. Ensure the new hotkey is not already registered on any network
-        ensure!(
-            !Self::is_hotkey_registered_on_any_network(new_hotkey),
-            Error::<T>::HotKeyAlreadyRegisteredInSubNet
-        );
+        // 7. Ensure the new hotkey is not already registered on any network, only if netuid is none
+        if netuid.is_none() {
```
ales-otf: since you have two paths here for `netuid == None` and `netuid == Some(_)` you can merge this check and the next `if let` into a single `match`:

```rust
match netuid {
 Some(id) => ...,
None => ...,
}
```

ales-otf (dismissed):

pallets/subtensor/src/swap/swap_hotkey.rs
evgeny-s: https://github.com/opentensor/subtensor/pull/2487/changes/fa41075c84769d6890721fbc42a5e0c19a0683b8

open-junius (dismissed):

pallets/subtensor/src/macros/dispatches.rs
```diff
@@ -1061,8 +1069,9 @@ mod dispatches {
             hotkey: T::AccountId,
             new_hotkey: T::AccountId,
             netuid: Option<NetUid>,
+            maybe_keep_stake: Option<bool>,
```
l0r1s: We need to notify @basfroman @ibraheem-abe about this because some changes may be required from their end.

l0r1s: Are `None` and `Some(false)` supposed to have the same behavior? If yes then maybe just having a `bool` is sufficient?

pallets/subtensor/src/swap/swap_hotkey.rs
```diff
@@ -323,24 +358,27 @@ impl<T: Config> Pallet<T> {
...
+            // 2. Swap total hotkey shares on all subnets it exists on.
+            // TotalHotkeyShares( hotkey, netuid ) -> alpha -- the total alpha that the hotkey has on a specific subnet.
+            let share = TotalHotkeyShares::<T>::take(old_hotkey, netuid);
+            TotalHotkeyShares::<T>::mutate(new_hotkey, netuid, |value| {
+                *value = value.saturating_add(share)
+            });
+            weight.saturating_accrue(T::DbWeight::get().reads_writes(1, 1));
+        }
 
         // 3. Swap all subnet specific info.
```
l0r1s, line 381: When keep_stake is true, we only keep the top hotkey stake, right? Because we still run this in both cases:
```
// 4. Swap ChildKeys.
// 5. Swap ParentKeys.
// 6. Swap PendingChildKeys.
Self::parent_child_swap_hotkey(old_hotkey, new_hotkey, netuid, weight)?;
```

pallets/subtensor/src/macros/dispatches.rs
```diff
@@ -1051,7 +1051,15 @@ mod dispatches {
...
+        /// # Arguments
+        /// * `origin` - The origin of the transaction (must be signed by the coldkey).
+        /// * `hotkey` - The old hotkey to be swapped.
+        /// * `new_hotkey` - The new hotkey to replace the old one.
+        /// * `netuid` - Optional subnet ID. If `Some`, swap only on that subnet; if `None`, swap on all subnets.
+        /// * `maybe_keep_stake` - If `true`, stake remains on the old hotkey and the rest metadata
+        ///   is transferred to the new hotkey.
         #[pallet::call_index(70)]
         #[pallet::weight((Weight::from_parts(275_300_000, 0)
         .saturating_add(T::DbWeight::get().reads(52_u64))
```
l0r1s, line 1067: Maybe we should put this in Pays::Yes

pallets/subtensor/src/macros/dispatches.rs
evgeny-s: Let's create a new extrinsic for this specific reason with the `keep_stake` parameter. It doesn't make sense to bother the clients. WDYT?

evgeny-s: I don't like the fact that we add a required parameter after the optional. In this case, clients need to explicitly specify `netuid: None` and `maybe_keep_stake: false`, even if they don't want it, but they would just ignore it, since both are optional. I would keep it optional.

pallets/subtensor/src/macros/dispatches.rs
evgeny-s: I see we have a rate limit for this extrinsic. So, we should be fine.

pallets/subtensor/src/macros/dispatches.rs
evgeny-s: We can also mark the current extrinsic as deprecated, so they will have plenty of time to migrate.

evgeny-s: https://github.com/opentensor/subtensor/pull/2487/changes/acfa24f57cd4b04926dbfb27a8306ee2d0606aad

pallets/subtensor/src/swap/swap_hotkey.rs
evgeny-s: It seems like we need to keep it untouched in case `keep_stake is true`, updated it here: https://github.com/opentensor/subtensor/pull/2487/changes/2be9066dacc7b5fabf426c68b03718826d9ac395

evgeny-s: @l0r1s , we might also need to keep the old hotkey as a parent and make a new hotkey its child, no? In order to keep the stake delegation.

evgeny-s: Or the client will set up a new delegation manually via `set_children` after the hotkey swap.

pallets/subtensor/src/macros/dispatches.rs
sam0x17: great approach!

l0r1s: Clients need to explicitly pass netuid in any case. There's no way to omit a parameter.

For example with dedot:
```
// Correct
.swapHotkeyV2(alice.address, bob.address, undefined, undefined)
.swapHotkeyV2(alice.address, bob.address, undefined, false)
// Incorrect, compile time error
.swapHotkeyV2(alice.address, bob.address)
```

Also the first thing we do in the code is:
```
let keep_stake = maybe_keep_stake.unwrap_or(false);
```

It adds unnecessary wrapping that needs to be dealt with client side. It would make sense if None and Some(false) meant something different fundamentally.

It's only used at the dispatch layer because the inner function calls take a `bool` directly and not `Option<bool>`.

evgeny-s: Sure, it makes sense. I will update it.

evgeny-s: https://github.com/opentensor/subtensor/pull/2487/commits/a78c36a62de615d94739d7a6dc6a89017e620ae1

pallets/subtensor/src/swap/swap_hotkey.rs
l0r1s: After looking at the code, it only involves changing the relationships and it doesn't move stake or change counters (Alpha or shares).

Probably better to remove the wrapping if and just let it run as it was before?

evgeny-s: https://github.com/opentensor/subtensor/pull/2487/commits/5b28aedca9a7003813750b1a234f149f161c57e5

l0r1s (approved):
lgtm

sam0x17 (approved):

PR #2488, opened 2026-03-05
Fix MEV shield timing

lorismoulin:
## Fix shielded transactions getting stuck in the pool

Addition to #2477 

Companion of opentensor/polkadot-sdk#11

## Problem

Shielded transactions were getting permanently stuck in the pool due to three compounding issues:

1. **Block boundary timing**: `NextKey` pointed to the N+1 block author's key. Users encrypting near a block boundary would read `NextKey`, but by the time their transaction landed in the pool the key had already rotated — the `InBlock` check in `CheckShieldedTxValidity` rejected the transaction because it no longer matched `CurrentKey` or `NextKey`.

2. **Non-authority submission path**: When a shielded transaction was submitted to a non-authority node and gossiped to an authority, the authority's extension rejected it during block proposal (`InBlock` source) but never reported it as invalid back to the pool since it was a gossiped transaction. The non-authority node (which never proposes blocks) kept the transaction in its pool and continued re-gossiping it.

3. **Broken eviction**: The `longevity: 3` on shielded transactions was meant as a safeguard to evict stale ones from the pool, but since the extension returned a constant value on every revalidation the countdown reset each block and transactions were never actually evicted.

## Breaking changes

**Mortal era enforcement**: `submit_encrypted` calls with immortal or >8-block eras are now rejected with `InvalidTransaction::Stale`. You must explicitly set a short era (<=8 blocks) when sending shielded transactions (the actual submit_encrypted wrapper). Non-shielded transactions are not affected.

**Inner transaction era latest block**: Some clients, may cause issues if they are using the latest block number in era to build the inner extrinsic (the one that will be encrypted). With these clients, you need to override the era's block else the inner extrinsic will fail silently and only the outer extrinsic will be included in the block. For inner extrinsics, it is recommended to use era with block `n-1` or a finalized block.

### Key rotation rework

- `NextKey` now exposes the **N+2** author's key (user-facing), giving users a full 12-24s submission window instead of 0-12s, eliminating the block boundary timing issue.
- New `PendingKey` storage holds the N+1 block author's key, staged one block ahead so the proposer can read it at the parent hash.
- `announce_next_key` rotates in a clear 4-step order: `CurrentKey <- PendingKey`, `PendingKey <- NextKey`, `NextKey <- next_next_author's key`, then `AuthorKeys[current] <- announced key`.

### Key expiration tracking

New `PendingKeyExpiresAt` and `NextKeyExpiresAt` storage values expose the block number at which each user-facing key stops being valid (exclusive upper bound). Updated every block during rotation. Clients can read these directly to know how long a key remains usable.

### New `is_shielded_using_current_key`

Checks a transaction's key hash against `PendingKey`. The block proposer (in the companion polkadot-sdk PR) reads runtime state at the parent hash to determine if a shielded tx is meant for the block being built, and skips it if not.

### Simplified transaction extension

`CheckShieldedTxValidity` now only rejects malformed ciphertext. Key hash matching and the constant `longevity: 3` override have been removed — inclusion decisions are handled by the block proposer, and eviction is handled by mortal eras.

### Mortal era enforcement

New `CheckMortality` runtime extension wraps Substrate's `CheckMortality` and rejects `submit_encrypted` calls with immortal or >8-block eras. This replaces the broken `longevity: 3` with a proper eviction mechanism — `CheckMortality` computes `death - current_block`, which naturally decrements each block as the pool expects.

### Localnet

Added third validator node for proper round-robin testing.

### Pallet tests
- `key_rotation_round_robin`: 5-block simulation of 3-validator round-robin verifying the full pipeline (CurrentKey, PendingKey, NextKey, AuthorKeys, expirations, `is_shielded_using_current_key`) at each block.
- Error paths: signed origin, bad key length, missing author, None key cleanup, pre-update AuthorKeys ordering.
- Extension: malformed ciphertext rejection, well-formed acceptance across all sources.

### Runtime tests
- `check_mortality`: era rejection for shield txs, pass-through for non-shield calls.

### E2E tests
- Basic: encrypted submission, decryption by proposer, multi-tx batching, malformed/stale key rejection.
- Timing: submission at various points within a block window confirms transactions land within 1-2 blocks.
- Mortality: confirms `submit_encrypted` with default era (64) is rejected, and transactions with short era (8) are properly evicted from the pool after expiry.

### DIFF
--- a/pallets/shield/README.md
+++ b/pallets/shield/README.md
@@ -4,19 +4,19 @@ FRAME pallet for opt-in, per-block ephemeral-key encrypted transactions (MEV shi
 
 ## Overview
 
-Block authors rotate ML-KEM-768 key pairs every slot via a mandatory inherent. Users encrypt their extrinsics to the next block author's public key, preventing front-running and sandwich attacks.
+Block authors rotate ML-KEM-768 key pairs every slot via a mandatory inherent. Users encrypt their extrinsics to the next block author's encapsulation key, preventing front-running and sandwich attacks.
 
 ### Key rotation
 
 Each block includes an `announce_next_key` inherent that:
 
 1. Shifts `NextKey` into `CurrentKey` (so the previous key is still accepted during the transition).
-2. Stores the current author's freshly generated public key in `AuthorKeys`.
+2. Stores the current author's freshly generated encapsulation key in `AuthorKeys`.
 3. Looks up the *next* author's key from `AuthorKeys` and exposes it as `NextKey`.
 
 ### Encrypted transaction flow
 
-1. User reads `NextKey` from storage (ML-KEM-768 public key, 1184 bytes).
+1. User reads `NextKey` from storage (ML-KEM-768 encapsulation key, 1184 bytes).
 2. User encrypts a signed extrinsic with ML-KEM-768 + XChaCha20-Poly1305, producing:
 
    ```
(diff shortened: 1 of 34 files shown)


sam0x17 (dismissed):

JohnReedV (dismissed):

open-junius (dismissed):

pallets/shield/README.md
```diff
@@ -38,10 +38,10 @@ Each block includes an `announce_next_key` inherent that:
 | Item | Description |
 |------|-------------|
 | `CurrentKey` | Previous block's `NextKey`, kept for one-block grace period |
-| `NextKey` | Public key users should encrypt to |
-| `AuthorKeys` | Per-authority latest announced public key |
+| `NextKey` | Encapsulation key users should encrypt to |
+| `AuthorKeys` | Per-authority latest announced encapsulation key |
```
antlerminator, line 51: This should be updated to reflect changes done in the PR

l0r1s: updated the README to reflect the latest changes

sam0x17 (approved):

JohnReedV (approved):

PR #2500, opened 2026-03-16
Fix the benchmark for disable_lp

open-junius:
## Description
The main purpose of the PR is to get the weigth for dissolve_all_liquidity_providers function,
which is called in dissolve_network. I'd like to update it before continue with my dissolve network PR.
The disable_lp also call it, just use the computed weight to set weight of disable_lp extrinsic. 
I didn't update the benchmark because later, disable_lp will be removed after balancer implementation https://github.com/opentensor/subtensor/pull/2505/changes#diff-59099c09418a27c73c83511de538c50ab52c17424cfb89582e241c95f0ec9c06. 

## Related Issue(s)

- Closes #[issue number]

## Type of Change

- [ ] Documentation update

## Breaking Change

## Checklist

## Additional Notes


pallets/swap/src/pallet/mod.rs
```diff
@@ -606,14 +606,17 @@ mod pallet {
         ///
         /// Emits `Event::UserLiquidityToggled` on success
         #[pallet::call_index(5)]
-        #[pallet::weight(<T as pallet::Config>::WeightInfo::modify_position())]
-        pub fn disable_lp(origin: OriginFor<T>) -> DispatchResult {
+        #[pallet::weight(<T as pallet::Config>::WeightInfo::disable_lp())]
+        pub fn disable_lp(origin: OriginFor<T>) -> DispatchResultWithPostInfo {
```
shamil-gadelshin: Why do we need `DispatchResultWithPostInfo`? Can't we just rely on the correct benchmark?

open-junius: Please check my explanation in description, I should add it, thanks.

shamil-gadelshin: But how will you get the exact weight? The `dissolve_all_liquidity_providers` part will be combined with the rest of the computed weight. Won't it be more convenient to create a feature-gated extrinsic (wrapper around` dissolve_all_liquidity_providers`) specifically to measure this function, use it during our `runtime-benchmarks` workflow, and collect the weight data?

open-junius: Yes, we should just use the weight from benchmark. Actually, I have tested it. It makes the solution difficult to understand. will update it,.

shamil-gadelshin (dismissed):

pallets/subtensor/src/coinbase/root.rs
```diff
@@ -213,7 +213,8 @@ impl<T: Config> Pallet<T> {
         Self::finalize_all_subnet_root_dividends(netuid);
 
         // --- Perform the cleanup before removing the network.
-        T::SwapInterface::dissolve_all_liquidity_providers(netuid)?;
+        // Will handle it in dissolve network PR.
+        let _ = T::SwapInterface::dissolve_all_liquidity_providers(netuid);
```
l0r1s: We should ignore the weight but still propagate the error, I guess?

```suggestion
        // Will handle it in dissolve network PR.
        let _ = T::SwapInterface::dissolve_all_liquidity_providers(netuid)?;
```

l0r1s (changes requested):

pallets/subtensor/src/coinbase/root.rs
open-junius: Add the error handler here.

l0r1s: I get the impression it got modified at the wrong place? I see a propagation at the bottom in the `disable_lp` but this one is used in a lot of different places. Any reason why we don't want to propagate the error as it was before?

open-junius: Yes. I just modified the place after search in the code. I was think the whole dissolve_network will be updated so didn't realize it need change. just fix it.

l0r1s (approved):
Looks good!

sam0x17 (approved):

PR #2501, opened 2026-03-16
Add new subnet identity version with agent_docs_url

bdmason:
## Description

Adds a new field `agent_docs_url` to subnet identities.

## Related Issue(s)

NA

## Type of Change

- [ ] Documentation update

## Breaking Change

Any applications reading the old storage item would need to use the new one.

## Checklist

## Additional Notes


pallets/subtensor/src/migrations/migrate_subnet_identity_v3_to_v4.rs
```diff
@@ -0,0 +1,56 @@
...
+    HasMigrationRun::<T>::insert(&migration_name, true);
+    weight = weight.saturating_add(T::DbWeight::get().writes(1));
+
+    log::info!(
+        "Migration '{:?}' completed successfully. Migrated {migrated_count:?} subnet identities.",
+        String::from_utf8_lossy(&migration_name)
+    );
+
+    weight
+}
```
l0r1s, line 52: This should be done in place for V3.

pallets/subtensor/src/lib.rs
```diff
@@ -316,6 +315,34 @@ pub mod pallet {
         pub additional: Vec<u8>,
     }
 
+    ///  Struct for SubnetIdentitiesV4.
+    pub type SubnetIdentityOfV4 = SubnetIdentityV4;
```
l0r1s: This line is useless. What should be defined is `SubnetIdentityOf<T>` and it should be used across the codebase to avoid having to rename everytime there is a change.

You can get an example [here](https://github.com/opentensor/subtensor/blob/aae0ca738a0fe3d2da214b6428b67bd7d854d55b/pallets/crowdloan/src/lib.rs#L84)

pallets/subtensor/src/lib.rs
```diff
@@ -2153,11 +2150,16 @@ pub mod pallet {
...
-    /// --- MAP ( netuid ) --> SubnetIdentityOfV3
+    /// --- MAP ( netuid ) --> SubnetIdentityOfV3 (DEPRECATED for V4)
     #[pallet::storage]
     pub type SubnetIdentitiesV3<T: Config> =
         StorageMap<_, Blake2_128Concat, NetUid, SubnetIdentityOfV3, OptionQuery>;
 
+    /// --- MAP ( netuid ) --> SubnetIdentityOfV4
+    #[pallet::storage]
+    pub type SubnetIdentitiesV4<T: Config> =
+        StorageMap<_, Blake2_128Concat, NetUid, SubnetIdentityOfV4, OptionQuery>;
```
l0r1s: Not needed. We shouldn't be adding a new storage map for each version.

l0r1s (changes requested):

bdmason:
@l0r1s I've made the changes requested. I think this makes the storage item look more consistent, but it does not eliminate the downstream breaking change. It just makes it more subtle.

`SubnetIdentitiesV3` keeps the same storage name/key, but its value type changes from `SubnetIdentityV3` to `SubnetIdentityV4`. For typed consumers like Rust indexers using `subxt`, that is still a breaking
  runtime metadata change. Clients generated against the old metadata will continue to look up `SubnetIdentitiesV3`, but will attempt to decode the value as `SubnetIdentityV3`; after the upgrade/migration, those
  values are V4-encoded and decode will fail at runtime.

We still need to document this as a breaking change for anyone using `subxt`.

PR #2503, opened 2026-03-17
Auto CK root validators on subnet registration. 

shamil-gadelshin:
#### Summary

- Add `do_set_root_validators_for_subnet` function that automatically establishes parent-child relationships between all root network validators and a subnet owner's hotkey, enabling the subnet owner to inherit stake from root        
  validators and meet the `StakeThreshold` required to call `set_weights`
- Integrate the function into `do_register_network` so relationships are scheduled automatically at subnet creation time
- Each root validator becomes a parent of the subnet owner hotkey on the new subnet with full proportion (u64::MAX), using the existing `do_schedule_children` flow (with `PendingChildKeys` cooldown)

#### New tests

  - `test_root_children_enable_subnet_owner_set_weights` — verifies the function directly: subnet owner cannot set weights before, then can after root validator children are activated
  - `test_register_network_schedules_root_validators` — verifies end-to-end from register_network: pending children are automatically scheduled during subnet registration, activate after cooldown, and the subnet owner can then set weights
   via inherited stake


evgeny-s (dismissed):
The implementation looks good, but the fact that we add subnet hotkey to all the root validators as a child without them knowing this not good, I think. 

I think we might consider more options.
- We do it on subnet registration, but for a short period of time (N days)
- We require validator hotkey owners to confirm it during this period, otherwise it will be reverted back.

Another option for to providing the emission flow to the subnet, maybe it can be dedicated from the protocol,  instead of using validators stake.

TDB

pallets/subtensor/src/staking/set_children.rs
```diff
@@ -769,6 +769,67 @@ impl<T: Config> Pallet<T> {
...
+    /// * `netuid` - The subnet on which to establish relationships.
+    ///
+    /// # Returns
+    /// * `DispatchResult` - Ok if at least the setup completes; individual
+    ///   scheduling failures per validator are logged but do not abort the loop.
+    pub fn do_set_root_validators_for_subnet(netuid: NetUid) -> DispatchResult {
+        // Cannot set children on root network itself.
+        ensure!(
+            !netuid.is_root(),
+            Error::<T>::RegistrationNotPermittedOnRootSubnet
```
open-junius, line 789: need a different error type.

pallets/subtensor/src/staking/set_children.rs
```diff
@@ -769,6 +769,67 @@ impl<T: Config> Pallet<T> {
...
+        );
+
+        // Subnet must exist.
+        ensure!(Self::if_subnet_exist(netuid), Error::<T>::SubnetNotExists);
+
+        // Get the subnet owner hotkey.
+        let subnet_owner_hotkey =
+            SubnetOwnerHotkey::<T>::try_get(netuid).map_err(|_| Error::<T>::SubnetNotExists)?;
```
open-junius, line 797: need a different error type.

pallets/subtensor/src/staking/set_children.rs
shamil-gadelshin: Seems close enough. Also, I don't expect errors here at all because it's called only during subnet registration with concrete netuid. What do you suggest instead?

pallets/subtensor/src/staking/set_children.rs
shamil-gadelshin: Do we have a better one? It seems appropriate. We don't have subnet owner hotkey only when it doesn't exist. What do you suggest?

sam0x17 (dismissed):

gztensor (dismissed):

sam0x17:
this one is a bit contentious so we will need to continue to explore how much the community wants this (opendev/product call) to ensure it is coming from them

sam0x17:
merge conflict

sam0x17 (approved):

PR #2506, opened 2026-03-17
Allow clear reannouncement + clear scheduled CK swap

l0r1s:
## Description

- **Fix migration**: `migrate_coldkey_swap_scheduled_to_announcements` now cancels the `pallet_scheduler` tasks that were scheduled under the old system, preventing stale `swap_coldkey` calls from executing after the migration.
- **New extrinsic**: `clear_coldkey_swap_announcement` (call index 133) lets a coldkey voluntarily withdraw its own swap announcement once the reannouncement delay has elapsed.

## Details

The previous migration converted `ColdkeySwapScheduled` entries to `ColdkeySwapAnnouncements` but left the original scheduled calls alive in `pallet_scheduler::Agenda`. This PR adds a `translate_values` pass that tombstones those tasks (sets them to `None`) while preserving all other scheduler entries in the same blocks.

The new `clear_coldkey_swap_announcement` extrinsic enforces two guards before removing the announcement: the caller must be the owner of the announcement, and the current block must be at or past `announcement_block + ColdkeySwapAnnouncementDelay`. Disputed swaps are blocked at the transaction extension level as usual.

## Type of Change

- [ ] Documentation update

## Breaking Change

## Checklist

## Additional Notes

### DIFF
--- a/pallets/subtensor/src/macros/dispatches.rs
+++ b/pallets/subtensor/src/macros/dispatches.rs
@@ -241,7 +241,7 @@ mod dispatches {
 		.saturating_add(T::DbWeight::get().reads(10_u64))
 		.saturating_add(T::DbWeight::get().writes(2)), DispatchClass::Normal, Pays::No))]
         pub fn commit_weights(
-            origin: T::RuntimeOrigin,
+            origin: OriginFor<T>,
             netuid: NetUid,
             commit_hash: H256,
         ) -> DispatchResult {
@@ -275,7 +275,7 @@ mod dispatches {
 		.saturating_add(T::DbWeight::get().reads(7))
 		.saturating_add(T::DbWeight::get().writes(2)), DispatchClass::Normal, Pays::No))]
         pub fn commit_mechanism_weights(
-            origin: T::RuntimeOrigin,
+            origin: OriginFor<T>,
             netuid: NetUid,
             mecid: MechId,
             commit_hash: H256,
@@ -359,7 +359,7 @@ mod dispatches {
 		.saturating_add(T::DbWeight::get().reads(17_u64))
 		.saturating_add(T::DbWeight::get().writes(2)), DispatchClass::Normal, Pays::No))]
         pub fn reveal_weights(
-            origin: T::RuntimeOrigin,
+            origin: OriginFor<T>,
             netuid: NetUid,
             uids: Vec<u16>,
             values: Vec<u16>,
@@ -414,7 +414,7 @@ mod dispatches {
 		.saturating_add(T::DbWeight::get().reads(16))
 		.saturating_add(T::DbWeight::get().writes(2)), DispatchClass::Normal, Pays::No))]
         pub fn reveal_mechanism_weights(
-            origin: T::RuntimeOrigin,
+            origin: OriginFor<T>,
             netuid: NetUid,
             mecid: MechId,
             uids: Vec<u16>,
@@ -512,7 +512,7 @@ mod dispatches {
 		.saturating_add(T::DbWeight::get().reads(7_u64))
 		.saturating_add(T::DbWeight::get().writes(2)), DispatchClass::Normal, Pays::No))]
         pub fn commit_crv3_mechanism_weights(
-            origin: T::RuntimeOrigin,
+            origin: OriginFor<T>,
             netuid: NetUid,
             mecid: MechId,
             commit: BoundedVec<u8, ConstU32<MAX_CRV3_COMMIT_SIZE_BYTES>>,
@@ -572,7 +572,7 @@ mod dispatches {
 		.saturating_add(T::DbWeight::get().reads(17_u64))
 		.saturating_add(T::DbWeight::get().writes(2_u64)), DispatchClass::Normal, Pays::No))]
         pub fn batch_reveal_weights(
-            origin: T::RuntimeOrigin,
+            origin: OriginFor<T>,
             netuid: NetUid,
             uids_list: Vec<Vec<u16>>,
             values_list: Vec<Vec<u16>>,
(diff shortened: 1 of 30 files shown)


pallets/subtensor/src/macros/dispatches.rs
```diff
@@ -2561,13 +2558,36 @@ mod dispatches {
...
+        .saturating_add(T::DbWeight::get().reads(2))
+        .saturating_add(T::DbWeight::get().writes(1)))]
+        pub fn clear_coldkey_swap_announcement(origin: OriginFor<T>) -> DispatchResult {
+            let who = ensure_signed(origin)?;
+            let now = <frame_system::Pallet<T>>::block_number();
+
+            let Some((when, _)) = ColdkeySwapAnnouncements::<T>::get(who.clone()) else {
+                return Err(Error::<T>::ColdkeySwapAnnouncementNotFound.into());
+            };
+            let min_when = when.saturating_add(ColdkeySwapAnnouncementDelay::<T>::get());
```
shamil-gadelshin: The description mentions other setting: `... once the reannouncement delay has elapsed.`

shamil-gadelshin (approved):

pallets/subtensor/src/macros/dispatches.rs
```diff
@@ -2561,13 +2558,36 @@ mod dispatches {
...
         }
+
+        /// Clears a coldkey swap announcement if it has been announced and not disputed.
+        ///
+        /// The `ColdkeySwapCleared` event is emitted on successful clear.
+        #[pallet::call_index(133)]
+        #[pallet::weight(Weight::from_parts(10_000, 0)
+        .saturating_add(T::DbWeight::get().reads(2))
+        .saturating_add(T::DbWeight::get().writes(1)))]
+        pub fn clear_coldkey_swap_announcement(origin: OriginFor<T>) -> DispatchResult {
```
evgeny-s, line 2578: Shouldn't we add this extrinsic to the allowlist for `CheckColdkeySwap` dispatch guard? Otherwise, it can't ever be executed successfully, because if an announcement does exist, the guard will block it; otherwise, it will return `ColdkeySwapAnnouncementNotFound`.
Ideally, cover this scenario using an e2e test.

pallets/subtensor/src/macros/dispatches.rs
```diff
@@ -2561,13 +2558,36 @@ mod dispatches {
...
+            let who = ensure_signed(origin)?;
+            let now = <frame_system::Pallet<T>>::block_number();
+
+            let Some((when, _)) = ColdkeySwapAnnouncements::<T>::get(who.clone()) else {
+                return Err(Error::<T>::ColdkeySwapAnnouncementNotFound.into());
+            };
+            let min_when = when.saturating_add(ColdkeySwapAnnouncementDelay::<T>::get());
+            ensure!(now >= min_when, Error::<T>::ColdkeySwapClearTooEarly);
+
+            ColdkeySwapAnnouncements::<T>::remove(&who);
```
evgeny-s, line 2589: We might want to add a check if `ColdkeySwapDisputes` contains the actor's key. Otherwise, the attacker can remove the dispute and initiate the new swap, and finally perform it if the real owner didn't notice the new announcement.

pallets/subtensor/src/migrations/migrate_coldkey_swap_scheduled_to_announcements.rs
```diff
@@ -75,6 +101,23 @@ pub fn migrate_coldkey_swap_scheduled_to_announcements<T: Config>() -> Weight {
...
     );
 
+    pallet_scheduler::Agenda::<T>::translate_values::<Vec<Option<ScheduledOf<T>>>, _>(
+        |mut agenda| {
+            for schedule in agenda.iter_mut() {
+                // SwapColdkey call is guaranteed to be inlined and below 128 bytes.
+                if let Some(Bounded::Inline(data)) = schedule.as_ref().map(|s| s.call.clone()) {
+                    if deprecated::RuntimeCall::<T>::decode(&mut &data[..]).is_ok() {
+                        // Remove calls that decode as a SwapColdkey call
+                        schedule.take();
```
evgeny-s: We might want to add a `write` here to the weight meter as well.

pallets/subtensor/src/macros/dispatches.rs
l0r1s: > The description mentions other setting: `... once the reannouncement delay has elapsed.`

Correct, fixed. Thanks!

pallets/subtensor/src/macros/dispatches.rs
l0r1s: > Shouldn't we add this extrinsic to the allowlist for `CheckColdkeySwap` dispatch guard? Otherwise, it can't ever be executed successfully, because if an announcement does exist, the guard will block it; otherwise, it will return `ColdkeySwapAnnouncementNotFound`. Ideally, cover this scenario using an e2e test.

Absolutely, thanks for catching this! Fixed. For the E2E I will make them on a temporary branch on top of devnet ready including this code because the new setup is not on main branch yet.

pallets/subtensor/src/macros/dispatches.rs
l0r1s: > We might want to add a check if `ColdkeySwapDisputes` contains the actor's key. Otherwise, the attacker can remove the dispute and initiate the new swap, and finally perform it if the real owner didn't notice the new announcement.

The dispute status is checked on every dispatch by the `CheckColdkeySwap` dispatch guard. Only root is allowed to remove a dispute using the `reset_coldkey_swap` which clears both announcement and dispute for an account so this is safe

What can happens though is that the attacker can wait for the reannouncement to have passed, clear the announcement and reannounce. In this case, he will still need to wait 5 days.

pallets/subtensor/src/migrations/migrate_coldkey_swap_scheduled_to_announcements.rs
l0r1s: > We might want to add a `write` here to the weight meter as well.

Correct! Thanks for the catch. Fixed.

evgeny-s (approved):
LGTM

JohnReedV (approved):

gztensor (approved):

PR #2509, opened 2026-03-18
avoid zero value insert

open-junius:
## Related Issue(s)

- Closes #2398 

## Type of Change

- [ ] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] Documentation update
- [ ] Other (please describe):

## Breaking Change

If this PR introduces a breaking change, please provide a detailed description of the impact and the migration path for existing applications.

## Checklist

- [ ] I have performed a self-review of my own code
- [ ] I have commented my code, particularly in hard-to-understand areas
- [ ] I have run `./scripts/fix_rust.sh` to ensure my code is formatted and linted correctly
- [ ] I have made corresponding changes to the documentation
- [ ] My changes generate no new warnings
- [ ] I have added tests that prove my fix is effective or that my feature works
- [ ] New and existing unit tests pass locally with my changes
- [ ] Any dependent changes have been merged and published in downstream modules

## Screenshots (if applicable)

Please include any relevant screenshots or GIFs that demonstrate the changes made.

## Additional Notes

Please provide any additional information or context that may be helpful for reviewers.

PR #2489, opened 2026-03-10
Extend node with build-patched-spec subcommand

ales-otf:
## Description

Introduces `build-patched-spec` subcommand, that does next:

- syncs the state of the given network
- exports it into a temporary chainspec
- patches the chainspec to make a test network chainspec from it (applying [this](https://gist.github.com/l0r1s/2d7de2c3e215fd576e3933c488ec1262) patch by @l0r1s )

So basically it produces a patched clone of a given network for testing. So you can test on mainnet clone, for example.

It can be used like this:

```shell
target/release/node-subtensor build-patched-spec --base-path target/mainnet-clone --chain chainspecs/raw_spec_finney.json --bootnodes /dns/bootnode.finney.chain.opentensor.ai/tcp/30333/ws/p2p/12D3KooWRwbMb85RWnT8DSXSYMWQtuDwh4LJzndoRrTDotTR5gDC --output target/mainnet-clone-chainspec.json
```

Then you normally run a testing node with produced chainspec like this:

```shell
target/release/node-subtensor --base-path target/clone --chain target/mainnet-clone-chainspec.json --database paritydb --force-authoring --alice --validator --unsafe-force-node-key-generation
```

The PR introduces a `history_backfill` flag to skip the [gap feeling](https://paritytech.github.io/polkadot-sdk/master/sc_consensus/block_import/struct.BlockImportParams.html#structfield.create_gap) phase after the initial sync. That phase is needed for the historical data, but for the testing purposes we can skip it. It's `keep` by default, so it doesn't affect current node behavior. It's `skip` by default for `build-patched-spec` to speed up the process.

Also unlike normal node default to rocksdb, `build-patched-spec` is default to `paritydb`. And unlike the patch, `build-patched-spec` produces chainspec with only single authority (**alice**) by default. This is more convenient and lightweight for local runs. But you can specify other conventional authorities using flags (`--alice`, `--bob`, `--charlie`) - any combination of them is allowed.

```shell
❯ target/release/node-subtensor build-patched-spec -h
Build a patched clone chainspec by syncing state, exporting raw state, and applying test patch

Usage: node-subtensor build-patched-spec [OPTIONS] --chain <CHAIN> --base-path <PATH> --output <FILE>

Options:
      --chain <CHAIN>                        Chain spec identifier or path (same semantics as `--chain`)
      --base-path <PATH>                     Base path used for syncing and state export
      --output <FILE>                        Output file path for the final patched chainspec JSON
[...]

PR #2513, opened 2026-03-18
fix: Prevent and clean up zero-valued entries in Alpha storage maps during hotkey and coldkey swaps

### LINKED ISSUE #2398: Alpha and hotkey storage maps accumulate persistent zero-valued entries
# Alpha and hotkey storage maps accumulate persistent zero-valued entries

## Summary
Coldkey and hotkey swap logic inserts zero-valued entries into Alpha-related storage maps, which persist indefinitely.

## Problem
- Zero-valued entries are inserted without checks.
- Entries persist unless explicitly cleaned by stake activity.
- A significant portion of Alpha storage contains zero values.

## Evidence
- ~52% of Alpha entries are zero-valued.
- A single coldkey swap produced ~18k zero entries.
- Fetching Alpha requires ~150MB of keys.

## Expected Outcome
- Zero-valued entries should not be inserted.
- Existing zero entries should be purged during a runtime upgrade.
- Storage size and query performance should improve.

## Context
Filed at the request of Rhef for tracking and prioritization.

bittoby:
## Description

Prevent and clean up zero-valued entries in Alpha storage maps during hotkey and coldkey swaps.

Coldkey and hotkey swap logic was inserting zero-valued entries into `Alpha`, `TotalHotkeyShares`, `TotalHotkeyAlphaLastEpoch`, and `AlphaDividendsPerSubnet` without checking if the value was zero first. These entries persist indefinitely and bloat storage - a single coldkey swap could produce ~18k zero entries, and roughly 52% of Alpha entries on-chain are zero-valued.

This PR adds zero-guards to all insertion points in `swap_coldkey.rs` and `swap_hotkey.rs`, and includes a one-time migration to purge existing zero entries from the four affected storage maps.

## Related Issue(s)

- Closes #2398

## Type of Change

## Breaking Change

No breaking changes.

## Checklist

## Additional Notes

The migration collects the full `Alpha` map into memory before removing zero entries. This is intentional - mutating a Substrate storage map while iterating over it is unsafe. The other three maps are smaller and iterated directly. All four maps use `HasMigrationRun` to ensure idempotency.

### DIFF
--- a/pallets/subtensor/src/lib.rs
+++ b/pallets/subtensor/src/lib.rs
@@ -2475,6 +2475,11 @@ pub mod pallet {
     #[pallet::storage]
     pub type HasMigrationRun<T: Config> = StorageMap<_, Identity, Vec<u8>, bool, ValueQuery>;
 
+    /// --- Tracks the current phase of the zero-alpha multi-block cleanup.
+    /// 0 = inactive/complete, 1-4 = active phases (Alpha, TotalHotkeyShares, etc.)
+    #[pallet::storage]
+    pub type ZeroAlphaCleanupPhase<T: Config> = StorageValue<_, u8, ValueQuery>;
+
     /// Default value for pending childkey cooldown (settable by root).
     /// Uses the same value as DefaultPendingCooldown for consistency.
     #[pallet::type_value]
--- a/pallets/subtensor/src/migrations/migrate_remove_zero_alpha.rs
+++ b/pallets/subtensor/src/migrations/migrate_remove_zero_alpha.rs
@@ -1,0 +1,216 @@
+use super::*;
+use frame_support::{traits::Get, weights::Weight};
+use log;
+use scale_info::prelude::string::String;
+
+/// The migration name used for the `HasMigrationRun` guard.
+const MIGRATION_NAME: &[u8] = b"migrate_remove_zero_alpha_v2";
+
+/// Called from `on_runtime_upgrade`. Schedules the cleanup by setting phase = 1
+/// if the migration hasn't run yet. This is O(1) — no iteration.
+pub fn migrate_remove_zero_alpha<T: Config>() -> Weight {
+    let migration_name = MIGRATION_NAME.to_vec();
+    let mut weight = T::DbWeight::get().reads(1);
+
+    if HasMigrationRun::<T>::get(&migration_name) {
+        log::info!(
+            "Migration '{}' already completed. Skipping.",
+            String::from_utf8_lossy(&migration_name)
+        );
+        return weight;
+    }
+
+    // Schedule the cleanup to run in on_idle by setting phase to 1
+    ZeroAlphaCleanupPhase::<T>::put(1u8);
+    weight = weight.saturating_add(T::DbWeight::get().writes(1));
+
+    log::info!(
+        "Migration '{}' scheduled. Will clean up zero entries via on_idle.",
+        String::from_utf8_lossy(&migration_name)
+    );
+
+    weight
+}
+
+/// Called from `on_idle` each block. Uses `remaining_weight` to dynamically
+/// bound how many entries to process. Stays on the same phase until all entries
+/// in that map are cleaned, then advances to the next phase.
+///
+/// Phases:
+///   0 = inactive/complete
+///   1 = cleaning Alpha
+///   2 = cleaning TotalHotkeyShares
+///   3 = cleaning TotalHotkeyAlphaLastEpoch
+///   4 = cleaning AlphaDividendsPerSubnet
+pub fn on_idle_remove_zero_alpha<T: Config>(remaining_weight: Weight) -> Weight {
+    let phase = ZeroAlphaCleanupPhase::<T>::get();
+
+    // Phase 0 means not active or already completed
+    if phase == 0 {
+        return Weight::zero();
+    }
+
+    // Minimum weight needed: 1 read (phase) + at least one iteration (read + write)
+    let min_weight = T::DbWeight::get().reads_writes(2, 1);
+    if remaining_weight.ref_time() < min_weight.ref_time() {
+        return Weight::zero();
+    }
+
+    let mut weight = T::DbWeight::get().reads(1); // reading phase
+
+    // Budget for batch work = remaining_weight minus overhead (phase read + phase write)
+    let overhead = T::DbWeight::get().reads_writes(1, 1);
+    let budget = remaining_weight.saturating_sub(overhead);
+
+    match phase {
+        1 => {
+            let (consumed, removed, done) = clean_alpha_batch::<T>(budget);
+            weight = weight.saturating_add(consumed);
+            log::info!(
+                "Zero-alpha cleanup phase 1 (Alpha): removed {removed} zero entries this batch. Done: {done}"
+            );
+            if done {
+                ZeroAlphaCleanupPhase::<T>::put(2u8);
+                weight = weight.saturating_add(T::DbWeight::get().writes(1));
+            }
+        }
+        2 => {
+            let (consumed, removed, done) = clean_total_hotkey_shares_batch::<T>(budget);
+            weight = weight.saturating_add(consumed);
+            log::info!(
+                "Zero-alpha cleanup phase 2 (TotalHotkeyShares): removed {removed} zero entries this batch. Done: {done}"
+            );
+            if done {
+                ZeroAlphaCleanupPhase::<T>::put(3u8);
+                weight = weight.saturating_add(T::DbWeight::get().writes(1));
+            }
+        }
+        3 => {
+            let (consumed, removed, done) = clean_total_hotkey_alpha_last_epoch_batch::<T>(budget);
+            weight = weight.saturating_add(consumed);
+            log::info!(
+                "Zero-alpha cleanup phase 3 (TotalHotkeyAlphaLastEpoch): removed {removed} zero entries this batch. Done: {done}"
+            );
+            if done {
+                ZeroAlphaCleanupPhase::<T>::put(4u8);
+                weight = weight.saturating_add(T::DbWeight::get().writes(1));
+            }
+        }
+        4 => {
+            let (consumed, removed, done) = clean_alpha_dividends_per_subnet_batch::<T>(budget);
+            weight = weight.saturating_add(consumed);
+            log::info!(
+                "Zero-alpha cleanup phase 4 (AlphaDividendsPerSubnet): removed {removed} zero entries this batch. Done: {done}"
+            );
+            if done {
+                // All phases complete — mark migration as done
+                HasMigrationRun::<T>::insert(MIGRATION_NAME.to_vec(), true);
+                ZeroAlphaCleanupPhase::<T>::put(0u8);
+                weight = weight.saturating_add(T::DbWeight::get().writes(2));
+                log::info!("Zero-alpha cleanup: All phases complete. Migration marked as done.");
+            }
+        }
+        _ => {
+            // Unknown phase, reset
+            ZeroAlphaCleanupPhase::<T>::put(0u8);
+            weight = weight.saturating_add(T::DbWeight::get().writes(1));
+        }
+    }
+
+    weight
+}
+
+/// Remove zero-valued entries from Alpha, bounded by weight budget.
+/// Returns (weight_consumed, entries_removed, is_done).
+fn clean_alpha_batch<T: Config>(budget: Weight) -> (Weight, u64, bool) {
+    let read_cost = T::DbWeight::get().reads(1);
+    let write_cost = T::DbWeight::get().writes(1);
+    let per_entry_max = read_cost.saturating_add(write_cost);
+    let mut weight = Weight::zero();
+    let mut removed = 0u64;
+
+    for ((hotkey, coldkey, netuid), value) in Alpha::<T>::iter() {
+        // Stop if not enough budget for one more entry (read + potential write)
+        if weight.saturating_add(per_entry_max).any_gt(budget) {
+            return (weight, removed, false);
+        }
+        weight = weight.saturating_add(read_cost);
+        if value == 0 {
+            Alpha::<T>::remove((hotkey, coldkey, netuid));
+            weight = weight.saturating_add(write_cost);
+            removed = removed.saturating_add(1);
+        }
+    }
+
+    // Iterator exhausted — phase is done
+    (weight, removed, true)
+}
+
+/// Remove zero-valued entries from TotalHotkeyShares, bounded by weight budget.
+fn clean_total_hotkey_shares_batch<T: Config>(budget: Weight) -> (Weight, u64, bool) {
+    let read_cost = T::DbWeight::get().reads(1);
+    let write_cost = T::DbWeight::get().writes(1);
+    let per_entry_max = read_cost.saturating_add(write_cost);
+    let mut weight = Weight::zero();
+    let mut removed = 0u64;
+
+    for (hotkey, netuid, value) in TotalHotkeyShares::<T>::iter() {
+        if weight.saturating_add(per_entry_max).any_gt(budget) {
+            return (weight, removed, false);
+        }
+        weight = weight.saturating_add(read_cost);
+        if value == 0 {
+            TotalHotkeyShares::<T>::remove(hotkey, netuid);
+            weight = weight.saturating_add(write_cost);
+            removed = removed.saturating_add(1);
+        }
+    }
+
+    (weight, removed, true)
+}
+
+/// Remove zero-valued entries from TotalHotkeyAlphaLastEpoch, bounded by weight budget.
+fn clean_total_hotkey_alpha_last_epoch_batch<T: Config>(budget: Weight) -> (Weight, u64, bool) {
+    let read_cost = T::DbWeight::get().reads(1);
+    let write_cost = T::DbWeight::get().writes(1);
+    let per_entry_max = read_cost.saturating_add(write_cost);
+    let mut weight = Weight::zero();
+    let mut removed = 0u64;
+
+    for (hotkey, netuid, value) in TotalHotkeyAlphaLastEpoch::<T>::iter() {
+        if weight.saturating_add(per_entry_max).any_gt(budget) {
+            return (weight, removed, false);
+        }
+        weight = weight.saturating_add(read_cost);
+        if value.is_zero() {
+            TotalHotkeyAlphaLastEpoch::<T>::remove(hotkey, netuid);
+            weight = weight.saturating_add(write_cost);
+            removed = removed.saturating_add(1);
+        }
+    }
+
+    (weight, removed, true)
+}
+
+/// Remove zero-valued entries from AlphaDividendsPerSubnet, bounded by weight budget.
+fn clean_alpha_dividends_per_subnet_batch<T: Config>(budget: Weight) -> (Weight, u64, bool) {
+    let read_cost = T::DbWeight::get().reads(1);
+    let write_cost = T::DbWeight::get().writes(1);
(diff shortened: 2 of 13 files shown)


bittoby:
@open-junius Could you please review this PR? Welcome to any feedbacks. 
thank you

open-junius:
change the target branch as devnet-ready, not main.

bittoby:
I changed target branch. Thanks! Ready to review

pallets/subtensor/src/swap/swap_hotkey.rs
```diff
@@ -364,10 +364,11 @@ impl<T: Config> Pallet<T> {
...
-
-            TotalHotkeyAlpha::<T>::mutate(new_hotkey, netuid, |value| {
-                *value = value.saturating_add(alpha)
-            });
+            if !alpha.is_zero() {
+                TotalHotkeyAlpha::<T>::mutate(new_hotkey, netuid, |value| {
+                    *value = value.saturating_add(alpha)
+                });
+            }
             weight.saturating_accrue(T::DbWeight::get().reads_writes(1, 1));
```
open-junius: update the weight accordingly.

pallets/subtensor/src/swap/swap_hotkey.rs
```diff
@@ -387,7 +388,9 @@ impl<T: Config> Pallet<T> {
 
             let total_old_plus_new_shares =
                 total_new_shares.add(&total_old_shares).unwrap_or_default();
-            TotalHotkeySharesV2::<T>::insert(new_hotkey, netuid, total_old_plus_new_shares);
+            if !total_old_plus_new_shares.is_zero() {
+                TotalHotkeySharesV2::<T>::insert(new_hotkey, netuid, total_old_plus_new_shares);
+            }
             weight.saturating_accrue(T::DbWeight::get().writes(3));
```
open-junius: the same as last comment

pallets/subtensor/src/swap/swap_hotkey.rs
```diff
@@ -397,7 +400,9 @@ impl<T: Config> Pallet<T> {
         // IsNetworkMember( hotkey, netuid ) -> bool -- is the hotkey a subnet member.
         let is_network_member: bool = IsNetworkMember::<T>::get(old_hotkey, netuid);
         IsNetworkMember::<T>::remove(old_hotkey, netuid);
-        IsNetworkMember::<T>::insert(new_hotkey, netuid, is_network_member);
+        if is_network_member {
+            IsNetworkMember::<T>::insert(new_hotkey, netuid, is_network_member);
+        }
         weight.saturating_accrue(T::DbWeight::get().reads_writes(1, 2));
```
open-junius: the same as last comment

pallets/subtensor/src/swap/swap_hotkey.rs
```diff
@@ -523,22 +528,25 @@ impl<T: Config> Pallet<T> {
...
-            );
+            let combined_alpha_last_epoch = old_alpha.saturating_add(new_total_hotkey_alpha);
+            if !combined_alpha_last_epoch.is_zero() {
+                TotalHotkeyAlphaLastEpoch::<T>::insert(
+                    new_hotkey,
+                    netuid,
+                    combined_alpha_last_epoch,
+                );
+            }
             weight.saturating_accrue(T::DbWeight::get().reads_writes(2, 2));
```
open-junius: the same as last comment

pallets/subtensor/src/swap/swap_hotkey.rs
```diff
@@ -523,22 +528,25 @@ impl<T: Config> Pallet<T> {
...
-                netuid,
-                new_hotkey,
-                old_hotkey_alpha_dividends.saturating_add(new_hotkey_alpha_dividends),
-            );
+            let combined_dividends =
+                old_hotkey_alpha_dividends.saturating_add(new_hotkey_alpha_dividends);
+            if !combined_dividends.is_zero() {
+                AlphaDividendsPerSubnet::<T>::insert(netuid, new_hotkey, combined_dividends);
+            }
             weight.saturating_accrue(T::DbWeight::get().reads_writes(2, 2));
```
open-junius: the same as last comment

pallets/subtensor/src/swap/swap_hotkey.rs
```diff
@@ -574,11 +582,14 @@ impl<T: Config> Pallet<T> {
...
-                    );
+                    // Only insert combined alpha if non-zero to avoid storage bloat.
+                    let combined_alpha = alpha.saturating_add(new_alpha);
+                    if combined_alpha != 0 {
+                        AlphaV2::<T>::insert(
+                            (new_hotkey, &coldkey, netuid),
+                            SafeFloat::from(combined_alpha),
+                        );
+                    }
                     weight.saturating_accrue(T::DbWeight::get().reads_writes(1, 2));
```
open-junius: the same as last comment

pallets/subtensor/src/swap/swap_hotkey.rs
```diff
@@ -602,10 +613,11 @@ impl<T: Config> Pallet<T> {
...
-                    AlphaV2::<T>::insert(
-                        (new_hotkey, &coldkey, netuid),
-                        alpha.add(&new_alpha_v2).unwrap_or_default(),
-                    );
+                    // Only insert combined alpha if non-zero to avoid storage bloat.
+                    let combined_alpha_v2 = alpha.add(&new_alpha_v2).unwrap_or_default();
+                    if !combined_alpha_v2.is_zero() {
+                        AlphaV2::<T>::insert((new_hotkey, &coldkey, netuid), combined_alpha_v2);
+                    }
                     weight.saturating_accrue(T::DbWeight::get().reads_writes(1, 2));
```
open-junius: the same as last comment

open-junius:
please check https://github.com/opentensor/subtensor/pull/2509/changes, which also prevent some zero value insert. Could you include these storage into the PR.

bittoby:
@open-junius Thanks for your feedback! I updated all following your feedback. Could you please review again?

open-junius:
please merge the latest devnet-ready, will fix the e2e test.

bittoby:
@open-junius I merged the latest devnet-ready. Could you take a look once more?
thank you

bittoby:
@open-junius All e2e tests passed. thank you. The benchmark testing issue is not relate to my change.

open-junius (dismissed):

.github/workflows/check-node-compat.yml
```diff
@@ -22,8 +22,10 @@ jobs:
     strategy:
       matrix:
         version:
```
evgeny-s, line 24: not relevant change, please revert

pallets/subtensor/src/migrations/migrate_remove_zero_alpha.rs
```diff
@@ -0,0 +1,98 @@
...
+        "Running migration '{}'",
+        String::from_utf8_lossy(&migration_name)
+    );
+
+    // ------------------------------
+    // Step 1: Remove zero entries in Alpha (StorageNMap, value type: U64F64)
+    // ------------------------------
+    let mut removed_alpha = 0u64;
+
+    for ((hotkey, coldkey, netuid), value) in Alpha::<T>::iter() {
```
evgeny-s, line 156: How much time does this migration take? Would be great to test it on the mainnet clone environment.

evgeny-s: Looks like a heavy one.

evgeny-s (commented):
Please merge latest devnet-ready and add similar TS e2e tests as well.

.github/workflows/check-node-compat.yml
bittoby: Actually I have changed this because syntax error. I will revert if you don't want this

pallets/subtensor/src/migrations/migrate_remove_zero_alpha.rs
bittoby: > Looks like a heavy one.

Thanks @evgeny-s You got correct point!
I have tested against Finney mainnet via try-runtime. The state download alone hit 8.2M+ keys after about 70 minutes before timing out and that's just scraping, not even running the migration yet.

This confirms the migration is too heavy for a single block. I'll convert it to a multi-block SteppedMigration that processes entries in batches.

bittoby:
@evgeny-s @open-junius 

Converting to `SteppedMigration` would require adding `pallet-migrations` to the runtime since `MultiBlockMigrator` is currently `()`  That's a bigger scope change.

I think the cleanest approach is to **drop the migration from this PR** and just ship the zero-value insert guards (which prevent new zeros from being created). The purge of existing zeros can be handled properly in a separate PR with the right multi-block infrastructure.  I’m planning to take care of that follow-up PR too after this one.

Let me know what you'd prefer.

bittoby:
@evgeny-s @open-junius did you check my description? Should I have to update my PR?

pallets/subtensor/src/migrations/migrate_remove_zero_alpha.rs
evgeny-s: You can test it using the approach from [this](https://github.com/opentensor/subtensor/pull/2489) PR.

pallets/subtensor/src/staking/claim_root.rs
```diff
@@ -263,8 +263,10 @@ impl<T: Config> Pallet<T> {
                     .saturating_to_num(),
             );
 
-            // Set the new root claimed value.
-            RootClaimed::<T>::insert((netuid, hotkey, coldkey), new_root_claimed);
+            // Set the new root claimed value, removing if zero to avoid storage bloat.
```
evgeny-s: The comment is saying "removing", but I don't see it here.

evgeny-s: But below we remove it. Fix it for consistency

evgeny-s:
> @evgeny-s @open-junius did you check my description? Should I have to update my PR?

I would rather keep it with the migration. Please test it first on the mainnet clone, and use multi-block migration if needed. Let's see other opinions on this regard.
Also, it would be great to add several TS tests for this PR.

bittoby:
> > @evgeny-s @open-junius did you check my description? Should I have to update my PR?
> 
> I would rather keep it with the migration. Please test it first on the mainnet clone, and use multi-block migration if needed. Let's see other opinions on this regard. Also, it would be great to add several TS tests for this PR.

I absolute agree with you. If possible, I want to keep migration in this PR, too. Now I am testing locally.  After finishing, I will share result here. Thanks @evgeny-s

bittoby:
@evgeny-s 

Updated:
- Converted migration to multi-block `on_idle` approach - `on_runtime_upgrade` is O(1), cleanup runs one map per block across 4 phases
- Fixed `claim_root.rs` - added missing `else { remove }` branch for consistency
- Tested against Finney mainnet clone via `build-patched-spec` - node runs and produces blocks normally with the migration code
- Added TS e2e test for zero-alpha prevention
- Reverted unrelated `check-node-compat.yml` change

ts-tests/suites/zombienet_staking/12-zero-alpha-prevention.test.ts
```diff
@@ -0,0 +1,95 @@
...
+
+                // Remove all stake
+                const rawStake = await getStakeRaw(api, hotkey.address, coldkey.address, netuid);
+                await removeStake(api, coldkey, hotkey.address, netuid, rawStake);
+
+                // After full removal, the Alpha entry should be zero or removed
+                const stakeAfterRemove = await getStake(api, hotkey.address, coldkey.address, netuid);
+
+                // Stake should be zero or very close to zero (owner minimum may be retained)
+                log(`Stake after full remove: ${stakeAfterRemove}`);
```
open-junius: If we can't check in the code. maybe we don't need this test case. since 0 is default value, we can't tell the difference. It seems we need lower level api to check if the key exists or not.

open-junius: In my opinion, it should be enough you already verify them in rust unit test. no need to add it in e2e.

bittoby:
@open-junius Thanks for your feedback. I removed e2e test. Please review again.

bittoby:
@open-junius Could you tell me why `assert-clean-merges` testing failed?

open-junius:
> @open-junius Could you tell me why `assert-clean-merges` testing failed?

You can ignore it. we need merge the main back to devnet-ready sometimes if there is any hotfix in main branch.

open-junius:
> These testing failure doesn't seem to relate to my change. Please check the log. @open-junius
I see, it is common issue in devnet-ready branch. I noticed you split the data cleanup into phases.
Could you follow the method recommended by @evgeny-s to get the amount and time needed to run migration?

bittoby:
I have updated the migration script more correctly:
It now uses `remaining_weight` to dynamically limit how many entries are processed per block instead of a fixed batch size. This helps prevent block timeouts on large maps.

bittoby:
Tested on Finney mainnet clone via `build-patched-spec`:
- **Alpha zero entries:** ~580,000
- **Time:** ~14 min (70 blocks) to remove 99% of them

No OOM or block timeout issues.

bittoby:
@open-junius @evgeny-s I'd like you review once more ^^. It works well for me

open-junius:
> I have updated the migration script more correctly: It now uses `remaining_weight` to dynamically limit how many entries are processed per block instead of a fixed batch size. This helps prevent block timeouts on large maps.

It is great, let me review it again.

open-junius:
> Tested on Finney mainnet clone via `build-patched-spec`:
> 
> * **Alpha zero entries:** ~580,000
> * **Time:** ~14 min (70 blocks) to remove 99% of them
> 
> No OOM or block timeout issues.

Thanks for collect the data, it is vaulable.

pallets/subtensor/src/migrations/migrate_remove_zero_alpha.rs
```diff
@@ -0,0 +1,216 @@
...
+use frame_support::{traits::Get, weights::Weight};
+use log;
+use scale_info::prelude::string::String;
+
+/// The migration name used for the `HasMigrationRun` guard.
+const MIGRATION_NAME: &[u8] = b"migrate_remove_zero_alpha_v2";
+
+/// Called from `on_runtime_upgrade`. Schedules the cleanup by setting phase = 1
+/// if the migration hasn't run yet. This is O(1) — no iteration.
+pub fn migrate_remove_zero_alpha<T: Config>() -> Weight {
```
open-junius, line 11: After the actual cleanup executed in the on_idle, I am wondering if we really need the migration part.
For instance, we can set the ZeroAlphaCleanupPhase init value as 1 in storage.
then on_idle start to handle it until the value go to 0. 
Or use different values for different steps, make it clearer. 
What's your opinion?

bittoby: Good point. We could simplify by setting the default to 1, but the issue is: after cleanup completes and sets phase=0, a node restart or fresh sync would read the default (1) again and re-trigger cleanup. The current approach uses `HasMigrationRun` as a permanent guard to prevent this.

bittoby: > Good point. We could simplify by setting the default to 1, but the issue is: after cleanup completes and sets phase=0, a node restart or fresh sync would read the default (1) again and re-trigger cleanup. The current approach uses `HasMigrationRun` as a permanent guard to prevent this.

@open-junius what do you think about it? Still should i need to update?

open-junius: Do you mean dev mod in local env, I think it never happened in real network, right?

bittoby: @open-junius I see your point. Your word is right. On a real network, nodes always sync from existing chain, so the storage value already be set correctly and default value wouldn't matter.

The reason I kept the `on_runtime_upgrade` + `HasMigrationRun` approach is just for consistency. Every other migration in the codebase follows this pattern.  It's just more defensive. The `HasMigrationRun` guard is extra safety that doesn't hurt anything.

But if you'd prefer to simplify, I will update. Pls let me know. Thanks

open-junius: It is fine for me. you can keep it if you prefer, no worry.

bittoby:
@sam0x17 @open-junius Could you re-run CI testing?

bittoby:
Additionally, I added `AlphaV2` and `TotalHotkeySharesV2` to migration cleanup and revert out-of-scope `LastRateLimitedBlock` change. 

Ready for review now. @open-junius @evgeny-s @sam0x17

pallets/subtensor/src/staking/set_children.rs
```diff
@@ -412,15 +412,11 @@ impl<T: Config> Pallet<T> {
...
+            Self::set_childkeys(parent.clone(), netuid, ck);
             weight.saturating_accrue(T::DbWeight::get().reads_writes(1, 1));
         }
         //    2c) Clear direct maps of old_hotkey
-        ChildKeys::<T>::insert(
-            old_hotkey.clone(),
-            netuid,
-            Vec::<(u64, T::AccountId)>::new(),
-        );
+        ChildKeys::<T>::remove(old_hotkey.clone(), netuid);
```
evgeny-s, line 419: We might probably can also use `Self::set_childkeys` here.

pallets/subtensor/src/lib.rs
```diff
@@ -2475,6 +2475,11 @@ pub mod pallet {
     #[pallet::storage]
     pub type HasMigrationRun<T: Config> = StorageMap<_, Identity, Vec<u8>, bool, ValueQuery>;
 
+    /// --- Tracks the current phase of the zero-alpha multi-block cleanup.
+    /// 0 = inactive/complete, 1-4 = active phases (Alpha, TotalHotkeyShares, etc.)
```
evgeny-s, line 2479: 1-6? update comment plz

evgeny-s (approved):
Added several comments. 
Did we do any measurements with the mainnet clone? How much time does it take to migrate the whole thing? Do we have any block delay?
Could you post the logs here from the migration test?

Also:
- update your branch from the latest devnet-ready, there is some fix for the TS tests 
- bump the runtime version as well.

Otherwise, LGTM.

bittoby:
> Tested on Finney mainnet clone via `build-patched-spec`:
> 
> * **Alpha zero entries:** ~580,000
> * **Time:** ~14 min (70 blocks) to remove 99% of them
> 
> No OOM or block timeout issues.

@evgeny-s can you see this?

pallets/subtensor/src/migrations/migrate_remove_zero_alpha.rs
```diff
@@ -0,0 +1,284 @@
...
+/// Remove zero-valued entries from Alpha, bounded by weight budget.
+/// Returns (weight_consumed, entries_removed, is_done).
+fn clean_alpha_batch<T: Config>(budget: Weight) -> (Weight, u64, bool) {
+    let read_cost = T::DbWeight::get().reads(1);
+    let write_cost = T::DbWeight::get().writes(1);
+    let per_entry_max = read_cost.saturating_add(write_cost);
+    let mut weight = Weight::zero();
+    let mut removed = 0u64;
+
+    for ((hotkey, coldkey, netuid), value) in Alpha::<T>::iter() {
```
evgeny-s, line 156: BTW, we don't have any cursors here. Theoretically, if there is not enough space in the block, we might be stuck on the same stage until we have it.

sam0x17:
thanks, this is not needed after the recent rework of hk+ck swap by @gregzaitsev

PR #2517, opened 2026-03-19
Rework benchmarks

lorismoulin:
## Summary

Converts the benchmark and weight system to the standard FRAME `WeightInfo` pattern. This lays the groundwork for using parameterized weights where extrinsic cost scales with input size (e.g., number of contributors, batch size), which was not possible with hardcoded values.

**Before:** Extrinsics had `Weight::from_parts(X, 0)` hardcoded directly in `#[pallet::weight(...)]` annotations. CI parsed benchmark stdout and patched values in source files.

**After:** Each pallet has a `weights.rs` file, a `WeightInfo` trait, and extrinsics reference `T::WeightInfo::fn_name()`. CI uses a syn-based Rust tool to compare old vs new weight values with a threshold.

Note: we generate weights using `--runtime` + `--genesis-builder` flags (not `--chain`) because the runtime uses custom host functions that require the full node binary.

### Pallet changes

- **pallet_subtensor** — 51 extrinsics converted to `T::WeightInfo::fn()`, 18 without benchmarks kept hardcoded. Removed `#[benchmark(extra)]` from `register_leased_network` and `terminate_lease` so they run in normal benchmark passes.
- **pallet_admin_utils** — 55 extrinsics converted, 19 kept hardcoded
- **pallet_drand** — all 3 extrinsics converted
- **pallet_shield** — all 2 extrinsics converted
- **pallet_commitments** — all 2 extrinsics converted, stale `weights.rs` fixed (`set_rate_limit` renamed to `set_max_space`), benchmark test suite restored
- **pallet_crowdloan** — `weights.rs` regenerated (cleanup, no functional change)

For each pallet: added `pub mod weights`, `type WeightInfo` in Config, wired `SubstrateWeight<Runtime>` in runtime, added `type WeightInfo = ()` to all test mocks.

### Benchmark fixes

- **pallet_subtensor_swap** — fixed `remove_liquidity` (`HotKeyAccountNotExists`) and `modify_position` (`SubtokenDisabled`) benchmark setup. Added benchmarks for `add_liquidity`, `toggle_user_liquidity`, and `disable_lp`.
- **pallet_registry** — fixed `set_identity` and `clear_identity` benchmarks failing with `FundsUnavailable` (replaced `set_balance(max_value)` with `mint_into`, set identity on caller instead of separate account).

### New tooling (`support/weight-tools/`)

Lightweight crate (~4s build, no runtime dependency):

- **`weight-compare`** — parses two `weights.rs` files with `syn`, compares base weights (threshold-based), reads/writes (exact), and per-component slopes for parameterized weights. Used by CI.

### Scripts

- **`scripts/benchmark_all.sh`** — rewritten to generate `weights.rs` via `--output`/`--template`, supports single-pallet runs and `SKIP_BUILD=1`
- **`scripts/benchmark_action.sh`** — rewritten to use `weight-compare` with 40% drift threshold
- Removed 4 redundant per-pallet `scripts/benchmark.sh` files

### CI

- **`run-benchmarks.yml`** — cleaned up, extracted repeated skip-label check into `.github/scripts/check-skip-label.sh`, added `weight-compare` build step

### Other

- `.maintain/frame-weight-template.hbs` synced with upstream polkadot-sdk
- `docs/benchmarks-and-weights.md` — guide covering new pallet setup, adding extrinsics, parameterized weights, CI workflow


docs/benchmarks-and-weights.md
```diff
@@ -0,0 +1,195 @@
...
+
+6. Wire up in `runtime/src/lib.rs`:
+   ```rust
+   type WeightInfo = pallet_<name>::weights::SubstrateWeight<Runtime>;
+   ```
+
+7. Add `type WeightInfo = ();` to all test mocks implementing your pallet's
+   `Config`.
+
+8. Register the pallet in the `define_benchmarks!` macro in
```
shamil-gadelshin, line 67: Is there a way to automate 8-9? We'll definitely miss these steps without some warning from CI or other tools.

shamil-gadelshin (dismissed):
Nice change! I added a small optional request for automation.

JohnReedV (dismissed):

docs/benchmarks-and-weights.md
l0r1s: Good idea. I added the `discover_pallets.sh` script and benchmark scripts will error when trying to benchmark a pallet not added to the `define_benchmarks!`.

pallets/registry/src/lib.rs
evgeny-s, line 8: Can we also scope it under the `runtime-benchmarks` feature?
And make sure you did it for all the places where we import this module in other pallets

ts-tests/utils/coldkey_swap.ts
```diff
@@ -6,8 +6,8 @@ import type { TypedApi } from "polkadot-api";
 import type { subtensor } from "@polkadot-api/descriptors";
 import { FixedSizeBinary } from "polkadot-api";
 
-export const ANNOUNCEMENT_DELAY = 10;
```
evgeny-s, line 9: some unrelated change? Do we want to keep it?

l0r1s: > some unrelated change? Do we want to keep it?

It was fixing flaky tests with zombienet so I think we can keep it here.

Cargo.lock
```diff
@@ -18331,6 +18333,15 @@ dependencies = [
  "subtensor-swap-interface",
 ]
 
+[[package]]
```
evgeny-s, line 18336: Do we want to add it to the lock file, since this is a separate tool

scripts/benchmark_all.sh
```diff
@@ -1,33 +1,84 @@
...
+# ── Benchmark loop ───────────────────────────────────────────────────────────
+for pallet in "${PALLETS[@]}"; do
+  output="${PALLET_OUTPUTS[$pallet]}"
+
+  echo ""
+  echo "════════════════════════════════════════════════════════"
+  echo " Benchmarking $pallet -> $output"
+  echo "════════════════════════════════════════════════════════"
+
+  "$NODE_BIN" benchmark pallet \
```
evgeny-s, line 65: Can we consider using frame omni bencher?

l0r1s: > Can we consider using frame omni bencher?

We can't because we have different host functions and frame-omni-bencher doesn't work in our case.

Cargo.lock
l0r1s: > Do we want to add it to the lock file, since this is a separate tool

I added it under the workspace to make it easier to have hoisted dependencies, the same way it is done on the Polkadot SDK for utilities like `subkey`.

evgeny-s (approved):
Added several minor comments. In general, a great initiative, keeping benchmarks in separate files instead of in-place values 🙌.

pallets/registry/src/lib.rs
l0r1s: > Can we also scope it under the `runtime-benchmarks` feature? And make sure you did it for all the places where we import this module in other pallets

Benchmarks use the real runtime and the mock is only used during benchmark tests suite when running `cargo test --features runtime-benchmarks -p pallet-registry` to ensure the benchmarks compile and run without errors. The registry mock is scoped to this pallet and is never used anywhere else so I don't think we should scope it under `runtime-benchmarks`.

JohnReedV (approved):

Issue #2403, opened 2026-02-05
Use imbalances to maintain TAO emission and burning


gztensor:
- [x] Convert Balance from u64 to TaoBalance (former TaoCurrency) - #2417
- [x] Maintain TAO staked and locked via PalletId-based subnet account IDs - #2524

MkDev11:
Hi @gztensor, I’d like to take a shot at this and wanted to check with you first before opening a PR.

My approach is to switch `Balance` from `u64` to `TaoCurrency` in `common`, then fix everything that breaks. That means adding whatever trait impls `TaoCurrency` needs for `pallet_balances`, updating the runtime and pallets to use the new type (and `TaoCurrency::from(...)` where we use raw numbers), and adding something like `TryFrom<U256>` for the precompiles. I’d also update the test mocks so everything still compiles and passes.

Since `TaoCurrency` is just a transparent wrapper over `u64`, I’m assuming we don’t need a storage migration.

Does that line up with what you had in mind, or would you rather do it in a different way (e.g. only some pallets first)? I am willing to change direction if you have a preference.

PR #2524, opened 2026-03-19
PalletId-based subnet account Ids maintain subnet TAO reserves

gztensor:
## Description

- Every subnet will have an account ID and the subnet swap pool reserves will be stored in subnet account balance.
- The subnet account ID is derived from SubtensorModule account ID and does not have a private key, so no transactions may be initiated by this account.
- Use of subnet account IDs as hotkeys for staking, registration of neurons, or as a subnet owner hotkey will be forbidden.

## TAO management for subnet registration

- When a new subnet is registered, the lock cost is transferred to the subnet account and is recorded to the `SubnetLocked` map for the newly registered netuid. 
- When subnet is dissolved, some of the `SubnetLocked` amount may be returned to the subnet owner. The owner emission TAO is excluded from the returned portion. If owner emission TAO is higher than lock cost, zero is returned for lock cost and the remainder is recycled.

## Implementation Checklist

- [x] Derive subnet account IDs
- [x] Create runtime API `subnetInfoRuntimeApi::getSubnetAccountId` for clients to read the subnet account IDs
- [x] Organize all TAO balance operations, especially creating and destroying TAO, into one place: `mint_tao`, `burn_tao`, `recycle_tao`, `transfer_tao`.
  - [x] Cover all functions with tests
- [x] Migration to mint TAO stored in `SubnetTAO` into subnet accounts. This will make balances pallet `TotalIssuance` match the subtensor pallet `TotalIssuance`.
- [x] Move `remove_balance_from_coldkey_account` and `add_balance_to_coldkey_account` methods to test-only code (tests+benchmarks). Convert original TAO moving logic into `transfer_tao` and use it inside `stake_into_subnet` and `untake_from_subnet`.
- [x] Use subnet accounts for storing subnet reserves for Alpha <> TAO swaps.
- [x] Use ballet balances `mint` in `get_block_emission`, make it return imbalance that is consumed later in `run_counbase`
- [x] Use subnet accounts for injecting of newly minted TAO (consume imbalances) in `run_coinbase`.
- [x] Handle unspent credit (error log + recycle), error-log attempts to spend more.
- [x] Subnet Locks
  - [x] Use subnet accounts for subnet locks
  - [x] Add corresponding migration: Mint TAO locked into subnet accounts
  - [x] In rao release (v2.0.0) the lock was burned (`TotalIssuance` reduction), in the subsequent migration `migrate_restore_subnet_locked` we restored locks into `SubnetLocked`, but did not increase the `TotalIssuance` back. So, in this PR we will increase subtensor pallet `TotalIssuance` by total of `SubnetLocked` update in `migrate_restore_subnet_locked`.
- [x] Maintain subtensor pallet total issuance within tao.rs file only (consider burning TAO when balances drop below ED)
- [x] Test dissolving network with stakes: Make sure all TAO is distributed to stakers, sn owner, etc. and nothing remains on the sn account ID.
- [x] Remove adjustment of TI on every runtime upgrade
- [x] Deprecate `sudo_set_total_issuance` in admin pallet
- [x] Implement consistent behavior of locks: store TAO + refund on dissolution and recycle the rest.
- [x] Document special accounts IDs (subnets, burn)

## Related Issue(s)

- Part of #2403 

## Type of Change
- [ ] Documentation update

## Checklist


JohnReedV (dismissed):

common/src/lib.rs
```diff
@@ -246,8 +246,8 @@ pub trait TokenReserve<C: Token> {
 pub trait BalanceOps<AccountId> {
     fn tao_balance(account_id: &AccountId) -> TaoBalance;
     fn alpha_balance(netuid: NetUid, coldkey: &AccountId, hotkey: &AccountId) -> AlphaBalance;
-    fn increase_balance(coldkey: &AccountId, tao: TaoBalance);
-    fn decrease_balance(coldkey: &AccountId, tao: TaoBalance) -> Result<TaoBalance, DispatchError>;
+    // fn increase_balance(coldkey: &AccountId, tao: TaoBalance);
```
shamil-gadelshin: Do we need this comment?

pallets/subtensor/src/migrations/migrate_rao.rs
```diff
@@ -91,7 +91,8 @@ pub fn migrate_rao<T: Config>() -> Weight {
         //         .checked_div(I96F32::from_num(1_000_000_000))
         //         .unwrap_or(I96F32::from_num(0.0)),
         // );
-        Pallet::<T>::add_balance_to_coldkey_account(&owner, remaining_lock.into());
+        let credit = Pallet::<T>::mint_tao(remaining_lock.into());
+        let _ = Pallet::<T>::spend_tao(&owner, credit, remaining_lock.into());
```
shamil-gadelshin, line 98: What to do with the result?

shamil-gadelshin: Please write a comment why exactly we skip the result.

pallets/subtensor/src/migrations/migrate_total_issuance.rs
```diff
@@ -19,6 +19,8 @@ pub mod deprecated_loaded_emission_format {
         StorageMap<Pallet<T>, Identity, u16, Vec<(AccountIdOf<T>, u64)>, OptionQuery>;
 }
 
+/// Note: Disabled as dangerous.
```
shamil-gadelshin: Should we remove it instead?

pallets/swap/src/pallet/impls.rs
```diff
@@ -829,131 +829,133 @@ impl<T: Config> Pallet<T> {
...
-        }
-
-        log::debug!(
-            "dissolve_all_liquidity_providers: netuid={netuid:?}, mode=V2-or-nonV3, leaving all liquidity/state intact"
-        );
+    pub fn do_dissolve_all_liquidity_providers(_netuid: NetUid) -> DispatchResultWithPostInfo {
+        // Deprecated in balancer anyway
+        let weight = Weight::default();
+
+        // if SwapV3Initialized::<T>::get(netuid) {
```
shamil-gadelshin: Can we remove this comment?

pallets/swap/src/pallet/mod.rs
```diff
@@ -463,49 +462,51 @@ mod pallet {
...
         ) -> DispatchResult {
-            let coldkey = ensure_signed(origin)?;
+            // Deprecated by balancer
 
-            // Ensure that the subnet exists.
-            ensure!(
-                T::SubnetInfo::exists(netuid.into()),
-                Error::<T>::MechanismDoesNotExist
-            );
+            // let coldkey = ensure_signed(origin)?;
```
shamil-gadelshin: Should we keep the comment?

pallets/swap/src/pallet/mod.rs
```diff
@@ -522,99 +523,101 @@ mod pallet {
         #[pallet::call_index(3)]
         #[pallet::weight(<T as pallet::Config>::WeightInfo::modify_position())]
         pub fn modify_position(
-            origin: OriginFor<T>,
-            hotkey: T::AccountId,
-            netuid: NetUid,
-            position_id: PositionId,
-            liquidity_delta: i64,
+            _origin: OriginFor<T>,
```
shamil-gadelshin, line 486: Do we need this comment?

pallets/subtensor/src/coinbase/block_emission.rs
```diff
@@ -17,7 +18,17 @@ impl<T: Config> Pallet<T> {
     /// # Returns
     /// * 'Result<u64, &'static str>': The calculated block emission rate or error.
     ///
-    pub fn get_block_emission() -> Result<TaoBalance, &'static str> {
+    pub fn get_block_emission() -> CreditOf<T> {
+        let maybe_tao_to_mint = Self::calculate_block_emission();
+        if let Ok(tao_to_mint) = maybe_tao_to_mint {
+            Self::mint_tao(tao_to_mint.into())
+        } else {
+            Self::mint_tao(0.into())
```
shamil-gadelshin: Why do we mint zero tao?

pallets/subtensor/src/migrations/migrate_rao.rs
```diff
@@ -108,7 +109,7 @@ pub fn migrate_rao<T: Config>() -> Weight {
             // Set Owner as the coldkey.
             SubnetOwnerHotkey::<T>::insert(netuid, owner_coldkey.clone());
             // Associate the coldkey to coldkey.
-            Pallet::<T>::create_account_if_non_existent(&owner_coldkey, &owner_coldkey);
+            let _ = Pallet::<T>::create_account_if_non_existent(&owner_coldkey, &owner_coldkey);
```
shamil-gadelshin, line 117: Why do we ignore the result?

pallets/subtensor/src/staking/remove_stake.rs
```diff
@@ -596,8 +592,20 @@ impl<T: Config> Pallet<T> {
...
         };
 
-        if !refund.is_zero() {
-            Self::add_balance_to_coldkey_account(&owner_coldkey, refund);
+        if !refund.is_zero()
+            && let Some(subnet_account) = Self::get_subnet_account_id(netuid)
+        {
+            // Transfer maximum transferrable up to refund to owner
+            let transferrable = Self::get_coldkey_balance(&subnet_account);
+            let _ = Self::transfer_tao(&subnet_account, &owner_coldkey, refund.min(transferrable));
```
shamil-gadelshin, line 602: Why did we ignore the result?

pallets/subtensor/src/subnets/registration.rs
```diff
@@ -74,7 +74,7 @@ impl<T: Config> Pallet<T> {
         );
 
         // 6) ensure pairing exists and is correct
-        Self::create_account_if_non_existent(&coldkey, &hotkey);
+        let _ = Self::create_account_if_non_existent(&coldkey, &hotkey);
```
shamil-gadelshin: Ignored result.

pallets/subtensor/src/staking/remove_stake.rs
```diff
@@ -561,7 +556,8 @@ impl<T: Config> Pallet<T> {
             // Credit each share directly to coldkey free balance.
             for p in portions {
                 if p.share > 0 {
-                    Self::add_balance_to_coldkey_account(&p.cold, p.share.into());
+                    // Cannot fail the whole transaction if this transfer fails
+                    let _ = Self::transfer_tao_from_subnet(netuid, &p.cold, p.share.into());
```
shamil-gadelshin, line 560: ignored result

pallets/subtensor/src/staking/claim_root.rs
```diff
@@ -179,12 +179,38 @@ impl<T: Config> Pallet<T> {
...
+                    owed_tao.amount_paid_out.to_u64().into(),
+                );
+
+                // Increase root subnet SubnetTAO
+                SubnetTAO::<T>::mutate(NetUid::ROOT, |total| {
+                    *total = total.saturating_add(owed_tao.amount_paid_out.into());
+                });
+
+                // Increase root `SubnetAlphaOut
+                SubnetAlphaOut::<T>::mutate(netuid, |total| {
```
shamil-gadelshin: Is it Alpha or Tao here? Should be tao paid out after `swap_alpha_for_tao`, no?

pallets/subtensor/src/staking/remove_stake.rs
```diff
@@ -596,8 +592,20 @@ impl<T: Config> Pallet<T> {
...
+            && let Some(subnet_account) = Self::get_subnet_account_id(netuid)
+        {
+            // Transfer maximum transferrable up to refund to owner
+            let transferrable = Self::get_coldkey_balance(&subnet_account);
+            let _ = Self::transfer_tao(&subnet_account, &owner_coldkey, refund.min(transferrable));
+        }
+
+        // 9) Recycle TAO remaining on the subnet account, forgive errors.
+        if let Some(subnet_account) = Self::get_subnet_account_id(netuid) {
+            let remaining_subnet_balance = Self::get_coldkey_balance(&subnet_account);
```
shamil-gadelshin: `get_coldkey_balance` and `recycle_tao` have different `Preservation` types. It seems like an issue.

pallets/subtensor/src/swap/swap_hotkey.rs
```diff
@@ -323,18 +319,14 @@ impl<T: Config> Pallet<T> {
...
-        Self::recycle_tao(actual_recycle_amount);
+        // 5. Remove the swap cost from the coldkey's account + Recycle the tokens
+        Self::recycle_tao(coldkey, swap_cost)?;
         weight.saturating_accrue(T::DbWeight::get().reads_writes(1, 1));
 
         // 7. Swap owner.
         // Owner( hotkey ) -> coldkey -- the coldkey that owns the hotkey.
         // Owner::<T>::remove(old_hotkey);
-        Owner::<T>::insert(new_hotkey, coldkey.clone());
+        Self::create_account_if_non_existent(coldkey, new_hotkey)?;
```
shamil-gadelshin: It's likely contains a bug for `swap_hotkey`. It should have `ensure` instead of `if !Self::hotkey_account_exists(hotkey) {`

shamil-gadelshin (changes requested):
Key findings:
- several possible bugs
- a lot of commented code 
- ignored results that could lead to inconsistent state

pallets/subtensor/src/migrations/migrate_rao.rs
gztensor: This migration has already run. We maintain the code only for reference. This code mimics what used to be here previously (add_balance_to_coldkey_account) as close as reasonably possible. I'll add that comment.

pallets/subtensor/src/migrations/migrate_total_issuance.rs
gztensor: I suggest to keep it for reference just because we have long commit history and it is a bit more convenient to have the code in the ~HEAD, but maybe the comment should be more descriptive. I'll adjust.

pallets/swap/src/pallet/impls.rs
gztensor: Ok, I left a longer explanation why we don't need this code and then removed the commented code.

pallets/swap/src/pallet/mod.rs
gztensor: Remvoed

pallets/subtensor/src/subnets/registration.rs
gztensor: Great catch! This was a bug. Fixed.

pallets/subtensor/src/staking/claim_root.rs
evgeny-s: Should we have `NetUid::ROOT` here or netuid?
Either update the comment or change the param.

pallets/subtensor/src/migrations/migrate_subnet_tao_to_subnet_balance.rs
```diff
@@ -0,0 +1,4 @@
+    // /// --- MAP ( netuid ) --> tao_in_subnet | Returns the amount of TAO in the subnet.
```
evgeny-s: Do we need it?

pallets/subtensor/src/subnets/subnet.rs
```diff
@@ -125,6 +127,12 @@ impl<T: Config> Pallet<T> {
             Error::<T>::NonAssociatedColdKey
         );
 
+        // Ensure that hotkey is not a special account
+        ensure!(
+            Self::is_subnet_account_id(hotkey).is_none(),
```
evgeny-s, line 132: We might need another error here, because the error is not related to the `NonAssociatedColdKey`

pallets/subtensor/src/staking/remove_stake.rs
gztensor: There's a comment above explaining why. Normally (and I have tested it multiple times on a mainnet clone), this place will not fail, but in the future if this fails because subnet account has insufficient TAO for any reason like a bug or an exploit, we cannot block subnet registration.

pallets/subtensor/src/macros/dispatches.rs
```diff
@@ -1778,7 +1778,7 @@ mod dispatches {
         pub fn try_associate_hotkey(origin: OriginFor<T>, hotkey: T::AccountId) -> DispatchResult {
             let coldkey = ensure_signed(origin)?;
 
-            let _ = Self::do_try_associate_hotkey(&coldkey, &hotkey);
+            Self::do_try_associate_hotkey(&coldkey, &hotkey)?;
```
evgeny-s, line 1779: Breaking change? Should we announce it to the clients?

pallets/subtensor/src/staking/claim_root.rs
gztensor: `owed_tao.amount_paid_out` contains TAO, but because we record it in `SubnetAlphaOut`, the data type needs to convert to `AlphaBalance`. Numerically it is equal to TAO amount because it is root, but it is alpha type.

`swap_alpha_for_tao` does not change `SubnetAlphaOut` or `SubnetTAO`, neither it transfers balances. 
`transfer_tao_from_subnet` does the transfer of TAO from origin subnet to root subnet.

pallets/subtensor/src/staking/stake_utils.rs
```diff
@@ -739,9 +737,11 @@ impl<T: Config> Pallet<T> {
     /// Unstakes alpha from a subnet for a given hotkey and coldkey pair.
     ///
     /// We update the pools associated with a subnet as well as update hotkey alpha shares.
+    /// Credits the unstaked TAO to the benefitiary account
     pub fn unstake_from_subnet(
         hotkey: &T::AccountId,
         coldkey: &T::AccountId,
+        benefitiary: &T::AccountId,
```
evgeny-s: nit: typo

pallets/subtensor/src/staking/remove_stake.rs
gztensor: Fixed.

pallets/subtensor/src/swap/swap_hotkey.rs
gztensor: I am not sure either. The intention was to check that new hotkey is a non-system hotkey to prevent someone setting subnet account ID to be the neuron. I just reverted this change and added an explicit check.

pallets/subtensor/src/staking/claim_root.rs
gztensor: Yes! Good catch.

pallets/subtensor/src/migrations/migrate_subnet_tao_to_subnet_balance.rs
gztensor: Removed.

pallets/subtensor/src/macros/dispatches.rs
gztensor: It is only a breaking change for the ones who will try to associate a system hotkey (subnet account ID), which is an exploit. Honest users are not affected.

pallets/subtensor/src/subnets/subnet.rs
gztensor: Added error `CannotUseSystemAccount`.

pallets/subtensor/src/staking/stake_utils.rs
gztensor: Fixed

pallets/subtensor/src/coinbase/tao.rs
```diff
@@ -0,0 +1,294 @@
...
+    /// Transfer TAO from one coldkey account to another.
+    ///
+    /// This is a plain transfer and may reap the origin account if `amount` reduces
+    /// its balance below the existential deposit (ED).    
+    pub fn transfer_tao(
+        origin_coldkey: &T::AccountId,
+        destination_coldkey: &T::AccountId,
+        amount: BalanceOf<T>,
+    ) -> DispatchResult {
+        let max_transferrable = <T as pallet::Config>::Currency::reducible_balance(
```
evgeny-s: `reducible_balance` returns the balance excluding the frozen ([reference](https://github.com/paritytech/polkadot-sdk/blob/215252e7b55a3239a4f0b6b86aa03e9946c7ae2e/substrate/frame/balances/src/impl_fungible.rs#L47)), we need to take this into the consideration when we want to mutate the TI when the remainder < ED.

pallets/subtensor/src/subnets/subnet.rs
```diff
@@ -193,8 +192,17 @@ impl<T: Config> Pallet<T> {
         Self::init_new_network(netuid_to_register, default_tempo);
         log::debug!("init_new_network: {netuid_to_register:?}");
 
-        // --- 13. Add the caller to the neuron set.
-        Self::create_account_if_non_existent(&coldkey, hotkey);
+        // --- 10. Perform the lock operation (transfer TAO from owner's coldkey to subnet account).
+        let actual_tao_lock_amount =
+            Self::transfer_tao_to_subnet(netuid_to_register, &coldkey, lock_amount.into())?;
```
evgeny-s, line 197: We transfer tao here with the mode `Preservation::Preserve`, that means that if user has less the ED, it will deduct it from the amount in favour of ED. But we check the `can_remove_balance_from_coldkey_account` with ED. It means that we might deduct a bit less than the lock cost.

evgeny-s: The same situation in this method: `do_register`

gztensor: This is how subnet registration, and especially `can_remove_balance_from_coldkey_account` and `remove_balance_from_coldkey_account` worked before this change. "can remove" uses `Preservation::Expendable` and includes ED and "remove balance" preserves the account even if we have to withdraw less.

evgeny-s (dismissed):
LGTM

pallets/subtensor/src/coinbase/tao.rs
gztensor: Thank you! Fixed.

shamil-gadelshin (dismissed):

JohnReedV (approved):

sam0x17 (approved):

PR #2533, opened 2026-03-24
MEV V2 infrastructure

shamil-gadelshin:
This PR introduces infrastructure for the upcoming MEV Shield V2 feature. The idea is to accumulate encrypted extrinsics in a queue and decrypt them in the next block during the `on_initialize` phase. Encryption and decryption are
  outside the scope of this PR and abstracted via the `ExtrinsicDecryptor` trait. Protective limits include encrypted call size (const), queue size, weight for extrinsic execution, and lifetime in the queue.                               

###  PR features
                                                                                                                                                                                                                          
  - Add `on_initialize` hook to process pending encrypted extrinsics with configurable weight limits, extrinsic lifetime, and queue size
  - Add `store_encrypted` extrinsic for queuing encrypted calls for deferred execution
  - Add three root-gated configuration extrinsics: `set_max_pending_extrinsics_number`, `set_on_initialize_weight`, and `set_stored_extrinsic_lifetime`
  - Introduce `ExtrinsicDecryptor `trait for decrypting stored extrinsics before dispatch, with expiration, weight budgeting, and event logging

###  Details

 #### New Storage Items

  - `PendingExtrinsics` — map of queued encrypted extrinsics awaiting execution
  - `NextPendingExtrinsicIndex` / `PendingExtrinsicCount` — auto-increment index and live count for the queue
  - `MaxPendingExtrinsicsLimit` — configurable max queue depth (default: 100)
  - `OnInitializeWeight` — configurable max weight budget for on_initialize processing (default: 500B ref_time, capped at 2T)
  - `ExtrinsicLifetime` — configurable max age in blocks before expiration (default: 10)

  #### New Extrinsics

  - `store_encrypted` — signed extrinsic to queue an encrypted call
  - `set_max_pending_extrinsics_number` — root-only, sets queue limit
  - `set_on_initialize_weight` — root-only, sets weight budget with absolute max guard
  - `set_stored_extrinsic_lifetime` — root-only, sets expiration window

 #### Processing Logic (`on_initialize`)

  Iterates pending extrinsics in insertion order. For each entry:
  1. Expired entries are removed and `ExtrinsicExpired` is emitted
  2. Decryption failures are removed and `ExtrinsicDecodeFailed` is emitted
  3. If dispatching would exceed the weight budget, processing stops (`ExtrinsicPostponed`)
  4. Otherwise, the call is dispatched from the submitter's signed origin

  #### Other Changes

  - Added `RuntimeCall` and `ExtrinsicDecryptor` associated types to pallet_shield::Config
  - Disambiguated `T::RuntimeCall` in `check_mortality.rs` to resolve ambiguity between `frame_system::Config` and `pallet_shield::Config`
  - Comprehensive tests for all new extrinsics, storage limits, expiration, weight budgeting, and edge cases

 #### Unit Test Plan

  - Unit tests for `store_encrypted`(success, queue full, event emission)
  - Unit tests for `set_max_pending_extrinsics_number` (root-only, value persistence)
  - Unit tests for `set_on_initialize_weight` (root-only, absolute max rejection)
  - Unit tests for `set_stored_extrinsic_lifetime` (root-only, value persistence)
  - Unit tests for `on_initialize` processing (dispatch, expiration, weight limits, decode failures)


#### Benchmarks
To be added after the initial review. (UPDATE: the benchmarks were added)


pallets/shield/src/lib.rs
```diff
@@ -93,11 +119,89 @@ pub mod pallet {
...
+        pub call: BoundedVec<u8, MaxCallSize>,
+        /// The block number when the extrinsic was submitted.
+        pub submitted_at: BlockNumberFor<T>,
+    }
+
+    /// Storage map for encrypted extrinsics to be executed in on_initialize.
+    /// Uses u32 index for O(1) insertion and removal.
+    #[pallet::storage]
+    pub type PendingExtrinsics<T: Config> =
+        StorageMap<_, Identity, u32, PendingExtrinsic<T>, OptionQuery>;
```
l0r1s: This is maybe one of the rare case where we could use a `CountedStorageMap`

pallets/shield/src/lib.rs
```diff
@@ -93,11 +119,89 @@ pub mod pallet {
     pub type HasMigrationRun<T: Config> =
         StorageMap<_, Identity, BoundedVec<u8, MigrationKeyMaxLen>, bool, ValueQuery>;
 
+    /// Maximum size of a single encoded call.
+    pub type MaxCallSize = ConstU32<8192>;
```
l0r1s: I'm not sure if this is the correct naming here because it seems to be the ciphertext, which may be bigger than the actual call, maybe something like `MaxCiphertextSize` or `MaxEncryptedCallSize` ?

pallets/shield/src/lib.rs
```diff
@@ -255,6 +445,98 @@ pub mod pallet {
...
+                    });
+                }
+            }
+        }
+
+        /// Remove a pending extrinsic from storage and decrement count.
+        fn remove_pending_extrinsic<T: Config>(index: u32, weight: &mut Weight) {
+            PendingExtrinsics::<T>::remove(index);
+            PendingExtrinsicCount::<T>::mutate(|c| *c = c.saturating_sub(1));
+            *weight = weight.saturating_add(T::DbWeight::get().writes(2));
```
l0r1s: `mutate` has a read too

pallets/shield/src/lib.rs
```diff
@@ -93,11 +119,89 @@ pub mod pallet {
...
+
+    /// A pending extrinsic stored for later execution.
+    #[freeze_struct("c5749ec89253be61")]
+    #[derive(Clone, Encode, Decode, TypeInfo, MaxEncodedLen, PartialEq, Debug)]
+    #[scale_info(skip_type_params(T))]
+    pub struct PendingExtrinsic<T: Config> {
+        /// The account that submitted the extrinsic.
+        pub who: T::AccountId,
+        /// The encoded call data.
+        pub call: BoundedVec<u8, MaxCallSize>,
```
l0r1s: Same here, probably more explicit name like `encrypted_call`?

pallets/shield/src/lib.rs
```diff
@@ -255,6 +445,98 @@ pub mod pallet {
...
+            let call = match T::ExtrinsicDecryptor::decrypt(&pending.call) {
+                Ok(call) => call,
+                Err(_) => {
+                    remove_pending_extrinsic::<T>(index, &mut weight);
+
+                    Self::deposit_event(Event::ExtrinsicDecodeFailed { index });
+
+                    continue;
+                }
+            };
```
l0r1s: let Ok() cleaner here, no?

pallets/shield/src/tests.rs
```diff
@@ -460,3 +460,627 @@ mod migration_tests {
...
+
+            // Verify ExtrinsicDecodeFailed event was emitted
+            System::assert_has_event(
+                crate::Event::<Test>::ExtrinsicDecodeFailed { index: 0 }.into(),
+            );
+        });
+    }
+
+    #[test]
+    fn on_initialize_handles_dispatch_failure() {
```
l0r1s, line 566: test name is `on_initialize_handles_dispatch_failure` but we check success on multiple calls?

l0r1s (commented):
Overall the design is good. My only concern is that the inner call is never charged for, the user gets free execution for whatever is inside. This could probably be fixed by implementing an additional check when trying to execute the call similar to what the `ChargeTransactionPayment` extension does and reject it if fee can't be paid, could even be done through a `DispatchExtension`.

JohnReedV (commented):
Looks good to me. Didn't see anything in addition to Loris’s comments.

shamil-gadelshin:
## Update summary

- applied review suggestions
- added benchmarks and weights for all new extrinsics
- migrated pallet benchmarks to the `weights trait` style
- added a new updatable contraint - max single extrinsic weights
- added a custom weights function for `store_encrypted` extrinsic                                                                                                                                              - new tests

l0r1s (dismissed):
Looks good! I'm not sure the fixed weight for the `store_encrypted` is the best way in the long run but it should be enough to start with and we can switch to a more dynamic weight calculation in the future.

shamil-gadelshin:
> Looks good! I'm not sure the fixed weight for the `store_encrypted` is the best way in the long run but it should be enough to start with and we can switch to a more dynamic weight calculation in the future.

I'm developing the algorithm we discussed previously (refund) in a separate PR.

sam0x17 (approved):

PR #2550, opened 2026-03-31
add origin in extension

### LINKED ISSUE #2549: Enable Caller Context Preservation in Chain Extension for Staking Operations

At present, whenever a contract invokes the chain extension, the extension assigns the contract itself as the caller. This results in:

- Loss of the original transaction sender context
- Inability for users to move or manage their stake via contract calls
- Limited composability for DeFi and staking abstractions

**Expected Behavior**

The chain extension should support preserving or forwarding the original caller context (msg.sender equivalent), allowing:

- Users to interact with staking logic through contracts
- Stake operations to be executed on behalf of the actual caller, not the contract
- More flexible and composable contract-based staking mechanisms 

**Proposed Solution**

Introduce a modified or new chain extension that:

- Allows passing the original caller as the effective actor in staking operations
- Alternatively, provides a mechanism to explicitly specify the caller context when invoking the extension
- Ensures proper authorization and security checks to prevent misuse

_No response_

### Additional context

_No response_

open-junius:
## Description
Add the origin in the Ink, use can choose the contract add as signed account or real origin as account

## Related Issue(s)

- Closes #2549 

## Type of Change

- [ ] Documentation update

## Breaking Change

## Checklist

## Additional Notes


camfairchild (changes requested):
Waiting on response on the issue about using `delegate_call` instead

chain-extensions/src/lib.rs
```diff
@@ -61,9 +61,524 @@ where
...
+
+    fn dispatch_remove_stake_v1<Env>(
+        env: &mut Env,
+        origin: RawOrigin<T::AccountId>,
+    ) -> Result<RetVal, DispatchError>
+    where
+        Env: SubtensorExtensionEnv<T>,
+        <<T as SysConfig>::Lookup as StaticLookup>::Source: From<<T as SysConfig>::AccountId>,
+    {
+        let weight = Weight::from_parts(196_800_000, 0)
```
shamil-gadelshin, line 103: Can we use `<T as crate::pallet::Config>::WeightInfo::remove_stake_limit` instead?

ales-otf (dismissed):

shamil-gadelshin (changes requested):
Can we use `pallet-subtensor` `WeightInfo` implementation for weights?

chain-extensions/src/lib.rs
open-junius: Yes, we should use weights from subtensor pallet.

open-junius:
> Waiting on response on the issue about using `delegate_call` instead

I did some research in try-delegate-call branch 
https://github.com/opentensor/subtensor/blob/try-delegate-call/contract-tests/delegate-call/lib.rs#L68. 
- user call -> contract -> runtime extension
caller is user's account and address is contract's address
- user call -> delegate call contract -> direct call -> runtime extension
 caller is user's account and address is delegate call contract's address

In Ink, we must delegate call with the contract codehash like follows. 
There is no codehash for runtime extension part, so we can't do 
- user call -> delegate call contract -> runtime extension

```rust
let my_return_value = build_call::<CustomEnvironment>()
                .delegate(ink::primitives::Hash::from(code_hash))
                .exec_input(ExecutionInput::new(Selector::new(ink::selector_bytes!(
                    "dummy"
                ))))
                .returns::<Result<(), ReadWriteErrorCode>>()
                .invoke();
```

chain-extensions/src/lib.rs
```diff
@@ -61,9 +62,514 @@ where
...
+        origin: RawOrigin<T::AccountId>,
+    ) -> Result<RetVal, DispatchError>
+    where
+        Env: SubtensorExtensionEnv<T>,
+        <<T as SysConfig>::Lookup as StaticLookup>::Source: From<<T as SysConfig>::AccountId>,
+    {
+        let weight =
+            <<T as pallet_subtensor::Config>::WeightInfo as SubtensorWeightInfo>::add_stake();
+
+        env.charge_weight(weight)?;
```
evgeny-s, line 80: We might want to charge full execution weights of the extrinsic after the decode; in case of a decode issue, the contract will pay for the whole execution.

open-junius: make sense, let me check it.

sam0x17 (approved):

gztensor (approved):

camfairchild:
> In Ink, we must delegate call with the contract codehash like follows. 
> There is no codehash for runtime extension part, so we can't do 
> - user call -> delegate call contract -> runtime extension

Are we able to modify the delegate function to allow it to work?

This might be a better solution.

Or possibly a dedicated "delegate_extension" call?

open-junius:
> > In Ink, we must delegate call with the contract codehash like follows.
> > There is no codehash for runtime extension part, so we can't do
> > 
> > * user call -> delegate call contract -> runtime extension
> 
> Are we able to modify the delegate function to allow it to work?
> 
> This might be a better solution.
> 
> Or possibly a dedicated "delegate_extension" call?

delegate_extension, do you mean the origin as caller case? If yes, it should be doable, just add a new extension.

To make the delegate_call works for runtime extension, I need do some research in Ink. will update here after it is done.

sam0x17:
tests still failing

open-junius:
> tests still failing

Fixed the tests, please check again.

sam0x17 (dismissed):

contract-tests/test/wasm.contract.test.ts
```diff
@@ -50,33 +53,50 @@ describe("Test wasm contract", () => {
...
-            contractAddress,
+            dest,
             netuid,
         ))?.stake
 
         assert.ok(stake !== undefined)
         assert.ok(stake > BigInt(0))
     }
 
+    async function initSecondColdAndHotkey() {
```
gztensor, line 64: Test "Can set coldkey auto stake hotkey" and "Can add and remove proxy" read hotkey2.publicKey without calling this helper first. This may lead to order dependent failures.

gztensor (changes requested):

contract-tests/test/wasm.contract.test.ts
open-junius: It should use the hotkey in both test cases. will fix it.

open-junius:
> > In Ink, we must delegate call with the contract codehash like follows.
> > There is no codehash for runtime extension part, so we can't do
> > 
> > * user call -> delegate call contract -> runtime extension
> 
> Are we able to modify the delegate function to allow it to work?
> 
> This might be a better solution.
> 
> Or possibly a dedicated "delegate_extension" call?

read through Ink documentation, don't find out the approach that we can delegate to runtime extension. 
delegate just between two smart contracts.

sam0x17 (approved):

JohnReedV (approved):

PR #2551, opened 2026-03-31
Fix testnet warp sync and protocol isolation

prop-opentensor:
## Summary
- set the raw testnet chainspec `protocolId` to `bittensor-testnet` so testnet nodes stop sharing a peer set with mainnet
- add a testnet-only GRANDPA warp sync provider/verifier override for the known historical authority-set transitions that broke warp sync on testnet
- preserve the justifications needed to continue serving and consuming warp proofs on testnet

## Problem
Fresh testnet warp sync is currently broken for two separate reasons.

First, the raw testnet spec still uses the mainnet `protocolId`, so testnet nodes can connect to the wrong peer set and receive cross-chain traffic.

Second, testnet has historical GRANDPA authority-set transitions that cannot be handled correctly by the current generic testnet warp configuration alone. In particular, the canonical pre-fragment states around blocks `4589660`, `4589686`, and `5534451` need explicit handling, and the `5534451` transition must be verified against pre-change `set_id = 3`, not `2`.

Without those testnet-specific corrections, a fresh node can accept the wrong peer set, reject the relevant warp proof fragments, or stall partway through warp sync.

There is also a generic serving-side issue in the SDK for delayed GRANDPA changes. That is tracked separately in the companion SDK PR: https://github.com/opentensor/polkadot-sdk/pull/22

## Fix
This PR keeps the change surface testnet-specific.

- `raw_spec_testfinney.json` gets a distinct protocol ID so testnet discovery and gossip are isolated from mainnet.
- `service.rs` adds a testnet-only warp provider keyed off `protocolId == bittensor-testnet`.
- That provider prefixes the canonical historical transition points before any poisoned or incomplete local `authority_set_changes()` history.
- The verifier side uses explicit authority-set overrides for those exact transition blocks instead of relying on a global testnet set-id offset.
- Testnet no longer skips the justifications that later warp fragments depend on.

For all later normal scheduled GRANDPA changes, the provider still falls back to the backend’s live authority-set history.

## Why This Is Safe
- mainnet behavior is unchanged; the existing live-chain path remains intact
- the custom warp provider only activates for `protocolId = bittensor-testnet`
- the hard-coded overrides are limited to three known historical testnet transition blocks
- later normal authority changes still use the normal GRANDPA scheduled-change path
- no runtime APIs or external interfaces change beyond the testnet `protocolId` string

## Companion SDK Change
This Subtensor PR fixes the testnet-specific consumer/history side.

Official peers also need the generic SDK serving-side fix from https://github.com/opentensor/polkadot-sdk/pull/22 so they can encode delayed GRANDPA change fragments correctly during warp sync.

Because that SDK change extends the warp-proof encoding, production deployment also needs Subtensor to pick up that SDK revision when it is merged. The two PRs are complementary:
- this PR corrects testnet’s historical warp verification state and peer isolation
- the SDK PR fixes delayed-change warp proof generation/verification generically for serving nodes

## Notes
This patch is narrow by design. It should continue to work for future normal authority-set changes. Another code change should only be needed if testnet introduces another abnormal/manual historical GRANDPA transition that cannot be reconstructed from the normal aux history.

### DIFF
--- a/chainspecs/raw_spec_testfinney.json
+++ b/chainspecs/raw_spec_testfinney.json
@@ -6,7 +6,7 @@
     "/dns/bootnode.test.chain.opentensor.ai/tcp/30333/p2p/12D3KooWPM4mLcKJGtyVtkggqdG84zWrd7Rij6PGQDoijh1X86Vr"
   ],
   "telemetryEndpoints": null,
-  "protocolId": "bittensor",
+  "protocolId": "bittensor-testnet",
   "properties": {
     "ss58Format": 42,
     "tokenDecimals": 9,
--- a/node/src/chain_spec/testnet.rs
+++ b/node/src/chain_spec/testnet.rs
@@ -55,7 +55,7 @@ pub fn finney_testnet_config() -> Result<ChainSpec, String> {
             .parse()
             .unwrap(),
     ])
-    .with_protocol_id("bittensor")
+    .with_protocol_id("bittensor-testnet")
     .with_id("bittensor")
     .with_chain_type(ChainType::Development)
     .with_genesis_config_patch(testnet_genesis(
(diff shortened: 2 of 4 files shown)


sam0x17:
@prop-opentensor some merge conflicts

PR #2554, opened 2026-04-01
Pallet advanced limit orders

girazoki:
## Description                                                                                                                                                                                                          
  Introduces pallet-limit-orders, a new FRAME pallet that allows users to sign limit orders off-chain and have relayers submit them on-chain. Orders are executed against subnet AMM pools,  supporting three order types: `LimitBuy`, `TakeProfit`, and `StopLoss`.
                                                                                                                                                                                               
 Two execution paths:                                                                                                                                                                         
  - `execute_orders` — processes orders individually, best-effort. Invalid/expired/price-unmet orders are silently skipped; the call always returns Ok.
  - `execute_batched_orders`(netuid, orders) — atomic batch for a single subnet. Nets buy and sell sides before touching the AMM pool (reduced price impact). Fails the entire call if any order 
  is invalid.                                               
                                                                                                                                                                                               
  **Validation enforced on both paths:**                        
  - `Sr25519`-only signatures (`Ed25519`|`ECDSA` rejected for now, there is a possibility to add in the future)                                                                                                                                             
  - Expiry checks against wall-clock time (pallet_timestamp).                                                                                                                         
  - Price condition checks against current AMM spot price.
  - Minimum stake enforcement (delegated to pallet_subtensor)  **per order**, but not applicable for inner-pallet operations                                                                                                                     
  - Hotkey–coldkey association required for order creators, but not for the inner pallet for now (although can be added).
  - `SubtokenEnabled` must be true for the target subnet.                                                                                                                                      
  - Root netuid (0) explicitly blocked.
                                                                                                                                                                                               
  **Fee model:** per-order `fee_rate`: Perbill + `fee_recipient`: AccountId fields on the order struct. No global protocol fee. Fees are batched per unique recipient and transferred in a single call per recipient at the end of each batch. Enables flexibility of fee-percentage and fee-account per order.                                                                                                                                                
                                                                                                                                                                                               
  Replay protection: executed and cancelled orders are written to `Orders`: StorageMap<H256, OrderStatus> by their `blake2_256` hash.                                                              
                                                                                         
  ---                                                                                                                                                                                          
  Open questions / TODOs                                    
- [X] `LastColdkeyHotkeyStakeBlock` on pallet intermediary account — when execute_batched_orders transfers alpha from sellers into the pallet intermediary account, should `LastColdkeyHotkeyStakeBlock` be updated for the pallet account? If the block tracking is used for rate-limiting or cooldown enforcement, intermediate transfers through the pallet account  could trigger unintended restrictions on subsequent operations in the same block.
**Answer**: The pallet should not be rate-limited as it would break it's existence. rate-limiting is already applied to the cold keys that use the pallet so we are protected against that, but the pallet does operations like collecting all TAO/alpha and then distribute in batch mode. so no, we should not rate-limit
 
- [X]  Hotkey ownership verification for the pallet intermediary — the pallet intermediary account stakes/unstakes via PalletHotkey without going through the normal hotkey–coldkey association checks. Should we add an explicit ownership or registration check, or is the pallet account's privileged position sufficient justification to bypass it?                                     
 
- [X] Auto-associate `PalletHotkey` with the pallet coldkey on first use — if the pallet hotkey is not yet registered/associated with the pallet's derived coldkey, another cold-key could potentially claim the hotkey being used by the pallet.
- [X]  TypeScript / e2e tests — no TypeScript integration tests yet. These should cover: order signing, relayer submission flow, fee collection, and the batched netting path against a live node.
- [X]  Since it's a complex feature, I would like to add `isPalletEnabled` storage item, initially set to false, so that it can be enabled in testnet but not in mainnet
- [X] Benchmarks

## Type of Change

- [ ] Documentation update


## Checklist

- [X] I have run `./scripts/fix_rust.sh` to ensure my code is formatted and linted correctly
- [X] I have made corresponding changes to the documentation
- [X] Any dependent changes have been merged and published in downstream modules


## Additional Notes

I moved pallets/swap-interface to primitives/swap-interface. it was not a pallet, so I thought it belonged more there


camfairchild:
is this meant to be like an EIP-712 ?

JohnReedV (dismissed):

pallets/limit-orders/src/lib.rs
```diff
@@ -0,0 +1,1220 @@
...
+                &T::PalletHotkey::get(),
+            );
+        }
+    }
+
+    // ── Hooks ─────────────────────────────────────────────────────────────────
+
+    #[pallet::hooks]
+    impl<T: Config> Hooks<BlockNumberFor<T>> for Pallet<T> {
+        fn on_runtime_upgrade() -> Weight {
```
open-junius: Need to avoid running it during every upgrade. can set a migration function which include a flag like https://github.com/opentensor/subtensor/blob/main/pallets/crowdloan/src/lib.rs#L282

pallets/subtensor/src/staking/order_swap.rs
```diff
@@ -0,0 +1,205 @@
...
+        Self::increase_stake_for_hotkey_and_coldkey_on_subnet(
+            to_hotkey, to_coldkey, netuid, amount,
+        );
+        LastColdkeyHotkeyStakeBlock::<T>::insert(
+            to_coldkey,
+            to_hotkey,
+            Self::get_current_block_as_u64(),
+        );
+        if validate_receiver {
+            ensure!(
```
open-junius: put the check before the state update.

gorka-i: it's transactional so it does not really matter no? in any case I think its good practice yes

open-junius: The result is fine. Yes, it is a good practice, put all checks at the beginning of extrinsic.

gorka-i: addressed

pallets/limit-orders/src/lib.rs
```diff
@@ -0,0 +1,1232 @@
...
+                    &order.signer,
+                    &order.hotkey,
+                    order.netuid,
+                    tao_after_fee,
+                    TaoBalance::from(effective_swap_limit),
+                    true,
+                )?;
+
+                // Forward the fee TAO to the order's fee recipient.
+                Self::forward_fee(&order.signer, &order.fee_recipient, fee_tao);
```
open-junius: should handle the failure of forward_fee. otherwise the user could use fee_rate and its balance to avoid pay it.

https://github.com/opentensor/subtensor/pull/2685#discussion_r3311255631

gztensor (dismissed):

lorismoulin (changes requested):
Looks good, just taking a look at the `meta-tx` pallet in the polkadot-sdk so we can get the relayer logic for any extrinsic and not coupled to limit orders.

gorka-i:
> Looks good, just taking a look at the `meta-tx` pallet in the polkadot-sdk so we can get the relayer logic for any extrinsic and not coupled to limit orders.

Not in a straight forward manner for a couple of reasons:

- this pallet allows the relayer to inject orders coming from the same user in whatever order as long as the orders pass the validations. these would not be true with meta-tx, since each tx would have a nonce, and a tx with nonce 2 would invalidate a previous tx with nonce 1. this does not play well with the pallet, as in our case we submit orders to be executed whenever the price condition satisfies, regardless if it has nonce 1 or 2.

- meta-tx disables the `batched_execution`. in our case the relayer is supposed to be smart enough to detect that 2 or more orders happen against the same alpha token, group them together, and execute them in batched mode, influencing less the price than if we execute them individually.

In short is meta-tx a good pallet? definitely, but just by itself it does not seem to fulfill the requirements of this pallet.

lorismoulin:
@gorka-i make sense, thanks for looking at it

open-junius:
The current implementation looks good to me. I'd like to know if you will continue the PR to handle three issues in the TODO list. I have a suggestion about the intermediary hotkey, we could skip it, not calling transfer alpha. Instead, use the burn_alpha and mint_alpha for buyers and sellers.

girazoki:
> The current implementation looks good to me. I'd like to know if you will continue the PR to handle three issues in the TODO list. I have a suggestion about the intermediary hotkey, we could skip it, not calling transfer alpha. Instead, use the burn_alpha and mint_alpha for buyers and sellers.

Some of them have already been done. The only thing left to do is we apply rate-limitting to the pallet account but that would break the nature of it (the pallet might do several operations within a single-block, as it collects in the batched mode all alpha/tao and then distributes). so I'd say is fine if we leave it as is.

Updated the list

open-junius (approved):

gztensor (approved):

lorismoulin (approved):

sam0x17 (approved):

PR #2558, opened 2026-04-01
Stake Lock as Owner Alpha on Subnet Registration

JohnReedV:
## Summary

This PR updates **subnet registration** so the **owner key receives initial staked alpha at registration time**.

- recycled the remainder,
- and set the subnet owner metadata.

The owner did **not** receive any initial alpha stake from the registration spend.

With this change, the owner now receives **staked alpha equivalent to the TAO spent on registration**, converted using the **current median subnet alpha price** at the time of registration.

> **In one line:**  
> **Registering a subnet now grants the owner initial staked alpha priced off the median subnet alpha market price.**

---

## Registration flow change

After the registration lock is taken and the subnet is initialized:

1. the new subnet pool is still seeded using `network_min_lock`
2. the current **median subnet alpha price** is computed from existing subnets
3. the TAO actually spent on registration is converted into alpha using that median price
4. that alpha is assigned as staked alpha to the **subnet owner** hotkey/coldkey pair

This preserves the existing subnet initialization flow while adding an owner stake position at creation time.

### DIFF
--- a/pallets/subtensor/src/subnets/subnet.rs
+++ b/pallets/subtensor/src/subnets/subnet.rs
@@ -1,5 +1,7 @@
 use super::*;
+use safe_math::FixedExt;
 use sp_core::Get;
+use substrate_fixed::types::U96F32;
 use subtensor_runtime_common::{NetUid, TaoBalance};
 impl<T: Config> Pallet<T> {
     /// Returns true if the subnetwork exists.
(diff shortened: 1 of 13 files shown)


pallets/subtensor/src/subnets/subnet.rs
```diff
@@ -219,13 +231,22 @@ impl<T: Config> Pallet<T> {
...
         SubnetAlphaInProvided::<T>::insert(netuid_to_register, AlphaBalance::ZERO);
-        SubnetAlphaOut::<T>::insert(netuid_to_register, AlphaBalance::ZERO);
+        SubnetAlphaOut::<T>::insert(netuid_to_register, owner_alpha_stake);
         SubnetVolume::<T>::insert(netuid_to_register, 0u128);
         RAORecycledForRegistration::<T>::insert(
             netuid_to_register,
             actual_tao_lock_amount_less_pool_tao,
         );
 
+        if owner_alpha_stake > AlphaBalance::ZERO {
```
ales-otf, line 241: can also be just `!owner_alpha_stake.is_zero()`.

ales-otf (approved):

sam0x17 (approved):

PR #2560, opened 2026-04-02
Feat/recycle burn chain exts

### LINKED ISSUE #2564: Add chain exts for direct recycle/burn from ink! smart contracts

ink! smart contracts on subtensor that need to recycle or burn alpha stake must currently transfer funds to a designated external wallet, which then performs the operation. This intermediary wallet holds controllable funds, creating custodial risk regardless of intent. There is no way for a WASM smart contract to call recycle or burn directly via the chain extension interface.

Add 4 new chain extension functions allowing ink! contracts to invoke recycle/burn directly
                                                                                                                                                            
| Name | Description |
|---|---|                                                                                                                                               
| `RecycleAlphaV1` | Recycle alpha stake from the contract |
| `BurnAlphaV1` | Burn alpha stake from the contract |      
| `AddStakeRecycleV1` | Atomically add TAO stake then recycle alpha |                                                                                  
| `AddStakeBurnV1` | Atomically add TAO stake then burn alpha |

The current `AddStakeV1` chain extension only returns a success/error code, not the resulting alpha amount. While a contract could work around this by calling `AddStakeV1` followed by `RecycleAlphaV1(MAX)`, this recycles **all** alpha the contract holds on that subnet — not just the alpha from that specific stake. This is fine for simple contracts, but may break down for contracts that hold alpha for multiple purposes on the same subnet.
                                                                                                                                                              
The `AddStakeRecycleV1` and `AddStakeBurnV1` functions handle this atomically in a single call, recycling/burning only the alpha produced by that specific stake. This provides the robustness needed for a general-purpose chain extension interface.

The current approach: transferring collected fees to a designated wallet that performs recycle/burn externally. This works functionally but requires trust in the wallet operator and introduces a single point of failure. It also prevents fully autonomous contract operation.

### Additional context

Primary use case: smart contracts collecting subnet fees that need to autonomously recycle those fees back into the subnet without human intervention or off-chain infrastructure.

LandynDev:
## Description

Add 4 new chain extension functions (IDs 16-19) enabling WASM smart contracts to recycle and burn alpha stake.

- `RecycleAlphaV1` (16): Recycle alpha stake
- `BurnAlphaV1` (17): Burn alpha stake
- `AddStakeRecycleV1` (18): Atomically add TAO stake then recycle
- `AddStakeBurnV1` (19): Atomically add TAO stake then burn

All functions return the actual alpha amount via the output buffer.

## Related Issue(s)

Closes #2564

## Type of Change

## Breaking Change

No breaking changes. Existing chain extension functions (IDs 0-15) are unaffected. New function IDs 16-19 are additive.

## Checklist

## Additional Notes

7 unit tests cover: success paths for all 4 functions, root subnet rejection, nonexistent subnet, and insufficient balance. E2E validated on localnet with an ink! contract calling `add_stake_recycle` via chain extension.

**I also ran some local E2E testing with a local running subtensor chain (with the new extensions), deployed contract, etc.** 

Screenshot 1 shows:
- Accumulated fees of 0.0034 TAO on the contract (to be recycled)
- Total recycled fees of 0.0000 TAO on the contract
- Contract Balance of 5.0000 TAO
image

Screenshot 2 shows:
- Recycle of 0.0034 TAO succeeded (CLI -> Smart contract func -> New add stake + recycle chain ext)
- Accumulated fees of 0.0000 TAO on the contract (it got recycled)
- Total recycled fees of 0.0034 TAO on the contract
- Contract Balance of 4.9966 TAO
image

Unit tests successfully passing:
image


open-junius:
Thanks for the contribution. Can you add the ticket for the requirement? I am wondering the extension like  AddStakeRecycleV1,  who really want to do it. just confirm all is from the real requirements.

LandynDev:
> Thanks for the contribution. Can you add the ticket for the requirement? I am wondering the extension like AddStakeRecycleV1, who really want to do it. just confirm all is from the real requirements.

Just got it ticketed up here - https://github.com/opentensor/subtensor/issues/2564

Particular snippet from the issue to elaborate on the AddStakeBurn + AddStakeRecycle funcs and why those 2 are included:

> The current `AddStakeV1` chain extension only returns a success/error code, not the resulting alpha amount. While a contract could work around this by calling `AddStakeV1` followed by `RecycleAlphaV1(MAX)`, this recycles **all** alpha the contract holds on that subnet — not just the alpha from that specific stake. This is fine for simple contracts, but may break down for contracts that hold alpha for multiple purposes on the same subnet.
>                                                                                                                                                               
> The `AddStakeRecycleV1` and `AddStakeBurnV1` functions handle this atomically in a single call, recycling/burning only the alpha produced by that specific stake. This provides the robustness needed for a general-purpose chain extension interface.

open-junius:
Looks good to me, please fix the 'Cargo fmt'

open-junius (dismissed):

LandynDev:
> Looks good to me, please fix the 'Cargo fmt'

:+1: 
it dismissed your review bc of the repush, but should be good to go now

LandynDev:
it also needed `devnet-ready` rebase, got outdated but thats not situated

open-junius (dismissed):
Looks good for me.

chain-extensions/src/lib.rs
```diff
@@ -523,6 +523,172 @@ where
...
+                let (hotkey, amount, netuid): (T::AccountId, AlphaBalance, NetUid) =
+                    env.read_as()?;
+
+                let caller = env.caller();
+
+                let alpha_available =
+                    pallet_subtensor::Pallet::<T>::get_stake_for_hotkey_and_coldkey_on_subnet(
+                        &hotkey, &caller, netuid,
+                    );
+                let actual_amount = amount.min(alpha_available);
```
evgeny-s: We might not need this, because `do_recycle_alpha` will do this validation anyway.

chain-extensions/src/lib.rs
```diff
@@ -523,6 +523,172 @@ where
...
```
evgeny-s: The same

chain-extensions/src/lib.rs
```diff
@@ -523,6 +523,172 @@ where
...
+                let alpha = pallet_subtensor::Pallet::<T>::do_add_stake(
+                    RawOrigin::Signed(caller.clone()).into(),
+                    hotkey.clone(),
+                    netuid,
+                    tao_amount,
+                );
+
+                match alpha {
+                    Ok(alpha) => {
+                        let burn_result = pallet_subtensor::Pallet::<T>::burn_alpha(
```
evgeny-s: What will happen with the stake if this fails? We might want to either rollback for the stake in the Err or do it atomically.

chain-extensions/src/tests.rs
```diff
@@ -726,6 +726,534 @@ fn remove_proxy_success_removes_proxy_relationship() {
     });
 }
 
+#[test]
+fn recycle_alpha_success_reduces_stake_and_returns_actual_amount() {
```
evgeny-s, line 731: Let's also add a test for the rollback check

LandynDev:
Addressed the review feedback:
- Removed redundant `.min()` clamping in RecycleAlphaV1/BurnAlphaV1. Pallet handles this internally
- Wrapped `AddStakeRecycleV1/AddStakeBurnV1` in `with_transaction` for atomicity. If recycle/burn fails, stake is fully rolled back
- Added 2 rollback tests (`add_stake_recycle_rollback_on_recycle_failure`, `add_stake_burn_rollback_on_burn_failure`)

image

@evgeny-s / @open-junius if you could give this a re-look when you have a chance, thanks

Commit of changes: https://github.com/opentensor/subtensor/pull/2560/changes/65357d303af1317dd5513cde74d5c8c86454e32c

chain-extensions/src/lib.rs
```diff
@@ -523,6 +524,160 @@ where
...
+                let call_result = pallet_subtensor::Pallet::<T>::recycle_alpha(
+                    RawOrigin::Signed(caller).into(),
+                    hotkey,
+                    amount,
+                    netuid,
+                );
+
+                match call_result {
+                    Ok(_) => {
+                        env.write_output(&amount.encode())
```
open-junius: I can see you remove the real_amount, and put the amount into result directly. but it could be wrong if the amount is more than you staked. So I suggest we can update the recycle_alpha function with the real recycled alpha as result, or doesn't return any value in RecycleAlphaV1. and the same for burn function.

LandynDev: Good call, I got it updated w this commit - https://github.com/opentensor/subtensor/pull/2560/changes/60dffb39a6a67135e7fb263b09f9cc45f45af4f7

Trying to keep the diff concise with what i add/change. Reran unit tests and passing

chain-extensions/src/lib.rs
```diff
@@ -523,6 +525,159 @@ where
...
+                    <<T as pallet_subtensor::Config>::WeightInfo as SubtensorWeightInfo>::add_stake_burn();
+
+                env.charge_weight(weight)?;
+
+                let (hotkey, netuid, tao_amount): (T::AccountId, NetUid, TaoBalance) =
+                    env.read_as()?;
+
+                let caller = env.caller();
+
+                let result = transactional::with_transaction(|| {
```
open-junius: There is do_add_stake_burn function, can we use it instead of call both do_add_stake and do_burn_alpha

chain-extensions/src/lib.rs
```diff
@@ -523,6 +525,159 @@ where
...
+                }
+            }
+            FunctionId::AddStakeRecycleV1 => {
+                let weight =
+                    <<T as pallet_subtensor::Config>::WeightInfo as SubtensorWeightInfo>::add_stake()
+                        .saturating_add(
+                            <<T as pallet_subtensor::Config>::WeightInfo as SubtensorWeightInfo>::recycle_alpha(),
+                        );
+
+                env.charge_weight(weight)?;
```
open-junius: we can charge the add_stake part first. If do_add_stake is Ok, then continue with recycle_alpha weight

chain-extensions/src/lib.rs
LandynDev: Looked into this but I don't think `do_add_stake_burn` is usable from a chain ext
- It's `pub(crate)`, so it's not accessible from the `chain-extensions` crate.  
- Also it calls `ensure_subnet_owner` on the origin which we won't want

Because of this, I kept it as the `do_add_stake` + `do_burn_alpha` composition

chain-extensions/src/lib.rs
LandynDev: Okay, I updated this.. `AddStakeRecycleV1` now charges `add_stake()` upfront and only charges `recycle_alpha()` after do_add_stake returns Ok. Applied the same split to AddStakeBurnV1 for consistency. Both ops still run inside with_transaction, & tests updated and passing.

open-junius (dismissed):
Looks good to me

chain-extensions/src/lib.rs
```diff
@@ -523,6 +525,180 @@ where
 
                 Ok(RetVal::Converging(Output::Success as u32))
             }
+            FunctionId::RecycleAlphaV1 => {
+                let weight =
+                    <<T as pallet_subtensor::Config>::WeightInfo as SubtensorWeightInfo>::recycle_alpha();
+
+                env.charge_weight(weight)?;
+
+                let (hotkey, amount, netuid): (T::AccountId, AlphaBalance, NetUid) =
```
evgeny-s, line 534: Can we make arguments order consistent with AddStakeRecycleV1?

evgeny-s (commented):
I don't like duplicating the composition logic - `AddStakeBurnV1` re-implements `do_add_stake_burn` inline, and `AddStakeRecycleV1` introduces a brand-new atomic primitive that doesn't exist as a pallet function at all. Ideally, it would be great to have the functions in the pallet itself and wire them here, something like `do_add_stake_burn_permissionless`. I still not sure, why we need a guard for the subnet owner, if we can call both functions separately, but this question out of scope for this PR.

@shamil-gadelshin  - could you weigh in?

Per your `subnet_buyback` PR (#2381), the owner-check there is a privilege gate (priority boost + canonical event + per-tempo cadence), not a security control on the underlying effect.

Could you confirm:
1. That reading is correct, and the underlying atomic `add_stake + burn_alpha` composition is safe to expose permissionlessly via chain extension?
2. Are we ok bypassing the rate limit via calling 2 extrinsics separately?

evgeny-s:
> I don't like duplicating the composition logic - `AddStakeBurnV1` re-implements `do_add_stake_burn` inline, and `AddStakeRecycleV1` introduces a brand-new atomic primitive that doesn't exist as a pallet function at all. Ideally, it would be great to have the functions in the pallet itself and wire them here, something like `do_add_stake_burn_permissionless`. I still not sure, why we need a guard for the subnet owner, if we can call both functions separately, but this question out of scope for this PR.
> 
> @shamil-gadelshin - could you weigh in?
> 
> Per your `subnet_buyback` PR (#2381), the owner-check there is a privilege gate (priority boost + canonical event + per-tempo cadence), not a security control on the underlying effect.
> 
> Could you confirm:
> 
> 1. That reading is correct, and the underlying atomic `add_stake + burn_alpha` composition is safe to expose permissionlessly via chain extension?
> 2. Are we ok bypassing the rate limit via calling 2 extrinsics separately?

The composition function is needed for the priority boost for subnet owners, so it makes sense.

The architectural point about extracting `do_add_stake_burn_permissionless` / `do_add_stake_recycle` into the pallet still stands in my opinion. Let's try to extract those as a pallet function and re-use it in the extension, because from what I can see, so far, all the extension functions are thin.

shamil-gadelshin:
> The composition function is needed for the priority boost for subnet owners, so it makes sense.

Yes, otherwise it could be MEVed.

chain-extensions/src/lib.rs
LandynDev: Cleaned this up - https://github.com/opentensor/subtensor/pull/2560/commits/7488747b65386237878f23cc18e3d7cab25d03e8

LandynDev:
> > I don't like duplicating the composition logic - `AddStakeBurnV1` re-implements `do_add_stake_burn` inline, and `AddStakeRecycleV1` introduces a brand-new atomic primitive that doesn't exist as a pallet function at all. Ideally, it would be great to have the functions in the pallet itself and wire them here, something like `do_add_stake_burn_permissionless`. I still not sure, why we need a guard for the subnet owner, if we can call both functions separately, but this question out of scope for this PR.
> > @shamil-gadelshin - could you weigh in?
> > Per your `subnet_buyback` PR (#2381), the owner-check there is a privilege gate (priority boost + canonical event + per-tempo cadence), not a security control on the underlying effect.
> > Could you confirm:
> > 
> > 1. That reading is correct, and the underlying atomic `add_stake + burn_alpha` composition is safe to expose permissionlessly via chain extension?
> > 2. Are we ok bypassing the rate limit via calling 2 extrinsics separately?
> 
> The composition function is needed for the priority boost for subnet owners, so it makes sense.
> 
> The architectural point about extracting `do_add_stake_burn_permissionless` / `do_add_stake_recycle` into the pallet still stands in my opinion. Let's try to extract those as a pallet function and re-use it in the extension, because from what I can see, so far, all the extension functions are thin.

Okay, I pulled the composition into the pallet as `do_add_stake_recycle` and `do_add_stake_burn_permissionless` and the chain extension now just calls them, so both handlers are thin like the rest. Commit: https://github.com/opentensor/subtensor/pull/2560/changes/0179ac521c66ad9afd0f0096fa4ea2f32f99be57

evgeny-s (dismissed):

open-junius (dismissed):

sam0x17 (approved):

open-junius (approved):

evgeny-s (approved):

sam0x17 (approved):

PR #2570, opened 2026-04-07
Charge hotkey-origin fees to the owning coldkey

evgeny-s:
# Charge hotkey-origin fees to the owning coldkey (mechanism only)

## Why

Hotkeys are operational, online keys and typically hold little or no balance; coldkeys
hold the funds and stay in cold storage. When a hotkey submits a hotkey-origin extrinsic
(weight setting, commit/reveal, serving, …), the transaction fee should come from its
owning coldkey rather than the (often empty) hotkey — without weakening the hotkey/coldkey
security separation.

## What

Adds fee-payer redirection to the transaction-payment wrapper: for a defined set of
hotkey-origin calls signed by a hotkey that has an owner, the protocol fee is charged to
`Owner[hotkey]` (the coldkey) rather than the hotkey; any signer-chosen tip is excluded from
the coldkey's charge. Otherwise the signer pays. Existing proxy `RealPaysFee` resolution keeps
precedence.

**This PR is the mechanism only.** All targeted calls remain `Pays::No`, so a redirected call
charges the coldkey `inclusion_fee (0) + tip (0) = 0` — no economic effect yet. It stays
dormant until a follow-up flips `Pays` per call.

## How

- **`ColdkeyFeeCallFilter<Call>` trait** (`runtime/src/transaction_payment_wrapper.rs`) —
  runtime-supplied policy that keeps the concrete allow-list out of the generic extension.
- **`ColdkeyPaysFeeCalls`** — a `call_filter_group!` in `runtime/src/fee_filters.rs`
  enumerating the 17 hotkey-origin calls (weight set/commit/reveal families, `serve_*`,
  `associate_evm_key`, `Commitments::set_commitment`); `Runtime` implements the trait via
  `ColdkeyPaysFeeCalls::contains`. Kept independent of the proxy call groups.
- **Wrapper `validate`** — resolves the payer as: `extract_real_fee_payer` (proxy) first, then
  `extract_coldkey_fee_payer` (group membership + `Owner::try_get`), else the signer. `try_get`
  makes an unassociated signer fall back safely. When the coldkey pays, the fee is recomputed
  with the tip zeroed.

## Security

- **Never charge a signer-chosen tip to the coldkey.** The tip is set by the signing hotkey,
  and `Pays::No` charges `final_fee = tip`, so billing a tipped call to the coldkey would let a
  hotkey drain it without any coldkey signature or opt-in. The coldkey pays the protocol fee
  only — the tip is dropped from its charge. This is invisible to senders because a tip does
  not buy priority in this wrapper (priority is overridden by dispatch class), so auto-tipping
  wallets keep working.
- **Explicit allow-list**, not "any call from a hotkey": bounds the blast radius so the coldkey
  only ever pays the protocol fee for the enumerated hotkey operations, never arbitrary calls.

## Testing

25 unit tests co-located in the extension source: coldkey selection for `set_weights` and a
second group member (`commit_weights`), no-owner fallback, out-of-group exclusion, the tip
guard (a tipped group call is billed to the coldkey with the tip excluded), plus the existing
proxy `RealPaysFee` suite. Assertions are at the `validate` level (payer selection), which
holds under `Pays::No`.

## Scope / follow-up

- Flip `Pays::No → Pays::Yes` per call (gradual rollout) — the actual economic change.
- Restore the balance-based runtime test and the TypeScript e2e test (both require a
  non-zero fee, so they belong with the `Pays` flip).


runtime/src/transaction_payment_wrapper.rs
```diff
@@ -230,6 +237,8 @@ where
         // Otherwise, the signer pays as usual.
         let fee_origin = if let Some(real) = Self::extract_real_fee_payer(call, &origin) {
             frame_system::RawOrigin::Signed(real).into()
+        } else if let Some(coldkey) = Self::extract_coldkey_fee_payer(&origin) {
+            frame_system::RawOrigin::Signed(coldkey).into()
```
l0r1s: On top of that, we should check on the `DispatchInfo` if the call is `Pays::No`

evgeny-s: For `Pays::No` extrinsics, fees won't be charged anyway.

runtime/src/transaction_payment_wrapper.rs
```diff
@@ -182,14 +185,28 @@ where
...
+
+    fn extract_coldkey_fee_payer(origin: &RuntimeOriginOf<T>) -> Option<AccountIdOf<T>> {
+        let signer = origin.as_system_origin_signer()?;
+
+        pallet_subtensor::Pallet::<T>::maybe_coldkey_for_hotkey(signer)
+    }
+
+    fn is_coldkey_fee_payer_eligible(call: &RuntimeCallOf<T>) -> bool {
+        match call.is_sub_type() {
+            Some(pallet_subtensor::Call::set_weights { .. }) => true,
```
evgeny-s: @l0r1s , added the additional validation for the list of extrinsics we want to apply this to.

pallets/subtensor/src/rpc_info/delegate_info.rs
```diff
@@ -188,4 +188,8 @@ impl<T: Config> Pallet<T> {
     pub fn get_coldkey_for_hotkey(hotkey: &T::AccountId) -> T::AccountId {
         Owner::<T>::get(hotkey)
     }
+
+    pub fn maybe_coldkey_for_hotkey(hotkey: &T::AccountId) -> Option<T::AccountId> {
+        Owner::<T>::try_get(hotkey).ok()
+    }
```
l0r1s, line 194: That is super weird that there is a `DefaultAccount` for a `Owner`, it should probably be an `OptionQuery` but that is something for another discussion

runtime/src/transaction_payment_wrapper.rs
```diff
@@ -230,6 +247,12 @@ where
         // Otherwise, the signer pays as usual.
         let fee_origin = if let Some(real) = Self::extract_real_fee_payer(call, &origin) {
             frame_system::RawOrigin::Signed(real).into()
+        } else if Self::is_coldkey_fee_payer_eligible(call) {
+            if let Some(coldkey) = Self::extract_coldkey_fee_payer(&origin) {
+                frame_system::RawOrigin::Signed(coldkey).into()
+            } else {
+                origin.clone()
+            }
```
l0r1s, line 255: Can't just have extract_coldkey_fee_payer returns None when not eligible directly in the same function to collapse both into one?

Make things cleaner, wdyt?


```suggestion
        } else if let Some(coldkey) Self::extract_coldkey_fee_payer(call, &origin) {
                frame_system::RawOrigin::Signed(coldkey).into()
```

runtime/tests/common/mod.rs
```diff
@@ -0,0 +1,139 @@
+#![allow(clippy::arithmetic_side_effects, clippy::unwrap_used)]
+
+use {
```
l0r1s, line 3: This is subtensor pallet specific if we think about it. Maybe exposing the mock from the subtensor pallet make it easier?

open-junius (approved):

UnArbosFour (approved):

UnArbosSix (approved):

PR #2578, opened 2026-04-09
Fix orphaned storage cleanup on subnet deregistration

### LINKED ISSUE #2573: Handle AllowancesStorage when subnet deregistration

AllowancesStorage is introduced in #2478, but not handled during subnet deregistration. 

It is a double map, the netuid is part of second key. need to consider how to clean up efficiently.

_No response_

### Additional context

_No response_

fury0928:
## Description

Clean up `AllowancesStorage`, `RootAlphaDividendsPerSubnet`, and 12 additional storage items that were not being removed during subnet deregistration in `remove_network`.

**AllowancesStorage** (EVM staking precompile) uses netuid embedded in the second key `(spender, netuid)`, so it cannot use `clear_prefix`. A new `PrecompileCleanupInterface` trait (following the existing `CommitmentsInterface` pattern) is added to handle this cross-crate cleanup.

**RootAlphaDividendsPerSubnet** and **VotingPower** have netuid as the first key and are cleaned via `clear_prefix`.

**11 additional simple StorageMaps** were also missing removal: `MinAllowedUids`, `MaxWeightsLimit`, `AdjustmentAlpha`, `AdjustmentInterval`, `MinNonImmuneUids`, `RootProp`, `RecycleOrBurn`, `RootClaimableThreshold`, `VotingPowerTrackingEnabled`, `VotingPowerDisableAtBlock`, `VotingPowerEmaAlpha`.

## Related Issue(s)

- Closes #2573

## Type of Change

- [ ] Documentation update

## Breaking Change

N/A — this only adds cleanup of storage that was previously leaked on subnet deregistration.

## Checklist

## Additional Notes

The `PrecompileCleanupInterface` trait follows the same pattern as the existing `CommitmentsInterface` — a trait defined in `pallet-subtensor`, implemented in the precompiles crate, and wired through the runtime `Config`. This avoids a circular dependency since precompiles depends on `pallet-subtensor`, not the reverse.

### DIFF
--- a/pallets/subtensor/src/tests/networks.rs
+++ b/pallets/subtensor/src/tests/networks.rs
@@ -2242,6 +2242,74 @@ fn dissolve_clears_all_mechanism_scoped_maps_for_all_mechanisms() {
     });
 }
 
+#[test]
+fn test_dissolve_network_clears_orphaned_storage() {
+    new_test_ext(0).execute_with(|| {
+        let cold = U256::from(1);
+        let hot = U256::from(2);
+        let net = add_dynamic_network(&hot, &cold);
+
+        let hotkey1 = U256::from(100);
+        let hotkey2 = U256::from(200);
+
+        // Seed RootAlphaDividendsPerSubnet (DMAP, netuid is first key).
+        RootAlphaDividendsPerSubnet::<Test>::insert(net, hotkey1, AlphaBalance::from(5_000u64));
+        RootAlphaDividendsPerSubnet::<Test>::insert(net, hotkey2, AlphaBalance::from(10_000u64));
+
+        // Seed VotingPower (DMAP, netuid is first key).
+        VotingPower::<Test>::insert(net, hotkey1, 42u64);
+
+        // Seed simple StorageMaps that were previously missed.
+        MinAllowedUids::<Test>::insert(net, 4u16);
+        MaxWeightsLimit::<Test>::insert(net, 100u16);
+        AdjustmentAlpha::<Test>::insert(net, 999u64);
+        AdjustmentInterval::<Test>::insert(net, 50u16);
+        MinNonImmuneUids::<Test>::insert(net, 2u16);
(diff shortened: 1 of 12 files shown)


precompiles/src/staking.rs
```diff
@@ -74,6 +74,25 @@ pub type AllowancesStorage = StorageDoubleMap<
     ValueQuery,
 >;
 
+/// Remove all allowance entries for the given `netuid`.
+///
+/// Because `netuid` is embedded in the second key `(spender, netuid)`, we must iterate
+/// all entries and filter. Called during subnet deregistration.
+pub fn purge_allowances_for_netuid(netuid: u16) {
```
plind-junior, line 81: This approach is already mentioned at #2478

eureka0928: Thanks for the context! Yes, the concern about unbounded iteration was raised in #2478. In practice though, `remove_network` already does the same iterate-and-filter pattern for several other DMAPs where netuid is not the first key (`ChildkeyTake`, `ChildKeys`, `ParentKeys`, `LastHotkeyEmissionOnNetuid`, `TransactionKeyLastBlock`, `StakingOperationRateLimiter`), so this is consistent with the existing approach.

The number of allowance entries per netuid should also be small in practice since each requires an explicit `approve` call from an EVM user.

pallets/subtensor/src/tests/networks.rs
```diff
@@ -2242,6 +2242,74 @@ fn dissolve_clears_all_mechanism_scoped_maps_for_all_mechanisms() {
     });
 }
 
+#[test]
+fn test_dissolve_network_clears_orphaned_storage() {
+    new_test_ext(0).execute_with(|| {
+        let cold = U256::from(1);
```
open-junius, line 2248: can you add the check AllowancesStorage is removed after subnet de-reg

plind-junior (commented):
-

PR #2582, opened 2026-04-10
Bottom half median owner alpha stake

JohnReedV:
## Summary

This PR updates **subnet registration** so the **new subnet pool and the owner’s starting stake are both initialized from the same snapshot-priced registration seed**, using the **bottom-half median subnet alpha price**.

- recycled the remainder,
- and set the subnet owner metadata.

With this change, registration now:
- computes the **bottom-half median subnet alpha price** from existing eligible subnets,
- injects **25% of the lock cost as TAO** into the new subnet pool,
- mints **snapshot-priced alpha against that seeded TAO amount**,
- injects that alpha into the pool as free alpha liquidity,
- grants the **same alpha amount again** to the subnet owner as **staked alpha**,
- and recycles the remaining locked TAO that is not used to seed the pool.

> **In one line:**  
> **Registering a subnet now seeds the pool with 25% of the lock cost in TAO, prices the pool alpha off the bottom-half median subnet market price, and grants the owner an equal amount of snapshot-priced alpha as initial stake.**

---

### DIFF
--- a/pallets/subtensor/src/staking/stake_utils.rs
+++ b/pallets/subtensor/src/staking/stake_utils.rs
@@ -70,73 +70,68 @@ impl<T: Config> Pallet<T> {
     }
 
     /// Gets the Median Subnet Alpha Price
-    pub fn get_median_subnet_alpha_price() -> U96F32 {
+    pub fn get_bottom_half_median_subnet_alpha_price() -> U96F32 {
         let default_price = U96F32::saturating_from_num(1_u64);
         let zero_price = U96F32::saturating_from_num(0_u64);
         let two = U96F32::saturating_from_num(2_u64);
 
-        let mut price_counts: BTreeMap<U96F32, usize> = BTreeMap::new();
-        let mut total_prices: usize = 0;
+        let mut prices: Vec<U96F32> = NetworksAdded::<T>::iter()
+            .filter_map(|(netuid, added)| {
+                if added && netuid != NetUid::ROOT {
+                    let price = T::SwapInterface::current_alpha_price(netuid);
+                    if price > zero_price {
+                        Some(price)
+                    } else {
+                        None
+                    }
+                } else {
+                    None
+                }
+            })
+            .collect();
 
-        for (netuid, added) in NetworksAdded::<T>::iter() {
-            if !added || netuid == NetUid::ROOT {
-                continue;
-            }
-
-            let price = T::SwapInterface::current_alpha_price(netuid);
-            if price <= zero_price {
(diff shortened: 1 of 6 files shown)


pallets/subtensor/src/staking/stake_utils.rs
```diff
@@ -70,73 +70,68 @@ impl<T: Config> Pallet<T> {
     }
 
     /// Gets the Median Subnet Alpha Price
-    pub fn get_median_subnet_alpha_price() -> U96F32 {
+    pub fn get_bottom_half_median_subnet_alpha_price() -> U96F32 {
         let default_price = U96F32::saturating_from_num(1_u64);
```
ales-otf, line 74: if it's something that could be reused, it's better to make it a constant instead of hardcoding, to simplify refactoring in case we would need to change the value.

pallets/subtensor/src/staking/stake_utils.rs
```diff
@@ -70,73 +70,68 @@ impl<T: Config> Pallet<T> {
...
-
-            if let Some(count) = price_counts.get_mut(&price) {
-                *count = count.saturating_add(1);
-            } else {
-                price_counts.insert(price, 1usize);
-            }
-        }
+        let mut prices: Vec<U96F32> = NetworksAdded::<T>::iter()
+            .filter_map(|(netuid, added)| {
+                if added && netuid != NetUid::ROOT {
```
ales-otf, line 80: `!netuid.is_root()` could be used here instead

ales-otf: also match or reverse `if` (guard return) would improve readability, but it's probably a personal preference.

pallets/subtensor/src/staking/stake_utils.rs
```diff
@@ -70,73 +70,68 @@ impl<T: Config> Pallet<T> {
...
+        let bottom_half_prices: Vec<U96F32> = prices.into_iter().take(bottom_half_len).collect();
+        let bottom_len = bottom_half_prices.len();
+
+        let Some(mid_index) = bottom_len.checked_div(2) else {
             return default_price;
         };
-        let Some(upper_target) = total_prices.checked_div(2) else {
+
+        let Some(remainder) = bottom_len.checked_rem(2) else {
             return default_price;
```
ales-otf, line 113: since your failure cases all return `default_price`, you could wrap the code into a function returning an `Option` and then just use `unwrap_or`

ales-otf (approved):

PR #2589, opened 2026-04-14
Make instant seal work and added interval

girazoki:
## Description
This PR fixes Instant seal which was not working because `consensus_providers` were not being passed and also adds interval sealing, which will create a block at the defined interval.

Instant sealing works better when offchain workers are disabled, as the offchain worker creates txs by itself.

Example usages:

`./target/release/node-subtensor --dev --tmp --sealing=instant --offchain-worker never
`

`./target/release/node-subtensor --dev --tmp --sealing=6000
`

- [ ] Documentation update
- [x] Other (please describe):
More flexibility when testing

## Checklist

- [X] I have run `./scripts/fix_rust.sh` to ensure my code is formatted and linted correctly

## Additional Notes


node/src/service.rs
```diff
@@ -778,32 +787,47 @@ fn run_manual_seal_authorship(
...
-                select_chain,
-                consensus_data_provider: None,
-                create_inherent_data_providers,
-            },
-        )),
+    let seal_stream: SealStream = match sealing {
+        Sealing::Manual => Box::pin(commands_stream),
+        Sealing::Instant => Box::pin(transaction_pool.import_notification_stream().map(|_| {
+            sc_consensus_manual_seal::rpc::EngineCommand::SealNewBlock {
+                create_empty: false,
```
lorismoulin, line 794: [Original behavior is `create_empty: true`](https://github.com/opentensor/polkadot-sdk/blob/8e64bb2e4d7554a813fe5a99805c9898123c1366/substrate/client/consensus/manual-seal/src/lib.rs#L237) for instant seal, should we preserve this?

lorismoulin (approved):
Added a small comment, not sure if it changes anything in our case though.

open-junius (approved):
verified the command, it works. Looks good!

node/src/service.rs
gorka-i: that was the reason why it wasnt working from what I had seen, but I did not dig too much into the issue. When you think about it instant means "as soon as there is a transaction" which by default means a non-empty block. before this PR this was not working but now is

PR #2594, opened 2026-04-16
Owner Alpha Stake Updates

JohnReedV:
## Summary

This PR updates **subnet registration** so the new subnet is initialized from a **pre-registration median subnet alpha price snapshot**, while the registration lock is retained in the subnet reserve instead of being mostly recycled.

- initialized the pool at a 1:1 reserve ratio,
- recycled the remainder.

The owner did **not** receive any initial alpha stake from the registration flow.

With this change:
- the current **median subnet alpha price** is snapshotted before the new subnet is created,
- the subnet TAO reserve is set to `max(registration_lock, network_min_lock)`,
- the subnet alpha reserve is sized from that TAO reserve at the snapshotted median price,
- the owner receives **separately minted staked alpha** at that same median price,
- and the recycled registration TAO is now normally zero.

> **In one line:**  
> **Registering a subnet now retains the registration lock in the subnet reserve, opens the subnet at the median subnet alpha price, and grants the owner initial staked alpha priced off that same median.**

---

### DIFF
--- a/pallets/subtensor/src/subnets/subnet.rs
+++ b/pallets/subtensor/src/subnets/subnet.rs
@@ -209,23 +209,37 @@ impl<T: Config> Pallet<T> {
         let symbol = Self::get_next_available_symbol(netuid_to_register);
         TokenSymbol::<T>::insert(netuid_to_register, symbol);
 
-        // Seed the new subnet pool at a 1:1 reserve ratio.
-        // Separately, grant the subnet owner outstanding alpha based on the TAO they actually spent
-        // on registration converted by the current median subnet alpha price.
+        // Keep the locked TAO in the pool instead of recycling the excess.
+        // Mint the owner alpha separately at the median subnet alpha price.
+        // Size the pool alpha reserve from the total TAO reserve at that same price.
         let pool_initial_tao: TaoBalance = Self::get_network_min_lock();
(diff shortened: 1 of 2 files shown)


evgeny-s (approved):

pallets/subtensor/src/subnets/subnet.rs
```diff
@@ -209,36 +209,47 @@ impl<T: Config> Pallet<T> {
...
 
-        // Seed the new subnet pool at a 1:1 reserve ratio.
-        // Separately, grant the subnet owner outstanding alpha based on the TAO they actually spent
-        // on registration converted by the current median subnet alpha price.
+        // Keep the locked TAO in the pool instead of recycling the excess.
+        // Mint the owner alpha separately at the median subnet alpha price.
+        // Size the pool alpha reserve from the total TAO reserve at that same price.
         let pool_initial_tao: TaoBalance = Self::get_network_min_lock();
-        let pool_initial_alpha: AlphaBalance = pool_initial_tao.to_u64().into();
+        let total_pool_tao: TaoBalance = if actual_tao_lock_amount >= pool_initial_tao {
```
open-junius, line 216: It is much easier to understand if we set total_pool_tao, owner_alpha_tao_equivalent and tao_recycled_for_registration in the if/else statement.

sam0x17 (approved):

PR #2595, opened 2026-04-16
Port alpha and address mapping precompiles tests to rust

ales-otf:
## Description

Ports some of E2E contract ts tests to rust, because they only checked the contracts on the runtime - not the API-node boundary.

## Related Issue(s)

- Closes #[issue number]

## Type of Change

- [ ] Documentation update

## Breaking Change

## Checklist

## Additional Notes


runtime/tests/precompiles.rs
```diff
@@ -54,6 +59,52 @@ fn addr_from_index(index: u64) -> H160 {
...
+/// Encodes a selector plus one uint16 ABI argument.
+fn call_data_u16(signature: &str, value: u16) -> Vec<u8> {
+    let mut input = Vec::with_capacity(4 + 32);
+    input.extend_from_slice(&selector(signature));
+    push_abi_word(&mut input, U256::from(value));
+    input
+}
+
+/// Encodes a selector plus `(uint16,uint64)` ABI arguments.
+fn call_data_u16_u64(signature: &str, first: u16, second: u64) -> Vec<u8> {
```
open-junius, line 92: Can we introduce any EVM lib to get calldata. We will need lots of functions for each different parameters type.

open-junius (approved):
Looks good to me. just one trivial comment.

shamil-gadelshin (changes requested):
We're going to migrate a lot of TS tests and it makes sense to introduce a structure for the tests. Please split `precompiles.rs` into several files with proper naming - shared, alpha, addressMapping or similar.

runtime/tests/precompiles.rs
ales-otf: i'll introduce more with more tests porting and we can merge all of them into a library afterwards if you want. maybe precompiles crate would be a better place for these functions then

ales-otf:
> We're going to migrate a lot of TS tests and it makes sense to introduce a structure for the tests. Please split precompiles.rs into several files with proper naming - shared, alpha, addressMapping or similar.

I intentionally put it into a single file for now, but split modules inside, because it's easier to navigate. if/when it will be huge, i'll break it up into modules, but so far I disagree that it's better.

oh and since it's integration tests, it won't be hierarchical - only flat. so we would need a lot of `precompile_*` as siblings to other tests.

shamil-gadelshin (approved):

PR #2599, opened 2026-04-17
Convictions (stake locks)

gztensor:
## Description

- Implements an exponential lock mechanism where coldkeys can lock alpha stake to a specific hotkey per subnet, building conviction over time.
- Lock state uses lazy evaluation: locked_mass stays constant over time and conviction grows as=t maturity rate calculated as c1 = c0 + (m - c0) * exp(-dt/tau), where c is conviction, and m is locked mass.
- Enforces lock invariant on unstake (validate_remove_stake) and move/transfer (validate_stake_transition): total coldkey alpha on subnet must remain >= current locked amount.
- New extrinsic `lock_stake` with `MaturityRate` configurable decay timescale (default half-life ~90 days).
- User can unlock X amount of locked stake (with new extrinsic `unlock_stake`), which immediately drops conviction by X, but does not immediately allow to unstake.
- The unlocked stake will become for the user to unstake or to lock again gradually over time with exponential decay rate of UnlockRate (default half-live ~30 days).

Updates:

- Request 2026-04-19 (const): Can change the hotkey of a lock but when you do it resets the conviction. 
- 2026-04-21: Implementation decision - coldkey swap will fail if new coldkey already has any locks. Hotkey swap will reset conviction.

### Known gaps

- [x] 1. **`do_recycle_alpha` / `do_burn_alpha`** bypass lock - calls `decrease_stake_for_hotkey_and_coldkey_on_subnet` without lock check.
- [x] 2. **`clear_small_nomination_if_required`** bypasses unlock period.
- [x] 3. **Subnet dissolution** orphans `Lock` entries - `destroy_alpha_in_out_stakes` removes all Alpha but not Lock storage; stale locks persist across netuid reuse.
- [x] 4. **Coldkey swap** does not migrate Lock entries.
- [x] 5. **Hotkey swap** does not update Lock entries.
- [x] 6. `subnet_king` needs optimization to avoid O(n) Lock iteration (another map possibly).

### Implementation checklist

- [x] Hotkey conviction RPC
- [x] Allow moving from one hotkey to another without resetting conviction if coldkey owns both hotkeys
- [x] Extrinsic to switch hotkey in the lock (with loss of conviction)
- [x] Do not store zero entries in Lock and HotkeyLock maps
- [x] Bake in flexibility in state maps to lock to multiple hotkeys, enforce 1-1 lock in code
- [x] Unlocking behavior (exp decay of unlocks)
- [x] Add stake availability check in lock extrinsic
- [x] Tests for unlocking:
  - [x] Cannot unstake immediately
  - [x] Cannot lock immediately
  - [x] Can unstake X after waiting Y
  - [x] Can lock X after waiting Y, in which case cannot unstake after locked again
  - [x] Can transfer or move, but the stake stays unavailable
- [x] More tests
  - [x] Cannot double-lock

### Questions

- Does hotkey account need to exist in order for coldkeys to be able to lock to it? Currently it is not enforced.

## Type of Change
- [ ] Documentation update

## Checklist


pallets/subtensor/src/staking/lock.rs
```diff
@@ -0,0 +1,578 @@
...
+
+        Ok(())
+    }
+
+    pub fn do_unlock_stake(
+        coldkey: &T::AccountId,
+        netuid: NetUid,
+        amount: AlphaBalance,
+    ) -> dispatch::DispatchResult {
+        ensure!(!amount.is_zero(), Error::<T>::AmountTooLow);
```
JohnReedV: I think we need to assert that they actually have the lock amount they are trying to remove. Looks like a user could just remove it all.

JohnReedV (changes requested):

pallets/subtensor/src/staking/lock.rs
gztensor: The user should be able to remove all. With new logic it doesn't mean it will all be available to unstake. Good point about the check though, I'll make the change.

sam0x17 (dismissed):

JohnReedV (approved):

sam0x17 (approved):

PR #2601, opened 2026-04-20
feat(subtensor): reset subnet identity on subnet create

### LINKED ISSUE #2572: Reset subnet identity upon subnet create

At present, when a new subnet is created and deregisters the previous owner key in the slot, the subnet identity information (if set by the previous owner) is not cleared.  This can lead to confusion among participants as to what the subnet actually is about.

I recommend that if a new subnet owner does not provide subnet identity information during the creation of the subnet, that subnet creation automatically reset the subnet's identity information (or clears out any identity information that the new subnet owner has not explicitly provided during the process of registering the new subnet).

_No response_

### Additional context

_No response_

ArtificialXai:
## Summary

Closes #2572.

`do_register_network` previously only *set* a `SubnetIdentityV3` entry when the caller passed `Some(identity)` and left the storage map untouched otherwise. Storage migrations and legacy code paths can leave an orphan `SubnetIdentitiesV3` entry on a netuid slot that later gets assigned to a new owner; that orphan then leaks into the fresh subnet and misrepresents what the subnet is about (wrong name, wrong GitHub repo, wrong Discord, etc.).

This PR makes the `identity` argument authoritative, matching the natural intuition of "the caller of `register_network` fully owns the resulting subnet's presentation":

- `Some(identity)` — validate and insert (existing behavior, unchanged).
- `None` — if an entry already exists for the target netuid, remove it and emit `SubnetIdentityRemoved`; no-op if the slot is already empty.

No runtime migration required — the fix is forward-only and only kicks in when `do_register_network` is called, so existing subnets are unaffected.

## Changes

- `pallets/subtensor/src/subnets/subnet.rs` — step 17 of `do_register_network` now uses a `match identity` instead of `if let Some(..)`, with a `None` branch that clears any stale `SubnetIdentitiesV3` entry and emits the existing `SubnetIdentityRemoved` event (already used by `do_dissolve_network` in `coinbase/root.rs`).
- `pallets/subtensor/src/tests/networks.rs` — three regression tests:
  - `register_network_with_none_identity_clears_stale_entry` — pre-inserts a stale `SubnetIdentityV3` at the predicted next netuid, calls `do_register_network(..., None)`, asserts the entry is gone.
  - `register_network_with_some_identity_overwrites_stale_entry` — same setup but passes `Some(new_identity)`, asserts the stored entry equals `new_identity`.
  - `register_network_with_none_identity_no_op_when_slot_empty` — registers with `None` against an empty slot, asserts no `SubnetIdentitiesV3` entry is created.

## Test plan

- [x] `cargo check -p pallet-subtensor`
- [ ] `cargo test -p pallet-subtensor --features pow-faucet -- tests::networks::register_network_with_none_identity_clears_stale_entry`
- [ ] `cargo test -p pallet-subtensor --features pow-faucet -- tests::networks::register_network_with_some_identity_overwrites_stale_entry`
- [ ] `cargo test -p pallet-subtensor --features pow-faucet -- tests::networks::register_network_with_none_identity_no_op_when_slot_empty`
- [ ] Full `cargo test -p pallet-subtensor` (CI)

## Notes

- `SubnetIdentityRemoved` event is already emitted by `do_dissolve_network` when a subnet is torn down, so consumers that index events should already handle it gracefully.
- The new `None`-branch behavior only *removes* stale state; it can never overwrite a caller's current subnet identity, since it runs inside `do_register_network` which creates the subnet.
- Fixes the user-visible symptom of a newly registered subnet appearing with a previous owner's `subnet_name` / `github_repo` / `discord` until the new owner manually calls `set_subnet_identity`.

### DIFF
--- a/pallets/subtensor/src/subnets/subnet.rs
+++ b/pallets/subtensor/src/subnets/subnet.rs
@@ -235,15 +235,30 @@ impl<T: Config> Pallet<T> {
             Self::increase_total_stake(pool_initial_tao);
         }
 
-        // --- 17. Add the identity if it exists
-        if let Some(identity_value) = identity {
-            ensure!(
-                Self::is_valid_subnet_identity(&identity_value),
-                Error::<T>::InvalidIdentity
-            );
+        // --- 17. Set or reset the subnet identity.
+        //
+        // If the caller provided identity info, validate and store it.
+        // Otherwise clear any identity that may still be attached to this
+        // netuid slot from a previous owner — storage migrations or older
+        // code paths can leave orphaned `SubnetIdentitiesV3` entries which
+        // would otherwise leak into the newly-created subnet and mislead
+        // participants about what the subnet is about (see issue #2572).
+        match identity {
+            Some(identity_value) => {
+                ensure!(
+                    Self::is_valid_subnet_identity(&identity_value),
+                    Error::<T>::InvalidIdentity
+                );
 
-            SubnetIdentitiesV3::<T>::insert(netuid_to_register, identity_value);
-            Self::deposit_event(Event::SubnetIdentitySet(netuid_to_register));
+                SubnetIdentitiesV3::<T>::insert(netuid_to_register, identity_value);
+                Self::deposit_event(Event::SubnetIdentitySet(netuid_to_register));
+            }
+            None => {
+                if SubnetIdentitiesV3::<T>::contains_key(netuid_to_register) {
+                    SubnetIdentitiesV3::<T>::remove(netuid_to_register);
+                    Self::deposit_event(Event::SubnetIdentityRemoved(netuid_to_register));
+                }
+            }
         }
 
         // --- 18. Emit the NetworkAdded event.
--- a/pallets/subtensor/src/tests/networks.rs
+++ b/pallets/subtensor/src/tests/networks.rs
@@ -1435,6 +1435,123 @@ fn register_network_fails_before_prune_keeps_existing() {
     });
 }
 
+// Regression for https://github.com/opentensor/subtensor/issues/2572:
+// a stale `SubnetIdentitiesV3` entry on a netuid slot that is about to be
+// assigned to a new subnet must be cleared when the new owner does not
+// provide identity information of their own.
+#[test]
+fn register_network_with_none_identity_clears_stale_entry() {
+    new_test_ext(0).execute_with(|| {
+        let expected_netuid = SubtensorModule::get_next_netuid();
+
+        let stale_identity = SubnetIdentityOfV3 {
+            subnet_name: b"Previous Owner".to_vec(),
+            github_repo: b"https://github.com/previous/subnet".to_vec(),
+            subnet_contact: b"previous@example.com".to_vec(),
+            subnet_url: b"https://previous.example.com".to_vec(),
+            discord: b"previous#1234".to_vec(),
+            description: b"left over from a prior owner".to_vec(),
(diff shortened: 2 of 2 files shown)


open-junius:
can you identity the "Storage migrations and legacy code paths can leave an orphan SubnetIdentitiesV3 entry". then we fix it, instead of add the check in registration.

ArtificialXai:
Signed and re-pushed (`fd2be3f`).

ArtificialXai:
@open-junius Thanks for pushing back — chased it down, and the orphan source is `migrate_subnet_identities_to_v3`.

**Introduced in `8bcfa2ae9` (2025-06-12, "feat: add logo_url to subnet identities"), file `pallets/subtensor/src/migrations/migrate_subnet_identities_to_v3.rs`:**

```rust
let old_subnet_identities = SubnetIdentitiesV2::<T>::iter().collect::<Vec<_>>();
for (netuid, old_subnet_identity) in old_subnet_identities.clone() {
    let new_subnet_identity = SubnetIdentityV3 { /* copy fields, blanks for new ones */ };
    SubnetIdentitiesV3::<T>::insert(netuid, &new_subnet_identity);
    SubnetIdentitiesV2::<T>::remove(netuid);
}
```

The loop copies **every** `SubnetIdentitiesV2` entry into V3 unconditionally. It never checks `NetworksAdded::<T>::get(netuid)` or `SubnetOwner::<T>::contains_key(netuid)`, so any V2 entry that was already orphaned at migration time gets faithfully carried forward into V3.

V2 had the same hazard before it: older dissolve paths didn't scrub `SubnetIdentitiesV2` (the V3 scrub in `do_dissolve_network` only lands in the same `8bcfa2ae9` that introduces V3 — for V2, dissolve never cleared identity). So any subnet dissolved before 2025-06-12 left its V2 entry behind, and the V3 migration then promoted that orphan.

`4999dcbe21` (2025-07-08, "fix: run subnet identity v3 migration (and ignore if already exists)") re-ran the migration after a skip, which would have re-promoted any V2 entries recreated in the interim if that path existed — but the primary leak is still the unconditional copy.

**Given this, the surgical source-level fix you're asking for is a one-shot cleanup migration** that iterates `SubnetIdentitiesV3` and removes entries where `NetworksAdded::<T>::get(netuid).unwrap_or(false) == false`. Something like:

```rust
pub fn migrate_clear_orphan_subnet_identities_v3<T: Config>() -> Weight {
    let migration_name = b"migrate_clear_orphan_subnet_identities_v3".to_vec();
    let mut weight = T::DbWeight::get().reads(1);
    if HasMigrationRun::<T>::get(&migration_name) { return weight; }

    let netuids: Vec<_> = SubnetIdentitiesV3::<T>::iter_keys().collect();
    for netuid in netuids {
        weight = weight.saturating_add(T::DbWeight::get().reads(1));
        if !NetworksAdded::<T>::get(netuid) {
            SubnetIdentitiesV3::<T>::remove(netuid);
            Pallet::<T>::deposit_event(Event::SubnetIdentityRemoved(netuid));
            weight = weight.saturating_add(T::DbWeight::get().writes(1));
        }
    }
    HasMigrationRun::<T>::insert(&migration_name, true);
    weight.saturating_add(T::DbWeight::get().writes(1))
}
```

Happy to fold that migration into this PR and drop the `None`-path clear in `do_register_network` if you'd rather land the source-level fix alone. Or keep both (migration + defensive clear at registration) as belt-and-braces against a future V3→V4 migration reintroducing the same class of bug — that decision is yours. Let me know which shape you'd like and I'll push.

PR #2198, opened 2025-11-12
Governance

lorismoulin:
## Summary

Adds an on-chain governance stack to Subtensor: a two-stage referendum flow backed by signed collective voting, three new generic pallets, and the runtime wiring that turns them into Subtensor's concrete governance model.

Root calls now reach dispatch only after passing through both stages, replacing the current sudo-style path with a structured, auditable process.

This `governance` PR is the umbrella / top-of-stack description for the full governance work. The implementation is being split into smaller reviewable PRs, but this description should explain the complete model and link the split PRs together.

Technical explanation:
https://governance-visualization.a5ede37144f01bdc6a27.workers.dev

Non-technical explanation:
https://gov-novelty.a5ede37144f01bdc6a27.workers.dev

## Split PR Stack

The full governance change is intentionally split into four PRs so each piece can be reviewed independently. The first three PRs add their pallets to the codebase only; they do not wire those pallets into the runtime. The final PR wires the stack together and adds the end-to-end tests.

- `pallet-multi-collective` (this PR #2198) — generic collective membership pallet. Codebase-only, not runtime-wired.
- #2713 — signed aye/nay voting backend. Codebase-only, not runtime-wired.
- #2714 — generic track-based referendum state machine. Codebase-only, not runtime-wired.
- #2715 — runtime wiring + E2E PR — wires the pallets into Subtensor, defines the concrete governance model, and adds the TypeScript E2E coverage.

## Governance Flow

1. A member of the `Proposers` collective submits a root call on the Triumvirate track.
2. The three-member `Triumvirate` votes to approve or reject.
3. On approval, the proposal is handed off to a separate review track. It is never directly submittable, so the first gate cannot be bypassed.
4. The review track is voted on by the deduplicated union of the `Economic` and `Building` collectives, who can fast-track, delay, or cancel enactment.
5. If the call survives review, the scheduler dispatches it as root through an atomic enact wrapper.

## New Pallets

- **`pallet-multi-collective`** — generic membership storage for several named collectives under one `CollectiveId`. Root-gated add/remove/swap/set, plus periodic term hooks for rotating collectives.
- **`pallet-referenda`** — generic track-based referendum state machine. Supports `PassOrFail` (approve/reject/delegate) and `Adjustable` (timing decision over a scheduled call, with fast-track and cancel thresholds). Owns proposal queues, per-proposer quotas, terminal statuses, and atomic enact wrapping so dispatch and the `Enacted` status transition happen together.
[...]

PR #2620, opened 2026-04-27
Governance refactoring (V1 -> V2)

shamil-gadelshin:
# Migrate governance to a modular V2 stack

## Summary

Replaces the V1 monolith (`pallet-governance`) with a four-pallet modular stack as specified in `DESIGN.md` and motivated by `pallets/governance/README.md`. V1 is unwired from the runtime; its source remains in-tree as a reference and is unaffected by this PR.

Anonymous voting is **out of scope** for this PR — `pallet-anonymous-voting` is wired but stubbed; the bLSAG flow is not exercised.

Compared to V1, the branch moves governance from one pallet with tightly-coupled proposer, triumvirate, scheduling, collective-delay, and rotation logic into reusable primitives with runtime wiring.

### Generic collective management via `pallet-multi-collective`

A reusable collective pallet that provides:

- runtime-defined collectives with `min_members`, `max_members`, and optional `term_duration`
- membership operations: `add_member`, `remove_member`, `swap_member`, `reset_members`
- `OnMembersChanged` hook for downstream cleanup
- `OnNewTerm` hook fired per `term_duration` in `on_initialize`, available for periodic rotation / election wiring
- `CollectiveInspect` trait for live membership queries (`is_member`, `members_of`, `member_count`)
- `integrity_test` asserting `min_members ≤ max_members ≤ T::MaxMembers` and rejecting `term_duration: Some(0)` at runtime construction

This is the structural replacement for the hard-coded V1 proposer / triumvirate / collective membership state that used to live inside `pallet-governance`.

### Generic referenda lifecycle via `pallet-referenda`

A track-based governance pallet with:

- `Proposal::Action(call)` for approval referenda
- `Proposal::Review(task_name)` for collective timing review (references an externally-scheduled named task)
- `DecisionStrategy::PassOrFail` for triumvirate approval
- `DecisionStrategy::Adjustable` for collective review / delay adjustment
- runtime-provided `TracksInfo`
- `submit`, `cancel`, and scheduler-driven `finalize_referendum` extrinsics
- A state machine in `on_tally_updated` that fast-tracks above `fast_track_threshold`, rejects above `reject_threshold`, and otherwise interpolates the delay
- A reaper alarm for `Adjustable` polls that closes the storage leak which would otherwise occur if no votes arrive on a Review and the upgrade executes naturally
- `TracksInfo::authorize_proposal(track, &call)` invoked for every `Action` proposal at submit (default trait impl returns `true`; runtime hook is a no-op until a per-track policy is defined)
- A granular event surface: `Submitted`, `Approved`, `Rejected`, `Cancelled`, `Expired`, `FastTracked` (distinct from `Approved` — fired on `Adjustable` fast-track), `TaskRescheduled` (every vote-driven reschedule, including fast-track), `TaskCancelled` (successful cancel of a Review's underlying task on reject), `SchedulerOperationFailed`
- `Polls` integration so voting pallets can push tally updates into referenda

This is the V2 replacement for V1's `propose`, `vote_on_proposed`, timeout cleanup, and scheduled delay-period lifecycle.

### Decoupled signed voting via `pallet-signed-voting`

A dedicated signed-voting pallet that provides:

- signed `vote(approve)` / `remove_vote` extrinsics over a `Polls` provider
- per-poll tally storage with `SignedVoteTally { ayes, nays, total }`
- `PollHooks::on_poll_created` / `on_poll_completed`
- `From<SignedVoteTally>` conversion to a generic `VoteTally` (Perbill triple)
- `remove_votes_for(who)` helper for V1 README §84 — wired via the runtime's `GovernanceVoteCleanup`

This extracts the signed-voting responsibilities that were previously embedded inside V1 triumvirate / collective vote storage.

### Runtime wiring

The runtime now wires:

- a `Proposers` collective (max 20, indefinite term — `README.md` §"Allowed Proposers")
- a `Triumvirate` collective (max 3, indefinite term — `README.md` §"Triumvirate")
- `Economic` and `Building` collectives (max 16 each, 60-day term — `README.md` §"Economic and Building Collectives")
- track 0 "triumvirate" — `PassOrFail`, decision period 7 days (`README.md` §"Proposal Submission"), 2/3 approve / 2/3 reject (`README.md` §"Triumvirate Approval"), proposer set `Proposers`, voter set `Triumvirate`
- track 1 "review" — `Adjustable`, initial delay 1 hour (`README.md` §"Delay Period"), `fast_track_threshold = 67%` (matches V1's `FastTrackThreshold`), `reject_threshold = 51%`, voter set = `Union(Economic, Building)`
- signed voting as the active voting implementation for both tracks
- vote cleanup on membership removal via `GovernanceVoteCleanup`, which routes `OnMembersChanged` outgoing events into `SignedVoting::remove_votes_for`
- `pallet-governance` removed from `construct_runtime!`; its slot is reused by `MultiCollective`, `SignedVoting`, and `Referenda`

This mirrors the V1 actors described in `README.md`: allowed proposers submit privileged actions, triumvirate approves / rejects, economic + building collectives review timing after approval.

### Trait scaffolding

`SetLike<T>`, `Polls<AccountId>`, and `PollHooks<PollIndex>` traits added so voting and referenda pallets compose without coupling. Tuple impl `(A, B): PollHooks` lets the runtime fan-out hooks across multiple voting pallets. `VoteTally` is a Perbill triple boundary type; voting pallets compute internally however they wish and only emit `(approval, rejection, abstention)`.

## What was ported from V1 and the docs

Ignoring anonymous voting, this branch implements a substantial portion of the V1 and `DESIGN.md` model:

- Governance is split into reusable primitives instead of one monolithic pallet.
- Proposer eligibility is runtime-defined through a collective-backed proposer set.
- Triumvirate approval is modeled as a dedicated approval track with 2/3 signed thresholds.
- Collective oversight is modeled as a separate review track with fast-track / reject / delay behavior.
- Scheduler-backed execution is preserved.
- Voting tallies are pushed into referenda through a generic `Polls` interface.
- Membership changes are routed into signed-voting cleanup through `OnMembersChanged`, matching the V1 README behavior even though `DESIGN.md` says the opposite.
- `multi-collective` exposes the `OnNewTerm` hook expected by `DESIGN.md` for future direct rotation / election behavior.
- `referenda::submit` enforces `TracksInfo::authorize_proposal` for `Action` proposals.
- The "review poll can leak forever if nobody votes" issue is mitigated with a reaper alarm.

### Two-phase approval → review handoff does not work as written

`DESIGN.md`'s "Worked Example" prescribes that track 0 approval execute `batch_all(schedule_named(...), referenda.submit(track: 1, Review(...)))`. After triumvirate approval, the batch is dispatched as Root, but `referenda::submit` is signed-only, so the inner track-1 submit fails with `BadOrigin` and `batch_all` rolls back. `DESIGN.md` flagged this as Open Issue #4 to verify; the verification fails. Track 1 can be reached only via a separate proposer-signed submit, decoupled from triumvirate approval. (This can be mitigated using Substrate's `dispatch_as` pattern.)

### Adjustable delay anchor differs from V1

The Adjustable delay is anchored on the Review's `submitted` block, not on the proposer's scheduled execution block. V1 anchored at the captured `dispatch_time`, so collective votes shifted execution relative to the proposer's chosen time. V2 has no notion of "the proposer's scheduled time" because the scheduled task is created externally by the proposer's batch and `submit(Review)` only stores the task name; the first vote can move execution earlier or later than the proposer wanted depending on `initial_delay`.

### No genesis seeding

None of the new pallets has a `pallet::genesis_config`. V1 had `pallet_governance::GenesisConfig { allowed_proposers, ... }`. The chain launches with all four collectives empty and is stuck until sudo intervention.

### Collective invariants are weaker than V1

The runtime currently sets `min_members: 0` for all governance collectives. V1 README §"Triumvirate" requires the Triumvirate to maintain 3 members; with `min_members = 0`, a 1-member triumvirate clears the 67% threshold on a single aye. Economic / Building collectives have similarly soft floors.

### Stakeholder rotation is unimplemented

`OnNewTerm` is wired structurally but the runtime sets it to `()`. `README.md` requires top-16 validators by stake and top-16 subnet owners refreshed every 6 months; `DESIGN.md` describes the Election mechanism that should live behind this hook. Term timer fires; nothing happens.

### Proposer self-cancel and self-eject are missing

V1 README promises proposers can cancel or withdraw their own proposals and eject their own keys from the allowed-proposers list — V1 implementation never delivered them either, and V2 inherits the gap. Cancellation and membership removal are root-only.

### Adjustable review semantics differ from `DESIGN.md`

`DESIGN.md` says adjustable review polls have no timeout and live until task execution or cancellation. V2 adds a synthetic reaper alarm that closes the no-vote storage leak but can also expire a still-live review poll under unusual scheduler conditions.

### Cancelling a Review does not cancel the reviewed task

`cancel()` removes the reaper alarm but leaves the underlying reviewed task scheduled. A separate scheduler-side cancel is required.

### Scheduler failures can finalize referenda into states that do not reflect reality

Approval, fast-track, and rejection paths log scheduler errors and emit `SchedulerOperationFailed`, but they still conclude the poll into the corresponding terminal status. Observers see `Approved` or `Rejected` even when the underlying scheduler call did not take effect.

### Review referenda are not actually coupled to triumvirate approval

Because the design's automatic handoff is broken (see above), track 1 can currently be submitted separately by any proposer and can reference any existing named scheduler task. The two-phase coupling described by `DESIGN.md` is convention, not enforcement.

### Overlapping Economic / Building members are double-counted in the review denominator

`Union::len()` sums both collectives, while actual vote eligibility is only boolean membership. A dual member counts twice in `total` but can vote once.

### Linear vs V1 exponential delay

V2 uses approval-only linear interpolation. V1 README specifies `initial_delay × delay_factor^net_score` where `net_score = nays - ayes` and 17 nays cancel; the graduated-pressure mechanism is gone in V2. This is closer to `DESIGN.md` than V1, but it is not V1-equivalent.

### Review fast-track threshold differs from the guide's example

`DESIGN.md` uses 75% in the worked example; runtime currently configures 67% to match V1's `FastTrackThreshold`. Worth flagging that `DESIGN.md` and V1 disagree on this constant.

### Known signed-voting issues still present

A rotated-out voter still cannot remove their own vote — `remove_vote` checks current membership and rejects post-rotation calls. And `ActivePolls` silently drops polls once full, which also causes `remove_votes_for` cleanup to miss those dropped polls. Both are documented by ignored regression tests in the signed-voting test suite. Today the global `MaxActivePolls` and `MaxQueued` are equal, so the silent drop is unreachable; the invariant is conventional, not enforced.

## Next steps

A practical way to mitigate the design discrepancies above is to introduce a new `pallet-governance-management` pallet that sits above the V2 primitives and provides:

- referenda helpers — replacing the brittle `batch_all` / `dispatch_as` flows from the design's worked example with a pallet-internal call that creates the track-1 review poll directly when track 0 approves
- sudo referenda cancellation — a single-poll path and a "cancel all + storage cleanup" path for emergency rollback
- collective management and member rotation — the wiring needed for `OnNewTerm` to reach top-N stakeholder selection from `pallet-subtensor`, plus the proposer self-eject / self-cancel paths the V1 README promised

This keeps the V2 primitives generic and pushes governance-policy concerns into a separate, replaceable pallet.


runtime/src/lib.rs
```diff
@@ -1639,60 +1637,304 @@ impl pallet_contracts::Config for Runtime {
...
+    /// Matches `ReferendaMaxQueued` — signed-voting tracks every ongoing poll
+    /// it sees, so the bound must cover all active referenda.
+    pub const SignedVotingMaxActivePolls: u32 = 20;
+    pub const GovernanceSignedScheme: GovernanceVotingScheme = GovernanceVotingScheme::Signed;
+    /// 60 days mainnet / 100 blocks fast-runtime.
+    pub const GovernanceCollectiveTermDuration: BlockNumber = prod_or_fast!(432_000, 100);
+    /// 7 days mainnet / 50 blocks fast-runtime — triumvirate voting window.
+    pub const GovernanceTriumvirateDecisionPeriod: BlockNumber = prod_or_fast!(50_400, 50);
+    /// 1 hour mainnet / 30 blocks fast-runtime — collective Review delay.
+    pub const GovernanceCollectiveInitialDelay: BlockNumber = prod_or_fast!(300, 30);
```
evgeny-s, line 1764: I think it makes sense to increase this delay up to 24 hours; otherwise, if the proposal passes the track 0, it might be not enough time to vote for building/economics.

runtime/src/lib.rs
```diff
@@ -1639,60 +1637,304 @@ impl pallet_contracts::Config for Runtime {
...
+                    name: name(b"review"),
+                    proposer_set: GovernanceMemberSet::Single(GovernanceCollectiveId::Proposers),
+                    voter_set: GovernanceMemberSet::Union(alloc::vec![
+                        GovernanceCollectiveId::Economic,
+                        GovernanceCollectiveId::Building,
+                    ]),
+                    voting_scheme: GovernanceVotingScheme::Signed,
+                    decision_strategy: DecisionStrategy::Adjustable {
+                        initial_delay: GovernanceCollectiveInitialDelay::get(),
+                        fast_track_threshold: Perbill::from_percent(67),
```
evgeny-s: We should have 75% for fast track here, based on the documentation.

runtime/src/lib.rs
```diff
@@ -1639,60 +1637,304 @@ impl pallet_contracts::Config for Runtime {
...
+                        approve_threshold: Perbill::from_rational(2u32, 3u32),
+                        reject_threshold: Perbill::from_rational(2u32, 3u32),
+                    },
+                },
+            },
+            RefTrack {
+                id: 1u8,
+                info: RefTrackInfo {
+                    name: name(b"review"),
+                    proposer_set: GovernanceMemberSet::Single(GovernanceCollectiveId::Proposers),
```
evgeny-s: Why do we need a proposer set here? Track 1 should not have the possibility to bypass triumverate voting from track 0.

pallets/referenda/src/lib.rs
```diff
@@ -79,146 +220,564 @@ pub mod pallet {
...
+    fn do_approve(index: ReferendumIndex, info: &ReferendumInfoOf<T>) {
+        if let Some((_when, ref address)) = info.scheduled_task
+            && let Err(err) = T::Scheduler::cancel(address.clone())
+        {
+            Self::handle_scheduler_error(index, "cancel", err);
+        }
 
-    fn track_ids() -> impl Iterator<Item = Self::Id> {
-        Self::tracks().map(|x| x.id)
+        if let Proposal::Action(ref bounded_call) = info.proposal
```
evgeny-s: Track 0 and Track 1 are not connected. I think we should schedule here the next proposal for track 1 with the Adjustable strategy.

pallets/referenda/src/lib.rs
```diff
@@ -41,10 +55,137 @@ pub type VotingSchemeOf<T> = <TracksOf<T> as TracksInfo<
...
+        approve_threshold: Perbill,
+        reject_threshold: Perbill,
+    },
+    /// Timing adjustment for an already-scheduled task.
+    Adjustable {
+        initial_delay: BlockNumber,
+        fast_track_threshold: Perbill,
+        reject_threshold: Perbill,
+    },
+}
```
lorismoulin: An enhanced `DecisionStrategy` could look like this:

```
pub enum DecisionStrategy<TrackId, Moment> {
    PassOrFail {
        decision_period: Moment,
        approve_threshold: Perbill,
        reject_threshold: Perbill,
        on_approval: ApprovalAction<TrackId>,
    },
    Adjustable {
        initial_delay: Moment,
        fast_track_threshold: Perbill,
        cancel_threshold: Perbill,
    },
}
```

The `on_approval` on the `PassOrFail` allows to have two-step proposal we want with triumvirate first approving/rejecting the proposal then the proposal getting reviewed on some other track. The `ApprovalAction` shape could be defined as below where `Execute` is next block execution from the approval block (avoid synchronous call else `on_tally_updated` will have unstable weight). 

```
pub enum ApprovalAction<TrackId> {
    Execute,
    ScheduleAndReview { review_track: TrackId },
}
```

The distinction between `reject_threshold` and `cancel_threshold` is made in the name because they have fundamentally different meaning. A `PassOrFail` get `rejected` before being scheduled while the `Adjustable` get cancelled while being scheduled.

pallets/referenda/src/lib.rs
```diff
@@ -79,146 +220,564 @@ pub mod pallet {
...
     impl<T: Config> Pallet<T> {
+        /// Submit a new referendum.
         #[pallet::call_index(0)]
         pub fn submit(
-            _origin: OriginFor<T>,
-            _track: TrackIdOf<T>,
-            _proposal: (),
+            origin: OriginFor<T>,
+            track: TrackIdOf<T>,
+            proposal: Proposal<BoundedCallOf<T>>,
```
lorismoulin: It will cause issue to have the `proposal` being of the `Proposal` type because it means there can be an external scheduling of the proposal and we submit it. We want the scheduling to be self-contained.

```suggestion
            call: Box<<T as Config>::RuntimeCall>,
```

The proposal type is implicit because bounded to the track type and branching on `track.decision_strategy`, we get something like:

```
let proposal = match track.decision_strategy {
    DecisionStrategy::PassOrFail { .. } => {
        // ...
        Proposal::Action(bounded_call)
    }
    DecisionStrategy::Adjustable { initial_delay, .. } => {
        // ...
        Proposal::Review
    };
}
```

pallets/referenda/src/lib.rs
```diff
@@ -79,146 +220,564 @@ pub mod pallet {
...
+            let submitter = ensure_signed(origin)?;
+
+            // 1. Validate track
+            let track_info = T::Tracks::info(track).ok_or(Error::<T>::BadTrack)?;
+
+            // 2. Validate proposal-strategy compatibility
+            ensure!(
+                Self::is_valid_configuration(&proposal, &track_info.decision_strategy),
+                Error::<T>::InvalidConfiguration
+            );
```
lorismoulin: Should probably be replaced by `T::Tracks::authorize_proposal`, this is an external concern and should be abstracted away, implemented in runtime wiring.

pallets/referenda/src/lib.rs
```diff
@@ -79,146 +220,564 @@ pub mod pallet {
...
+        }
+
+        if let Err(err) = T::Scheduler::reschedule_named(*task_name, DispatchTime::At(target)) {
+            Self::handle_scheduler_error(index, "reschedule_named", err);
+            return;
+        }
+
+        Self::deposit_event(Event::<T>::TaskRescheduled { index, at: target });
+    }
 }
```
lorismoulin: This will be executed for every vote with worst case weight with no easy way to refund, including all hooks. Ideally we want a fixed weight for the `on_tally_updated` that will advance the referendum by scheduling block + 1 task every time, similarly to how it is done in the substrate referenda pallet.

Showing code may give a better idea:
https://github.com/opentensor/subtensor/blob/4516472ad84b32c85cdb84d2f0375a5916e9e4f6/pallets/referenda/src/lib.rs#L360

pallets/referenda/src/lib.rs
```diff
@@ -79,146 +220,564 @@ pub mod pallet {
...
+            //     a call payload; `Review` proposals reference an external
+            //     task and have no payload to authorize.
+            if let Proposal::Action(bounded_call) = &proposal {
+                let (call, _) = T::Preimages::peek(bounded_call)
+                    .map_err(|_| Error::<T>::ProposalNotAuthorized)?;
+                ensure!(
+                    T::Tracks::authorize_proposal(track, &call),
+                    Error::<T>::ProposalNotAuthorized
+                );
+            }
```
lorismoulin: This can be collapsed if we follow above comment

pallets/referenda/src/lib.rs
```diff
@@ -79,146 +220,564 @@ pub mod pallet {
...
+                        .saturating_add(One::one()),
+                ),
+            };
+            let scheduled_task = if let Some(when) = alarm_when {
+                let call: CallOf<T> = Call::<T>::finalize_referendum { index }.into();
+                let bounded = T::Preimages::bound(call).map_err(|_| Error::<T>::SchedulerError)?;
+                let address = T::Scheduler::schedule(
+                    DispatchTime::At(when),
+                    None,
+                    128u8,
```
lorismoulin: 0 is higher priority, I think it may be better for governance.

pallets/multi-collective/src/lib.rs
```diff
@@ -104,15 +111,76 @@ pub mod pallet {
             let mut weight = Weight::zero();
 
             for collective in T::Collectives::collectives() {
-                if let Some(term_duration) = collective.info.term_duration {
-                    if n.checked_rem(&term_duration).unwrap_or(n).is_zero() {
-                        weight.saturating_accrue(T::OnNewTerm::on_new_term(collective.id));
-                    }
+                // Conservative upper bound for the iteration cost. Matches the
+                // storage-backed case; static `CollectivesInfo` impls pay a
+                // smaller CPU cost, so this is a safe overestimate.
```
lorismoulin, line 121: Ideally we could have the ::collectives() return `(Vec<CollectiveInfo>, Weight)`

runtime/src/lib.rs
```diff
@@ -1639,60 +1637,304 @@ impl pallet_contracts::Config for Runtime {
...
+                    term_duration: None,
+                },
+            },
+            McCollective {
+                id: GovernanceCollectiveId::Economic,
+                info: McCollectiveInfo {
+                    name: name(b"economic"),
+                    min_members: 0,
+                    max_members: Some(16),
+                    term_duration: Some(GovernanceCollectiveTermDuration::get()),
```
lorismoulin, line 1817: Can we just set this to `None` given there are no rotations yet?


```suggestion
                    term_duration: None,
```

Same for `Building` collective

runtime/src/lib.rs
```diff
@@ -1639,60 +1637,304 @@ impl pallet_contracts::Config for Runtime {
...
+                .zip(s.iter())
+                .for_each(|(dst, src)| *dst = *src);
+            out
+        }
+
+        [
+            McCollective {
+                id: GovernanceCollectiveId::Proposers,
+                info: McCollectiveInfo {
+                    name: name(b"proposers"),
```
lorismoulin, line 1796:
```suggestion
                    name: name(b"otf"),
```

May be better suited to make it more explicit who can propose

common/src/traits.rs
```diff
@@ -21,3 +24,19 @@ pub trait PollHooks<PollIndex> {
...
+impl<A: PollHooks<I>, B: PollHooks<I>, I: Copy> PollHooks<I> for (A, B) {
+    fn on_poll_created(poll_index: I) {
+        A::on_poll_created(poll_index);
+        B::on_poll_created(poll_index);
+    }
+    fn on_poll_completed(poll_index: I) {
+        A::on_poll_completed(poll_index);
+        B::on_poll_completed(poll_index);
+    }
+}
```
lorismoulin: You can use `impl_for_tuples`, it does implements for () too given there is no lower bound:

```suggestion
#[impl_trait_for_tuples::impl_for_tuples(10)]
impl<Index: Copy> PollHooks<I> for Tuple {
    fn on_poll_created(poll_index: Index) {
        for_tuples!( #( Tuple::on_poll_created(poll_index); )* );
    }

    fn on_poll_completed(poll_index: Index) {
        for_tuples!( #( Tuple::on_poll_completed(poll_index); )* );
    }
}
```

runtime/src/lib.rs
```diff
@@ -1731,7 +1973,11 @@ construct_runtime!(
         Swap: pallet_subtensor_swap = 28,
         Contracts: pallet_contracts = 29,
         MevShield: pallet_shield = 30,
-        Governance: pallet_governance = 31,
+
+        // Governance V2 (replaces pallet_governance which previously held index 31).
```
lorismoulin: We can safely remove the comment. Governance is not yet deployed.

runtime/src/lib.rs
```diff
@@ -1639,60 +1637,304 @@ impl pallet_contracts::Config for Runtime {
     type ApiVersion = ();
 }
 
+// ============================================================================
+// Governance V2: multi-collective + signed-voting + referenda
+// ============================================================================
+
```
lorismoulin, line 1644: Let's not call this governance V2 given there is no V1?

runtime/src/lib.rs
```diff
@@ -1639,60 +1637,304 @@ impl pallet_contracts::Config for Runtime {
...
+                },
+            },
+        ]
+        .into_iter()
     }
+}
+
+/// Static list of referenda tracks. Track 0 is the triumvirate approval track;
+/// track 1 is the collective oversight (Review) track.
+pub struct SubtensorTracks;
```
lorismoulin: I think we could move this to it's own file? Like governance.rs with all the static stuff, file is growing without limits. Polkadot is doing something similar.

runtime/Cargo.toml
```diff
@@ -156,7 +156,10 @@ stp-shield.workspace = true
 
 ethereum.workspace = true
 
-pallet-governance.workspace = true
+# Governance (V2)
```
lorismoulin:
```suggestion
# Governance
```

pallets/anonymous-voting/src/lib.rs
```diff
@@ -1,65 +1,32 @@
 #![cfg_attr(not(feature = "std"), no_std)]
```
lorismoulin, line 1: Can we just remove this file and also the primitives/crypto if we don't plan to use it in the initial release?

pallets/multi-collective/src/lib.rs
```diff
@@ -124,19 +192,24 @@ pub mod pallet {
...
+                    .map_err(|_| Error::<T>::TooManyMembers)?;
                 Ok(())
             })?;
 
-            T::OnMembersChanged::on_members_changed(collective_id, &[who.clone()], &[]);
+            T::OnMembersChanged::on_members_changed(
+                collective_id,
+                core::slice::from_ref(&who),
+                &[],
+            );
```
lorismoulin, line 217: It make me realize that we should probably have a `fn weight()` on each hook to be able to compute the weight correctly + refund if needed

pallets/multi-collective/src/lib.rs
```diff
@@ -171,21 +250,28 @@ pub mod pallet {
...
+            T::Collectives::info(collective_id).ok_or(Error::<T>::CollectiveNotFound)?;
 
-            Members::<T>::try_mutate(&collective_id, |members| -> DispatchResult {
-                let pos = members.iter().position(|m| m == &remove)
+            Members::<T>::try_mutate(collective_id, |members| -> DispatchResult {
+                let pos = members
+                    .iter()
+                    .position(|m| m == &remove)
                     .ok_or(Error::<T>::NotMember)?;
                 ensure!(!members.contains(&add), Error::<T>::AlreadyMember);
```
lorismoulin, line 265: At first I did chose to do the naive approach with a `contains` but complexity is O(n), maybe we could have a similar approach to collective pallet where they sort the members and use a binary search?

pallets/multi-collective/src/lib.rs
```diff
@@ -196,11 +282,13 @@ pub mod pallet {
             members: Vec<T::AccountId>,
         ) -> DispatchResult {
             T::ResetOrigin::ensure_origin(origin, &collective_id)?;
-            let info = T::Collectives::info(collective_id)
-                .ok_or(Error::<T>::CollectiveNotFound)?;
+            let info = T::Collectives::info(collective_id).ok_or(Error::<T>::CollectiveNotFound)?;
```
lorismoulin, line 290: Maybe calling this `set_members` is enough?

pallets/multi-collective/src/lib.rs
```diff
@@ -282,12 +375,22 @@ pub trait OnMembersChanged<CollectiveId, AccountId> {
     );
 }
 
+impl<CollectiveId, AccountId> OnMembersChanged<CollectiveId, AccountId> for () {
+    fn on_members_changed(_: CollectiveId, _: &[AccountId], _: &[AccountId]) {}
+}
```
lorismoulin, line 438: Could use `impl_for_tuples` here

pallets/multi-collective/src/lib.rs
```diff
@@ -282,12 +375,22 @@ pub trait OnMembersChanged<CollectiveId, AccountId> {
...
 pub trait OnNewTerm<CollectiveId> {
     /// A new term of a collective has started.
     fn on_new_term(collective_id: CollectiveId) -> Weight;
 }
 
+impl<CollectiveId> OnNewTerm<CollectiveId> for () {
+    fn on_new_term(_: CollectiveId) -> Weight {
+        Weight::zero()
+    }
+}
```
lorismoulin: Could use `impl_for_tuples` here

runtime/src/lib.rs
lorismoulin: @evgeny-s good catch, the proposer_set should probably be made optional and having a check in the submit to ensure track is accepting proposals

Cargo.toml
```diff
@@ -34,7 +34,7 @@ members = [
 	"support/*",
 	"chain-extensions",
 ]
-exclude = ["eco-tests"]
+exclude = ["eco-tests", "pallets/anonymous-voting", "pallets/governance"]
```
lorismoulin, line 37: Can we just remove both from the code?

runtime/src/governance/tracks.rs
```diff
@@ -0,0 +1,87 @@
...
+                        // body can fast-track or cancel before enactment.
+                        on_approval: ApprovalAction::Review { track: 1 },
+                    },
+                },
+            },
+            RefTrack {
+                id: 1u8,
+                info: RefTrackInfo {
+                    name: name(b"review"),
+                    proposer_set: Some(GovernanceMemberSet::Single(
```
evgeny-s, line 69: Can we use `None` here? Otherwise there is possibility to bypass the Triumvrate approva

runtime/src/governance/tracks.rs
```diff
@@ -0,0 +1,87 @@
...
+                        GovernanceCollectiveId::Proposers,
+                    )),
+                    voter_set: GovernanceMemberSet::Union(alloc::vec![
+                        GovernanceCollectiveId::Economic,
+                        GovernanceCollectiveId::Building,
+                    ]),
+                    voting_scheme: GovernanceVotingScheme::Signed,
+                    decision_strategy: DecisionStrategy::Adjustable {
+                        initial_delay: GovernanceCollectiveInitialDelay::get(),
+                        fast_track_threshold: Perbill::from_percent(67),
```
evgeny-s, line 79: I think we should have 75% here based on the design doc.

shamil-gadelshin:
The remaining issues will be addressed in the parent `governance` branch: https://github.com/opentensor/subtensor/pull/2198

Issue #2633, opened 2026-05-04
Configurable Tempo & Owner-Triggered Epochs — Specification


evgeny-s:
# Configurable Tempo & Owner-Triggered Epochs — Specification

**Status:** Draft / WIP

**Audience:** Subtensor pallet engineers, validator-tooling maintainers

**Goal:** Give subnet owners control over their epoch cadence, while keeping a network-wide guarantee that yield is distributed at least once per maximum tempo.

---

## 1. Motivation

The current design fixes every subnet to a single rigid cadence (tempo = 360 blocks ≈ 72 minutes) chosen by the protocol, not the builder. Builders whose subnet does not benefit from a 72-minute weight cycle (e.g. long-running training, infrequent inference benchmarks, asynchronous human-in-the-loop tasks) are forced to design around the cadence rather than have the cadence support their use case.

This feature deliberately moves authority from the protocol to the owner, with one network-level guarantee preserved: **stakers always receive yield at least once per `MaxTempo`**, regardless of owner behaviour.

---

## 2. Requirements

A subnet owner should be able to:

1. **R1** — Set the subnet's tempo to any value in `[MinTempo, MaxTempo]` blocks, rate-limited to `MinTempo` blocks (= 360), fixed regardless of current tempo (lets owners recover from a mistake within ~12 minutes instead of up to a week).
2. **R2** — Set the activity-cutoff factor for the subnet (in epochs of validator inactivity tolerated), rate-limited via the same `OwnerHyperparamUpdate` pattern.
3. **R3** — Manually trigger an epoch for their own subnet, rate-limited via the same `OwnerHyperparamUpdate` pattern. The triggered epoch becomes eligible to fire after `AdminFreezeWindow` blocks (giving validators their freeze window of warning before execution). Concurrent triggers from many owners are de-collided by the per-block epoch cap (§6.1) — at most `MaxEpochsPerBlock` epochs execute per block, the rest are deferred to the next block via `EpochDeferred`. Actual execution latency is therefore `AdminFreezeWindow + (queue position / MaxEpochsPerBlock)` blocks.

The protocol shall guarantee:

[...]

PR #2638, opened 2026-05-06
[Feature] Configurable Tempo & Owner-Triggered Epochs

evgeny-s:
## Description

Implementation of the [specification](https://github.com/opentensor/subtensor/issues/2633).

Demo (non-technical): https://dynamic-tempo-non-technical.netlify.app/
Demo (technical): https://dynamic-tempo-technical.netlify.app/#1

## Related Issue(s)

- Closes #[issue number]

## Type of Change

- [ ] Documentation update

## Breaking Change

Check the [11. Breaking changes & compatibility section](https://github.com/opentensor/subtensor/issues/2633).

## Checklist

## Additional Notes

## Simulations (miner rewards Events):
Trigger epoch:
image
Set tempo:
image


pallets/subtensor/src/coinbase/run_coinbase.rs
```diff
@@ -996,28 +1028,33 @@ impl<T: Config> Pallet<T> {
     /// # Returns
     /// * `bool` - True if the epoch should run, false otherwise.
     pub fn should_run_epoch(netuid: NetUid, current_block: u64) -> bool {
-        Self::blocks_until_next_epoch(netuid, Self::get_tempo(netuid), current_block) == 0
+        let tempo = Self::get_tempo(netuid);
+        if tempo == 0 {
+            return false;
```
gztensor, line 1069: This means that if tempo is set to 0, manual trigger or max cap of MAX_TEMPO will not work. Is this by design?

pallets/subtensor/src/coinbase/run_coinbase.rs
```diff
@@ -996,28 +1028,33 @@ impl<T: Config> Pallet<T> {
...
-    ///   2        1               0
-    ///   100      0              99
-    ///   100      1              98
-    /// Special case: tempo = 0, the network never runs.
-    ///
+    /// Returns the number of blocks remaining before the next automatic epoch under the
+    /// stateful scheduler (period `tempo + 1`, anchored on `LastEpochBlock`). Used by the
+    /// admin-freeze-window predicate and external tooling. Returns `u64::MAX` when
+    /// `tempo == 0` (legacy defensive short-circuit).
     pub fn blocks_until_next_epoch(netuid: NetUid, tempo: u16, block_number: u64) -> u64 {
```
gztensor: "Period is `tempo + 1`: next firing at `last + tempo + 1`." comment and the code below is not anymore correct.

pallets/subtensor/src/coinbase/tempo_control.rs
```diff
@@ -0,0 +1,102 @@
...
+
+        let tx = TransactionType::TempoUpdate;
+        ensure!(
+            tx.passes_rate_limit_on_subnet::<T>(&who, netuid),
+            Error::<T>::TxRateLimitExceeded
+        );
+
+        let now = Self::get_current_block_as_u64();
+
+        Tempo::<T>::insert(netuid, tempo);
```
gztensor: Should this be a function since Tempo and LastEpochBlock are always updated together?

gztensor (changes requested):

pallets/subtensor/src/coinbase/run_coinbase.rs
evgeny-s: Yes, the previous implementation meant that if tempo is 0, we don't run tempo on this subnet. 
We don't have this case in Mainnet, but we support it. There is still a possibility to set the tempo to 0 by the root. 
So we support the same semantics.

pallets/subtensor/src/coinbase/run_coinbase.rs
evgeny-s: Updated the function name and the comment to avoid the confusion: https://github.com/opentensor/subtensor/pull/2638/changes/ad7ba80dfd90a07636b6375cd664ec40eebb2a48

pallets/subtensor/src/coinbase/tempo_control.rs
evgeny-s: Added a helper function: https://github.com/opentensor/subtensor/pull/2638/changes/df184e33c7294b47e2c1f34581c57150953092f1

evgeny-s:
Hey, @basfroman , we have a bunch of tests that failed on the SDK side. 
The main reason is now we have a stateful calculation for the next epoch block, because the tempo is dynamic. 
So, [here](https://github.com/latent-to/bittensor/blob/master/bittensor/core/subtensor.py#L896) you can just call the new runtime API - `get_next_epoch_start_block`.

pallets/subtensor/src/utils/misc.rs
```diff
@@ -572,6 +590,27 @@ impl<T: Config> Pallet<T> {
         Self::deposit_event(Event::ActivityCutoffSet(netuid, activity_cutoff));
     }
 
+    /// Effective activity cutoff in blocks, derived from `ActivityCutoffFactorMilli` and `Tempo`.
+    /// `cutoff_blocks = (factor × tempo) / 1000`, clamped to ≥ 1.
+    pub fn get_activity_cutoff_blocks(netuid: NetUid) -> u64 {
```
open-junius, line 605: should we use it to replace get_activity_cutoff everywhere? for instance in metagraph.

evgeny-s: Yes, because ActivityCutoff is deprecated and will be removed in the follow-up release.
Who can help with updating it?

open-junius: It is fine to remove the storage later. Actually, my question is for https://github.com/opentensor/subtensor/blob/devnet-ready/pallets/subtensor/src/rpc_info/metagraph.rs#L713, if we should use get_activity_cutoff_blocks to replace get_activity_cutoff? It will change the return type, from u16 to u64.

evgeny-s: Good catch, for me to update:
- metagraph call
- breaking changes section in the description, regarding the metagraph

pallets/subtensor/src/macros/dispatches.rs
```diff
@@ -2585,7 +2585,7 @@ mod dispatches {
         /// When enabled, the caller's individual lock does not unlock through
         /// locked-mass decay. Passing `false` removes the flag, returning the
         /// caller's lock to normal decay.
-        #[pallet::call_index(138)]
+        #[pallet::call_index(139)]
```
lorismoulin: Why are we changing the call index here?

pallets/subtensor/src/macros/events.rs
```diff
@@ -612,6 +612,38 @@ mod events {
             netuid: NetUid,
         },
 
+        /// Activity-cutoff factor (per-mille) set on a subnet by its owner.
+        ActivityCutoffFactorMilliSet(NetUid, u32),
```
lorismoulin: We should prefer named fields over tuple for events, easier to access from client side. `event.netuid` is more meaningful than `event.0`.

pallets/subtensor/src/macros/events.rs
```diff
@@ -612,6 +612,38 @@ mod events {
...
+        },
+
+        /// `should_run_epoch` returned true but `is_epoch_input_state_consistent` returned false;
+        /// schedule advanced, epoch execution skipped.
+        EpochSkippedDueToInconsistentState {
+            /// The subnet identifier.
+            netuid: NetUid,
+            /// The block at which the slot was consumed.
+            block: u64,
+        },
```
lorismoulin: It would be more flexible to have `EpochSkipped`, more future-proof if we decide to add new skip reasons.

pallets/subtensor/src/macros/errors.rs
```diff
@@ -305,5 +305,11 @@ mod errors {
         CannotUseSystemAccount,
         /// Trying to unlock more than locked
         UnlockAmountTooHigh,
+        /// Tempo value out of `[MinTempo, MaxTempo]` bounds.
+        TempoOutOfBounds,
+        /// Activity-cutoff factor out of `[MinActivityCutoffFactorMilli, MaxActivityCutoffFactorMilli]` bounds.
+        ActivityCutoffFactorMilliOutOfBounds,
+        /// `trigger_epoch` called while a previously triggered epoch is still pending.
+        EpochTriggerAlreadyPending,
```
lorismoulin: Prefer plain English doc comments here rather than references to specific function names, since these comments are exposed through runtime metadata and may be shown client-side.

pallets/subtensor/src/macros/events.rs
```diff
@@ -612,6 +612,38 @@ mod events {
...
+            /// The subnet identifier.
+            netuid: NetUid,
+            /// Block at which the epoch was originally scheduled.
+            from_block: u64,
+            /// Block to which the epoch was deferred.
+            to_block: u64,
+        },
+
+        /// `should_run_epoch` returned true but `is_epoch_input_state_consistent` returned false;
+        /// schedule advanced, epoch execution skipped.
```
lorismoulin: Same comment as errors.

pallets/subtensor/src/coinbase/tempo_control.rs
```diff
@@ -0,0 +1,98 @@
...
+        Self::set_activity_cutoff_factor_milli(netuid, factor_milli);
+        tx.set_last_block_on_subnet::<T>(&who, netuid, now);
+
+        Ok(())
+    }
+
+    /// Owner-side `trigger_epoch` implementation.
+    /// Schedules the triggered epoch to fire after `AdminFreezeWindow` blocks; that
+    /// countdown engages the freeze window for the subnet via `is_in_admin_freeze_window`.
+    pub fn do_trigger_epoch(origin: OriginFor<T>, netuid: NetUid) -> Result<(), DispatchError> {
```
lorismoulin, line 76: Should we reject when `blocks_until_next_auto_epoch < AdminFreezeWindow`? In that case an automatic epoch is already scheduled before the manually triggered epoch can become eligible, so the pending trigger will be cleared by the automatic epoch. Rejecting seems less confusing for users than charging them and emitting `EpochTriggered` for a trigger that cannot actually fire.

pallets/subtensor/src/utils/misc.rs
lorismoulin: Could you please mark `ActivityCutoff` as `#[deprecated(note = "Some explanation")]`? It will propagate to metadata.

pallets/subtensor/src/lib.rs
```diff
@@ -1767,6 +1767,52 @@ pub mod pallet {
...
+    pub const MAX_TEMPO: u16 = 50_400;
+    /// Lower bound for activity-cutoff factor (per-mille). 1_000 = one full tempo.
+    pub const MIN_ACTIVITY_CUTOFF_FACTOR_MILLI: u32 = 1_000;
+    /// Upper bound for activity-cutoff factor (per-mille). 50_000 = 50 tempos.
+    pub const MAX_ACTIVITY_CUTOFF_FACTOR_MILLI: u32 = 50_000;
+    /// Default activity-cutoff factor (per-mille). 13_889 ≈ legacy 5000-block cutoff
+    /// at default tempo 360 (`13_889 * 360 / 1000 = 5_000`, exact via ceiling rounding).
+    pub const INITIAL_ACTIVITY_CUTOFF_FACTOR_MILLI: u32 = 13_889;
+    /// Per-block cap on number of epochs that may execute in a single `block_step`.
+    pub const MAX_EPOCHS_PER_BLOCK: u32 = prod_or_fast!(2, 32);
```
lorismoulin: Do we want this to be hardcoded here or maybe as config params?

pallets/subtensor/src/coinbase/tempo_control.rs
```diff
@@ -0,0 +1,98 @@
...
+        Ok(())
+    }
+
+    /// Owner-side `set_activity_cutoff_factor` implementation.
+    pub fn do_set_activity_cutoff_factor(
+        origin: OriginFor<T>,
+        netuid: NetUid,
+        factor_milli: u32,
+    ) -> DispatchResult {
+        let who = Self::ensure_subnet_owner(origin, netuid)?;
```
lorismoulin: Do we also want root to be able to change this?

pallets/subtensor/src/coinbase/reveal_commits.rs
```diff
@@ -38,8 +38,9 @@ impl<T: Config> Pallet<T> {
     /// The `reveal_crv3_commits` function is run at the very beginning of epoch `n`,
     pub fn reveal_crv3_commits_for_subnet(netuid: NetUid) -> dispatch::DispatchResult {
         let reveal_period = Self::get_reveal_period(netuid);
-        let cur_block = Self::get_current_block_as_u64();
-        let cur_epoch = Self::get_epoch_index(netuid, cur_block);
+        // If the subnet is deferred past this block the
+        // commits are taken once here and the later block(s) become no-ops.
+        let cur_epoch = Self::current_epoch_with_lookahead(netuid);
```
lorismoulin, line 43: Here we reveal commits before knowing whether the subnet epoch will actually execute in this block. If `MaxEpochsPerBlock` has already been reached, the epoch can be deferred, leaving the revealed weights public before being used. That seems to weaken the commit-reveal property and may allow actors to react to revealed weights through other epoch-relevant state. Can we either reveal only on the successful execution path, or document/test why this gap is not exploitable?

pallets/subtensor/src/coinbase/run_coinbase.rs
```diff
@@ -64,7 +64,14 @@ impl<T: Config> Pallet<T> {
...
         // --- 6. Distribute the emissions to the subnets.
+        //   Bonds masking inside `distribute_emission` reads `LastMechansimStepBlock` and
+        //   must see the previous successful run, so we delay the write until after.
         Self::distribute_emissions_to_subnets(&emissions_to_distribute);
+
+        // --- 7. Mark each successful epoch run as the last mechanism step.
+        for netuid in emissions_to_distribute.keys() {
+            LastMechansimStepBlock::<T>::insert(*netuid, current_block);
+        }
     }
```
lorismoulin: Could we move the `LastMechansimStepBlock` update into `distribute_emissions_to_subnets` after each successful `distribute_emission` call? Keeping the write outside the distribution function makes this sequencing easy to miss for future call sites.

pallets/subtensor/src/macros/dispatches.rs
evgeny-s: Merge conflict issue: [reverted](https://github.com/opentensor/subtensor/pull/2638/changes/6603db7c4daffef281d38a8264b62a587c7fe0b1)

pallets/subtensor/src/macros/events.rs
evgeny-s: We still have a batch of events, like this, examples: `RegistrationAllowed`, `PowRegistrationAllowed`. Anyway updated this one to named: https://github.com/opentensor/subtensor/pull/2638/changes/11cb8204365c1419c7f7ed36810dfc724d36bd8b

pallets/subtensor/src/macros/events.rs
evgeny-s: Makes sense: https://github.com/opentensor/subtensor/pull/2638/changes/ec34aa12dd76df762d79474f4298fd20ba0458a7

pallets/subtensor/src/macros/events.rs
evgeny-s: Fixed: https://github.com/opentensor/subtensor/pull/2638/changes/ec34aa12dd76df762d79474f4298fd20ba0458a7

pallets/subtensor/src/coinbase/tempo_control.rs
evgeny-s: Yes, agree. Updated: https://github.com/opentensor/subtensor/pull/2638/changes/61ef57f8df214dabceacc5496eb3cb516521bf0d

pallets/subtensor/src/utils/misc.rs
evgeny-s: Added: https://github.com/opentensor/subtensor/pull/2638/changes/be695df294fb01702944a8e2afeddc704d940fbf

evgeny-s: @l0r1s , BTW, it generates a lot of warnings with the deprecation notice in tests. But I planned to remove it in the follow-up PR. Are we ok to keep warnings?

pallets/subtensor/src/lib.rs
evgeny-s: Added as a pallet config param: https://github.com/opentensor/subtensor/pull/2638/changes/e9036db193e8b2296555c8233c6d7e7969f53853

evgeny-s: If we want to make it fully dynamic, as a follow-up PR - we can extract it to the storage item and add sudo extrinsics

pallets/subtensor/src/coinbase/tempo_control.rs
evgeny-s: Yes, added: https://github.com/opentensor/subtensor/pull/2638/changes/b528257e3516ff0f0eb598ab26f1db8816fa657c

pallets/subtensor/src/coinbase/reveal_commits.rs
evgeny-s: This is the edge case when we have > 2 epoch execution per block, and we we want to make it right, theoretically we could move `reveal_crv3_commits` to `drain_pending` after we 100% sure that it will be executed (`is_epoch_input_state_consistent` - true), but the thing is we have the test `test_reveal_crv3_commits_retry_on_missing_pulse` that checks the missing pulse and does retry, but in this case, we won't be able to retry, so I would keep it as is. What we can do is increase the `MaxEpochsPerBlock` to 3, which makes this edge case less achievable because it would take two subnets in collusion to manipulate this vulnerability.
WDYT?

pallets/subtensor/src/coinbase/run_coinbase.rs
evgeny-s: Yes, makes sense moved to distribute_emission: https://github.com/opentensor/subtensor/pull/2638/changes/9f33cd42d74f840089a186b7da4cc52ec819bd20

pallets/subtensor/src/macros/errors.rs
evgeny-s: Updated: https://github.com/opentensor/subtensor/pull/2638/changes/231f2bfee16f28f6a69a8ad56b5d400d7d0c9d02

pallets/subtensor/src/migrations/migrate_dynamic_tempo.rs
```diff
@@ -0,0 +1,163 @@
...
+        ActivityCutoffFactorMilli::<T>::insert(netuid, clamped);
+        activity_factor_seeded = activity_factor_seeded.saturating_add(1);
+        writes = writes.saturating_add(1);
+    }
+
+    // --- CR-v2: rewrite every `WeightCommits` entry to the
+    // `(hash, commit_epoch, commit_block, _)` layout. Field 1 was the absolute
+    // `commit_block`; it becomes `commit_epoch` (legacy modulo epoch). Field 2
+    // keeps the absolute `commit_block` (used by the epoch column-mask).
+    let crv2_entries: Vec<_> = WeightCommits::<T>::iter().collect();
```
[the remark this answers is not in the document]
evgeny-s: We have a few WeightCommits on mainnet, so we should be good.

pallets/subtensor/src/utils/misc.rs
evgeny-s: Updated Metagraph + for now commented the deprecation note, otherwise clippy is failing. We will remove it fully in the follow up PR: https://github.com/opentensor/subtensor/pull/2638/changes/a8fee8241e59308c45c30f66e21e20289042eb82

open-junius:
need to fix clippy and unit test.

pallets/subtensor/src/coinbase/reveal_commits.rs
JohnReedV: In the case that we must defer the epoch can't we defer the reveal too?

JohnReedV (approved):

sam0x17 (approved):

precompiles/src/mock.rs
```diff
@@ -116,6 +116,10 @@ parameter_types! {
     pub const InitialMaxBurn: TaoBalance = TaoBalance::new(1_000_000_000);
     pub const MinBurnUpperBound: TaoBalance = TaoBalance::new(1_000_000_000);
     pub const MaxBurnLowerBound: TaoBalance = TaoBalance::new(100_000_000);
+    pub const MinTempo: u16 = pallet_subtensor::MIN_TEMPO;
+    pub const MaxTempo: u16 = pallet_subtensor::MAX_TEMPO;
+    pub const MinActivityCutoffFactorMilli: u32 = pallet_subtensor::MIN_ACTIVITY_CUTOFF_FACTOR_MILLI;
+    pub const MaxActivityCutoffFactorMilli: u32 = pallet_subtensor::MAX_ACTIVITY_CUTOFF_FACTOR_MILLI;
```
lorismoulin, line 122: Maybe a bit risky if we inadvertently change the real runtime settings just because we wanted to change the mock?

pallets/subtensor/src/lib.rs
```diff
@@ -1952,6 +1996,7 @@ pub mod pallet {
         StorageMap<_, Identity, NetUid, u16, ValueQuery, DefaultImmunityPeriod<T>>;
 
     /// --- MAP ( netuid ) --> activity_cutoff
+    // #[deprecated(note = "Replaced by `ActivityCutoffFactorMilli` (per-mille of `Tempo`).")]
```
lorismoulin, line 1999: Should this be removed?

pallets/subtensor/src/subnets/subnet.rs
```diff
@@ -298,6 +298,12 @@ impl<T: Config> Pallet<T> {
         // --- 3. Fill tempo memory item.
         Tempo::<T>::insert(netuid, tempo);
 
+        // --- 3.1. Initialise `LastEpochBlock` with a per-netuid stagger
```
lorismoulin, line 301: Nit: this is more of a best-effort phase offset than a guaranteed stagger, since `now` participates in the anchor and different registration blocks can still collide. Maybe we could note that in the comment?

pallets/admin-utils/src/lib.rs
```diff
@@ -611,6 +611,9 @@ pub mod pallet {
         /// The extrinsic sets the activity cutoff for a subnet.
         /// It is only callable by the root account or subnet owner.
         /// The extrinsic will call the Subtensor pallet to set the activity cutoff.
+        // #[deprecated(
+        //     note = "Please use set_activity_cutoff_factor instead. This extrinsic will be removed soon."
+        // )]
```
lorismoulin, line 616: Maybe it could be removed?

pallets/subtensor/src/coinbase/run_coinbase.rs
```diff
@@ -1010,28 +1064,57 @@ impl<T: Config> Pallet<T> {
...
+        let last = LastEpochBlock::<T>::get(netuid);
+        let auto_next = last.saturating_add(tempo as u64);
+
+        let pending = PendingEpochAt::<T>::get(netuid);
+        if pending > 0 {
+            Some(auto_next.min(pending))
+        } else {
+            Some(auto_next)
+        }
     }
```
lorismoulin, line 1119: Nit: `get_next_epoch_start_block` feels slightly misleading now that epochs are modeled as scheduled execution slots rather than ranges with a start/end. Would `get_next_epoch_block`, `get_next_epoch_trigger_block`, or `get_next_epoch_fire_block` be clearer?

lorismoulin (commented):
Looks good, only small nits 👍

precompiles/src/mock.rs
evgeny-s: Makes sense, will add a hardcoded values here.

pallets/subtensor/src/lib.rs
evgeny-s: This map will be completely removed in the follow-up PR.

pallets/subtensor/src/subnets/subnet.rs
evgeny-s: Will add

pallets/admin-utils/src/lib.rs
evgeny-s: Yes, it will be removed

pallets/subtensor/src/coinbase/run_coinbase.rs
evgeny-s: Yes, maybe this one: `get_next_epoch_trigger_block`. Will update it.

PR #2645, opened 2026-05-07
Subnet Protocol Alpha Accounting

JohnReedV:
## Description
This PR adds protocol-owned alpha accounting for subnet chain buys and includes that alpha in subnet deregistration settlement.

- Stops recycling alpha bought during protocol TAO buys
- Adds per-subnet cached protocol alpha accounting
- Includes protocol alpha-in and cached protocol alpha in deregistration pro-rata settlement
- Clears cached protocol alpha when a subnet is dissolved
- Updates tests for the new deregistration settlement behavior


pallets/subtensor/src/coinbase/run_coinbase.rs
```diff
@@ -102,7 +102,9 @@ impl<T: Config> Pallet<T> {
                             if let Ok(buy_swap_result_ok) = buy_swap_result {
                                 let bought_alpha: AlphaBalance =
                                     buy_swap_result_ok.amount_paid_out.into();
-                                Self::recycle_subnet_alpha(*netuid_i, bought_alpha);
```
gztensor, line 110: Could you please call `burn_alpha` if no changes in total alpha issuance are required? We need to account for all kinds of alpha disappearance

gztensor (changes requested):

pallets/subtensor/src/coinbase/run_coinbase.rs
JohnReedV: It gets recycled later after being added to the protocol account on subnet dereg

gztensor (dismissed):

sam0x17 (approved):

pallets/subtensor/src/coinbase/run_coinbase.rs
camfairchild: We should keep it in an account then, no? cc: @gztensor

PR #2657, opened 2026-05-12
Subnet Emission Owner Toggle

JohnReedV:
## Description
This PR adds an owner/root admin-utils call to disable pool-side subnet emission.

- Adds a `SubnetEmissionEnabled` toggle
- Disables `alpha_in`, `tao_in`, and chain buys when toggled off
- Redistributes disabled subnet TAO emission to enabled subnets to preserve full block emission
- Preserves `alpha_out`, root prop, and normal validator/server emissions
- Resets the toggle for newly initialized subnets
- Adds owner/root rate-limit tracking for the new setting
- Adds tests for redistribution and multi-subnet toggle behavior


lorismoulin (changes requested):
Missing tests

pallets/admin-utils/src/lib.rs
```diff
@@ -2141,6 +2149,51 @@ pub mod pallet {
...
+
+        /// Enables or disables subnet pool-side emission for a subnet.
+        ///
+        /// This does not remove the subnet from emission share calculation and does not
+        /// change `alpha_out`, owner cut, root proportion, pending server emission, or
+        /// pending validator emission. It only zeros the pool-side `alpha_in`, `tao_in`,
+        /// and `excess_tao` chain-buy paths.
+        #[pallet::call_index(92)]
+        #[pallet::weight((
+            Weight::from_parts(25_000_000, 0)
```
evgeny-s: Would be great to add a benchmark for this extrinsic as well as the weight

pallets/subtensor/src/coinbase/subnet_emissions.rs
```diff
@@ -22,13 +22,33 @@ impl<T: Config> Pallet<T> {
         subnets_to_emit_to: &[NetUid],
         block_emission: U96F32,
     ) -> BTreeMap<NetUid, U96F32> {
-        // Get subnet TAO emissions.
+        // Get subnet TAO emission shares for all emission-eligible subnets first. This keeps
+        // disabled subnets in the returned map so they still continue through alpha_out/root-prop
+        // accounting, while their TAO-side emission is redistributed to enabled subnets.
```
lorismoulin: I'm not sure to understand that part, we keep them for alpha_out/root-prop accounting even though they receive no emissions? Doesn't this creates a discrepancy between what is actually done and what is recorded?

pallets/subtensor/src/coinbase/subnet_emissions.rs
```diff
@@ -22,13 +22,33 @@ impl<T: Config> Pallet<T> {
...
         let shares = Self::get_shares(subnets_to_emit_to);
         log::debug!("Subnet emission shares = {shares:?}");
 
+        let zero = U64F64::saturating_from_num(0.0);
+        let has_disabled_subnets = shares
+            .keys()
+            .any(|netuid| !SubnetEmissionEnabled::<T>::get(*netuid));
```
evgeny-s: We might want to collect all the flags with netuid to BTreeMap, to avoid reading it 3 times here.

pallets/subtensor/src/coinbase/subnet_emissions.rs
```diff
@@ -22,13 +22,33 @@ impl<T: Config> Pallet<T> {
...
+        let enabled_share_sum: U64F64 = shares
+            .iter()
+            .filter(|(netuid, _)| SubnetEmissionEnabled::<T>::get(**netuid))
+            .fold(zero, |acc, (_, share)| acc.saturating_add(*share));
```
lorismoulin: We are reading `SubnetEmissionEnabled` for a specific `netuid` 3 times in this hot path. You can use `.partition` here to get the enabled/disabled array first and build `BTreeMap` from them.

pallets/subtensor/src/coinbase/subnet_emissions.rs
JohnReedV: You're right. The comment is wrong.

lorismoulin (dismissed):

evgeny-s (dismissed):

lorismoulin (dismissed):

sam0x17 (approved):

PR #2679, opened 2026-05-21
Support real-pays-fee through three proxy levels

edpoptart:
## Description

Extends `ChargeTransactionPaymentWrapper` fee-payer resolution for `RealPaysFee` proxy calls to support a bounded proxy chain of up to three proxy levels.

This also supports homogeneous `batch`, `batch_all`, and `force_batch` calls at the innermost supported level, while preserving the existing fallback where batches containing mixed real accounts are charged to the original signer.

The transaction priority behavior remains unchanged.

## Related Issue(s)

- N/A

## Type of Change

- [x] Documentation update

## Breaking Change

N/A

## Checklist

N/A

## Additional Notes

Validation performed:

- `./scripts/fix_rust.sh` completed `cargo check`, `cargo clippy --fix`, `cargo fix`, and `cargo fmt`
- `cargo test -p node-subtensor-runtime --test transaction_payment_wrapper` passed with 25 tests.
- `git diff --check` passed.

The failing checks appear unrelated to this change:

- `cargo audit` is failing on existing yanked dependencies (`core2 0.4.0`, `keccak 0.1.5`).
- `Try Runtime / check testnet` fails in `migrate_fix_total_issuance_evm_fees` with `TotalIssuance diff greater than allowable delta`; `finney` and `devnet` passed.
- `Validate-Benchmarks` generated the standard benchmark patch artifact and suggests applying the `apply-benchmark-patch` label.

This PR only changes `ChargeTransactionPaymentWrapper` fee-payer resolution/tests/docs and does not touch dependencies, migrations, or pallet benchmarked extrinsics.

### DIFF
--- a/runtime/src/transaction_payment_wrapper.rs
+++ b/runtime/src/transaction_payment_wrapper.rs
@@ -28,6 +28,8 @@ type RuntimeOriginOf<T> = <T as frame_system::Config>::RuntimeOrigin;
 type AccountIdOf<T> = <T as frame_system::Config>::AccountId;
 type LookupOf<T> = <T as frame_system::Config>::Lookup;
 
+const MAX_REAL_PAYS_FEE_PROXY_DEPTH: u8 = 3;
+
 #[freeze_struct("f003cde1f9da4a90")]
 #[derive(Encode, Decode, DecodeWithMemTracking, Clone, Eq, PartialEq, TypeInfo)]
 #[scale_info(skip_type_params(T))]
@@ -85,13 +87,15 @@ where
 
     /// Determine who should pay the transaction fee for a proxy call.
     ///
-    /// Follows the RealPaysFee chain up to 2 levels deep:
+    /// Follows the RealPaysFee chain up to three proxy levels deep:
     /// - Case 1: `proxy(real=A, call)` → A pays if `RealPaysFee<A, signer>`
     /// - Case 2: `proxy(real=B, proxy(real=A, call))` → A pays if both
     ///   `RealPaysFee<B, signer>` and `RealPaysFee<A, B>` are set; B pays if only the former.
     /// - Case 3: `proxy(real=B, batch([proxy(real=A, ..), ..]))` → A pays if
     ///   `RealPaysFee<B, signer>`, all batch items are proxy calls with the same real A,
     ///   and `RealPaysFee<A, B>` is set; B pays if only the first condition holds.
+    /// - Case 4: `proxy(real=C, proxy(real=B, batch([proxy(real=A, ..), ..])))`
+    ///   → A pays if all three `RealPaysFee` relationships are set and the batch is homogeneous.
     ///
     /// Returns `None` if the signer should pay (no RealPaysFee opt-in).
     fn extract_real_fee_payer(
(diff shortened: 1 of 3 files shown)


runtime/src/transaction_payment_wrapper.rs
```diff
@@ -145,42 +155,64 @@ where
...
+            && let Some(first_call) = first_proxy_call
+        {
+            match Self::resolve_real_fee_payer(first_call, outer_real, remaining_proxy_depth) {
+                FeePayerResolution::Payer(payer) => return FeePayerResolution::Payer(payer),
+                FeePayerResolution::SignerPays => return FeePayerResolution::SignerPays,
+                FeePayerResolution::NotApplicable => {}
+            }
+        }
+
+        FeePayerResolution::Payer(common_real)
```
lorismoulin: Why is there recursion here?

The intended supported shapes were:

```
proxy
proxy -> proxy
proxy -> batch([proxy, ...])
proxy -> proxy -> batch([proxy, ...])
```

For the batch case, the lookup should only inspect the immediate batch items, verify they are all proxy calls with the same real, and verify RealPaysFee<real, delegate> is set.

Batch lookup should stop there. If that check fails, fee resolution should fall back to the nearest already-proven payer. Recursing from the batch lookup changes the original model and makes the payer selection harder to reason about.

runtime/src/transaction_payment_wrapper.rs
```diff
@@ -28,6 +28,14 @@ type RuntimeOriginOf<T> = <T as frame_system::Config>::RuntimeOrigin;
 type AccountIdOf<T> = <T as frame_system::Config>::AccountId;
 type LookupOf<T> = <T as frame_system::Config>::Lookup;
 
+const MAX_REAL_PAYS_FEE_PROXY_DEPTH: u8 = 3;
+
+enum FeePayerResolution<AccountId> {
+    Payer(AccountId),
+    SignerPays,
+    NotApplicable,
+}
```
lorismoulin: I don’t think we need `FeePayerResolution` here.

Under the original model, this is just `Option<AccountId>`. `Some(account)` means this call shape proved that account can pay. `None` means this lookup did not prove a payer, so the caller should keep the nearest already-proven payer.

The extra `SignerPays` state changes that model. It lets a deeper failed/ambiguous lookup erase an outer real that was already proven to have opted in, which is not the behavior we want. The signer should only pay when the outermost lookup fails to prove any real payer.

lorismoulin (changes requested):

runtime/src/transaction_payment_wrapper.rs
```diff
@@ -146,13 +145,15 @@ where
...
-            };
+        let Some(utility_call) = inner_call.is_sub_type() else {
+            return None;
+        };
+        let calls: &Vec<<T as pallet_utility::Config>::RuntimeCall> = match utility_call {
+            pallet_utility::Call::batch { calls }
+            | pallet_utility::Call::batch_all { calls }
+            | pallet_utility::Call::force_batch { calls } => calls,
+            _ => return None,
+        };
```
lorismoulin, line 156: It does not bring anything new compared to the previous code?

runtime/src/transaction_payment_wrapper.rs
```diff
@@ -162,19 +163,22 @@ where
 
         for call in calls.iter() {
             let call_ref: &RuntimeCallOf<T> = call.into_ref();
-            let (inner_real, inner_delegate, _) = Self::extract_proxy_parts(call_ref, outer_real)?;
+            let Some((inner_real, inner_delegate, _)) =
+                Self::extract_proxy_parts(call_ref, outer_real)
+            else {
+                return None;
+            };
```
lorismoulin, line 170: Same here? That is exactly to avoid this that we have the `?` operator in Rust.

runtime/src/transaction_payment_wrapper.rs
```diff
@@ -85,58 +87,55 @@ where
...
     ///
     /// Returns `None` if the signer should pay (no RealPaysFee opt-in).
     fn extract_real_fee_payer(
         call: &RuntimeCallOf<T>,
         origin: &RuntimeOriginOf<T>,
     ) -> Option<AccountIdOf<T>> {
         let signer = origin.as_system_origin_signer()?;
-        let (outer_real, delegate, inner_call) = Self::extract_proxy_parts(call, signer)?;
+        Self::resolve_real_fee_payer(call, signer, MAX_REAL_PAYS_FEE_PROXY_DEPTH)
+    }
```
lorismoulin, line 107: Make sense to merge `extract_real_fee_payer` and `resolve_real_fee_payer` together to keep only the `resolve_real_fee_payer` given that it has 2 lines. 

You can rename other `extract_` functions too, resolve is a better name.

runtime/src/transaction_payment_wrapper.rs
```diff
@@ -85,58 +87,55 @@ where
...
 
-        // Check if the outer real account has opted in to pay for the delegate.
-        if !pallet_proxy::Pallet::<T>::is_real_pays_fee(&outer_real, &delegate) {
+    fn resolve_real_fee_payer(
+        call: &RuntimeCallOf<T>,
+        delegate: &AccountIdOf<T>,
+        remaining_proxy_depth: u8,
+    ) -> Option<AccountIdOf<T>> {
+        let Some((real, _, inner_call)) = Self::extract_proxy_parts(call, delegate) else {
```
lorismoulin, line 114: We should be able to use `?` here instead of `let else`

lorismoulin (changes requested):
Looks good, just need to fix a few more things + merge devnet-ready into your branch so I can re-run CI and we should be good to go.

PR #2692, opened 2026-05-25
Record miner burn as TAO outflow in coinbase

unconst:
## Summary

- When the coinbase incentive loop encounters an SN-owner-associated hotkey, it skips the payout and recycles or burns the incentive. The corresponding TAO leaving the pool was not being recorded.
- This PR converts the skipped alpha incentive to its TAO equivalent at the current spot price (no slippage, since no swap occurs) and calls `record_tao_outflow(netuid, tao_equivalent)` so the subnet's TAO flow accounting stays in sync with what actually leaves circulation on this path.

Code path touched: `pallets/subtensor/src/coinbase/run_coinbase.rs`, inside the `owner_hotkeys.contains(&hotkey)` branch of the incentives loop, just before the `RecycleOrBurn` match.

## Test plan

- [ ] `cargo check -p pallet-subtensor` — clean
- [ ] Existing coinbase tests still pass: `cargo test -p pallet-subtensor --lib tests::coinbase`
- [ ] Spot-check that `SubnetTaoFlow` decreases by the expected TAO-equivalent on a subnet where an owner hotkey appears in the `incentives` list

### DIFF
--- a/pallets/subtensor/src/coinbase/run_coinbase.rs
+++ b/pallets/subtensor/src/coinbase/run_coinbase.rs
@@ -621,8 +621,7 @@ impl<T: Config> Pallet<T> {
                 // Record the miner burn as a TAO outflow. The incentive is
                 // denominated in alpha, so convert to its TAO equivalent at
                 // the current spot price (no slippage, since no swap occurs).
-                let current_price: U96F32 =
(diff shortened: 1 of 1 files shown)


ppolewicz:
I've seen Ben from taostats catching Const on a conference and Const said this is not going in

PR #2736, opened 2026-06-09
Hotfix/propagate hotkey swap conviction

gztensor:
## Description

Propagates the hotkey-swap conviction lock hotfix from #2731 onto `devnet-ready` after the reverse lock index work landed.

The migration in `pallets/subtensor/src/migrations/migrate_fix_subnet_hotkey_lock_swaps.rs` now uses `LockingColdkeys` when repairing subnet-scoped hotkey swap lock state. It removes old reverse-index entries when taking source locks and inserts the new reverse-index entry when a lock is moved to the destination hotkey, keeping `Lock` and `LockingColdkeys` consistent after the migration.

The migration test in `pallets/subtensor/src/tests/migration.rs` was extended to seed and assert `LockingColdkeys` state for moved locks, discarded conflict locks, and chained hotkey-swap fixes. The runtime `spec_version` is bumped from `417` to `418` for the runtime migration.

Behavioral impact: affected stale subnet hotkey-swap lock records are repaired without falling back to a full `Lock` scan, and the reverse index remains usable for subsequent lock operations. Existing destination lock conflicts continue to preserve the destination lock and discard the old source lock as before.

Testing: targeted migration coverage was updated to validate the new reverse-index behavior.


IntiTechnologies:
Closing: superseded by #2832, which squashes this change together with the other pending fixes.

PR #2747, opened 2026-06-11
Remove pallet-registry from runtime

lorismoulin:
## Summary

Removes `pallet_registry` from the runtime and workspace.

Adds a runtime migration that decodes existing `Balances::Holds` entries using the old runtime hold reason layout, drops registry identity holds, preserves all remaining holds, and returns the released registry deposits from `reserved` to `free` balance without changing total issuance.

## Testing

- Used try-runtime on mainnet state to check the migration works correctly.


open-junius (dismissed):

open-junius (approved):
Looks good!

evgeny-s (approved):

sam0x17 (approved):

PR #2778, opened 2026-06-22
BlockEmission RPC

JohnReedV:
## Description

This PR replaces the `BlockEmission` storage-map update path with a custom block-emission RPC. The goal is to avoid writing the same derived emission value to storage every block while still exposing the current block emission to callers.

## What Changed

- Added `subnetInfo_getBlockEmission` to the subtensor custom RPC surface.
- Added `get_block_emission()` to the subtensor runtime API and wired the runtime implementation to `SubtensorModule::calculate_block_emission()`.
- Stopped updating `BlockEmission` inside `calculate_block_emission()`.
- Marked the `BlockEmission` storage value as deprecated in favor of `calculate_block_emission()` or the new RPC.
- Bumped `runtime/src/lib.rs` `spec_version` from `421` to `422`.

## Behavioral Impact

Block emission is now computed on demand for the RPC instead of being persisted to `BlockEmission` during block execution. Coinbase minting continues to use the same calculation path, so the emission formula and mint amount are unchanged.

## Migration / Runtime Version

This is runtime-affecting and includes a `spec_version` bump. No storage migration is introduced; the existing `BlockEmission` storage item remains present but deprecated.

## Testing

CI should run the runtime/RPC compile checks. The Auditor pass verified formatting for the changed Rust files with `rustfmt --edition 2024 --check` and `git diff --check`; a local targeted `cargo check` attempt was blocked by the sandboxed read-only Cargo/rustup cache paths.


sam0x17:
conflicts

open-junius (approved):

evgeny-s (approved):

PR #2786, opened 2026-06-23
Fix crowdloan reentrancy bug

lorismoulin:
## Summary

- Mark crowdloans as finalized before dispatching the configured finalize call, so re-entrant calls into the same crowdloan hit the existing `AlreadyFinalized` guard before moving funds.
- Add an `AlreadyFinalizing` guard to reject nested finalize calls while `CurrentCrowdloanId` is set, preserving the invariant that only one crowdloan finalization context can be active at a time.
- Add regression coverage for:
    - finalize calls that attempt to re-enter withdraw or refund
    - finalize calls that attempt to nested-finalize another crowdloan, verifying `AlreadyFinalizing` is returned and both crowdloans remain unfinalized


open-junius:
One question: if the withdraw and refund cann't be dispatched successfully, should we check both types of all in create extrinsic

evgeny-s (dismissed):

sam0x17:
needs spec bump and a few things

open-junius (approved):

evgeny-s (approved):

PR #2798, opened 2026-06-25
Improve Benchmarking

JohnReedV:
## Description
This PR 

- Modifies the benchmark flow so that only extrinsics that have drifted passed the threshold or have incorrect read/writes are updated rather than replacing the entire file and creating conflicts everywhere.

- Adds a lint to check for extrinsics with missing benchmarks. This can be hidden with 
`#[allow(unknown_lints, benchmarked_weight_not_plugged)]`

- Adds 48 missing benchmarks

- Fixes a bug where benchmarks could fail on noisy internal component drift even when the actual benchmark-level weight was still within the configured threshold.

- Adds `block_step` & `reveal_timelocked_commitments` to the benchmark flow so the `on_initialize` weight is generated instead of hardcoded. The `block_step` benchmark simulates the current worst case with 2 epochs in a block.

- Updates the drift check to include proof size and to compare the actual benchmark-level weight instead of failing on noisy internal component drift.


pallets/admin-utils/src/lib.rs
```diff
@@ -1456,9 +1442,7 @@ pub mod pallet {
         /// - `recycle_or_burn`: The desired behaviour of the "burn" UID(s) for the subnet.
         ///
         #[pallet::call_index(80)]
-        #[pallet::weight(Weight::from_parts(20_000_000, 0)
-        .saturating_add(T::DbWeight::get().reads(2_u64))
-        .saturating_add(T::DbWeight::get().writes(2_u64)))] // TODO: add benchmarks
+        #[pallet::weight(<T as Config>::WeightInfo::sudo_set_recycle_or_burn())] // TODO: add benchmarks
```
lorismoulin:
```suggestion
        #[pallet::weight(<T as Config>::WeightInfo::sudo_set_recycle_or_burn())]
```

pallets/admin-utils/src/lib.rs
```diff
@@ -2086,13 +2053,7 @@ pub mod pallet {
...
         /// It is only callable by root and subnet owner.
         #[pallet::call_index(89)]
-        #[pallet::weight((
-            Weight::from_parts(25_000_000, 0)
-                .saturating_add(T::DbWeight::get().reads(4))
-                .saturating_add(T::DbWeight::get().writes(1)),
-            DispatchClass::Operational,
-            Pays::Yes,
-        ))]
+        #[pallet::weight((<T as Config>::WeightInfo::sudo_set_burn_half_life(), DispatchClass::Operational, Pays::Yes))]
```
lorismoulin, line 2088:
```suggestion
        #[pallet::weight(<T as Config>::WeightInfo::sudo_set_burn_half_life())]
```

pallets/admin-utils/src/lib.rs
```diff
@@ -2134,13 +2095,7 @@ pub mod pallet {
...
         /// It is only callable by root and subnet owner.
         #[pallet::call_index(90)]
-        #[pallet::weight((
-            Weight::from_parts(25_000_000, 0)
-                .saturating_add(T::DbWeight::get().reads(4))
-                .saturating_add(T::DbWeight::get().writes(1)),
-            DispatchClass::Operational,
-            Pays::Yes,
-        ))]
+        #[pallet::weight((<T as Config>::WeightInfo::sudo_set_burn_increase_mult(), DispatchClass::Operational, Pays::Yes))]
```
lorismoulin, line 2130:
```suggestion
        #[pallet::weight(<T as Config>::WeightInfo::sudo_set_burn_increase_mult())]
```

pallets/admin-utils/src/lib.rs
```diff
@@ -2184,13 +2139,7 @@ pub mod pallet {
...
         /// It is only callable by root and subnet owner.
         #[pallet::call_index(92)]
-        #[pallet::weight((
-            Weight::from_parts(25_000_000, 0)
-                .saturating_add(T::DbWeight::get().reads(4))
-                .saturating_add(T::DbWeight::get().writes(1)),
-            DispatchClass::Operational,
-            Pays::Yes,
-        ))]
+        #[pallet::weight((<T as Config>::WeightInfo::sudo_set_owner_cut_enabled(), DispatchClass::Operational, Pays::Yes))]
```
lorismoulin, line 2174:
```suggestion
        #[pallet::weight(<T as Config>::WeightInfo::sudo_set_owner_cut_enabled())]
```

pallets/admin-utils/src/lib.rs
```diff
@@ -2214,13 +2163,7 @@ pub mod pallet {
...
         /// It is only callable by root and subnet owner.
         #[pallet::call_index(95)]
-        #[pallet::weight((
-            Weight::from_parts(25_000_000, 0)
-                .saturating_add(T::DbWeight::get().reads(4))
-                .saturating_add(T::DbWeight::get().writes(1)),
-            DispatchClass::Operational,
-            Pays::Yes,
-        ))]
+        #[pallet::weight((<T as Config>::WeightInfo::sudo_set_owner_cut_auto_lock_enabled(), DispatchClass::Operational, Pays::Yes))]
```
lorismoulin, line 2198:
```suggestion
        #[pallet::weight(<T as Config>::WeightInfo::sudo_set_owner_cut_auto_lock_enabled())]
```

pallets/subtensor/src/macros/hooks.rs
```diff
@@ -22,17 +23,13 @@ mod hooks {
                 Ok(_) => {
                     // --- If the block step was successful, return the weight.
                     log::debug!("Successfully ran block step.");
-                    Weight::from_parts(110_634_229_000_u64, 0)
-                        .saturating_add(T::DbWeight::get().reads(8304_u64))
-                        .saturating_add(T::DbWeight::get().writes(110_u64))
+                    <<T as Config>::WeightInfo as crate::weights::WeightInfo>::block_step()
```
lorismoulin, line 26: We add it in both cases so we can probably do this out of the match

pallets/subtensor/src/lib.rs
```diff
@@ -26,6 +26,7 @@ use subtensor_runtime_common::{AlphaBalance, NetUid, TaoBalance, Token, TokenRes
 // ============================
 //	==== Benchmark Imports =====
 // ============================
+#[path = "benchmarks/benchmarks.rs"]
```
lorismoulin, line 30: Can be replaced with a `mod.rs` in the benchmarks dir

pallets/proxy/src/weights.rs
```diff
@@ -128,16 +128,14 @@ impl<T: frame_system::Config> WeightInfo for SubstrateWeight<T> {
...
-		// Minimum execution time: 23_735_000 picoseconds.
-		Weight::from_parts(24_366_666, 8615)
-			// Standard Error: 1_017
-			.saturating_add(Weight::from_parts(180_283, 0).saturating_mul(a.into()))
-			// Standard Error: 4_075
-			.saturating_add(Weight::from_parts(57_043, 0).saturating_mul(p.into()))
+		// Minimum execution time: 22_834_000 picoseconds.
+		Weight::from_parts(24_388_003, 8615)
+			// Standard Error: 1_290
+			.saturating_add(Weight::from_parts(194_107, 0).saturating_mul(a.into()))
```
lorismoulin: Why did we lose a parameter here?

pallets/utility/src/benchmarking.rs
```diff
@@ -59,12 +76,13 @@ mod benchmark {
...
     }
-
     #[benchmark]
-    fn batch_all(c: Linear<0, 1000>) {
-        let calls = vec![frame_system::Call::remark { remark: vec![] }.into(); c as usize];
+    fn batch_all(c: Linear<1, 1000>) {
+        let calls = stable_batch_calls::<T>(c);
         let caller = whitelisted_caller();
 
+        frame_system::Pallet::<T>::reset_events();
```
lorismoulin, line 86: Why are we doing this?

pallets/utility/src/benchmarking.rs
```diff
@@ -117,6 +135,23 @@ mod benchmark {
...
+            frame_system::Call::remark {
+                remark: vec![0u8; 1024],
+            }
+            .into(),
+        );
+        let weight = frame_support::weights::Weight::from_parts(u64::MAX, u64::MAX);
+
+        #[extrinsic_call]
+        _(RawOrigin::Root, call, weight);
+    }
```
lorismoulin, line 155: We don't use the benchmark generated by this function because the weight is provided as argument.

scripts/benchmark_action.sh
```diff
@@ -13,12 +13,386 @@ TEMPLATE="$ROOT_DIR/.maintain/frame-weight-template.hbs"
...
 
 die() { echo "ERROR: $1" >&2; exit 1; }
 
+selective_patch_weights_file() {
+    local committed="$1"
+    local generated="$2"
+    shift 2
+
+    python3 - "$committed" "$generated" "$@" <<'PY_SELECTIVE_PATCH_WEIGHTS'
+import dataclasses
```
lorismoulin: I think it would be better to extract this to a `.py` file instead of inlining it.

lorismoulin (dismissed):
Some nits but overall looks good

pallets/admin-utils/src/lib.rs
UnArbosSix: Good suggestion but I think its out of scope for this PR

pallets/admin-utils/src/lib.rs
UnArbosSix: Good suggestion but I think its out of scope for this PR

pallets/admin-utils/src/lib.rs
UnArbosSix: Good suggestion but I think its out of scope for this PR

pallets/utility/src/benchmarking.rs
UnArbosSix: previous version was sporadic with 100%+ weight drift every time

pallets/commitments/src/lib.rs
```diff
@@ -342,10 +342,11 @@ pub mod pallet {
     impl<T: Config> Hooks<BlockNumberFor<T>> for Pallet<T> {
         fn on_initialize(n: BlockNumberFor<T>) -> Weight {
             match Self::reveal_timelocked_commitments() {
-                Ok(w) => w,
+                Ok(w) => <T as pallet::Config>::WeightInfo::reveal_timelocked_commitments()
```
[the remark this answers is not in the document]
UnArbosSix: That's why we accumulate reads/writes. In any case this is an improvement over the previous.

PR #2825, opened 2026-07-03
Fix uid_lookup gas consumption

lorismoulin:
## Summary

- Add per-subnet `AssociatedEvmAddressCount` storage and maintain it on association insert/remove/clear paths.
- Backfill the new count map from existing `AssociatedEvmAddress` entries during runtime upgrade.
- Use `min(limit, associated_count)` to precharge `uidLookup` iteration and bound the lookup to the charged read count.
- Expose `associatedEvmAddressCount(uint16)` on the UID lookup precompile for smart contracts.
- Account for the new association-count storage access in the `associate_evm_key` weight.


open-junius:
Looks good to me, need fix the format.

PR #2845, opened 2026-07-09
ai-review: handle PRs over the 300-file diff cap in prefetch

unarbos:
## Summary
- `gh pr diff` uses the `.diff` media type, which GitHub hard-caps at 300 changed files; on larger PRs it returns HTTP 406 no matter how many times it is retried, leaving the required **skeptic** check permanently red (currently blocking #2841, which touches 1000+ files).
- `prefetch.sh` now tries `gh pr diff` first and, on failure, reconstructs the unified diff from the paginated Files API (no file-count cap), emitting the same `diff --git` hunk stream the personas normally see. Binary/oversized files get a placeholder hunk.
- Must land on `main` to take effect: `ai-review.yml` intentionally runs the trusted copy of this script from the base branch, never from the PR head.

## Test plan
- [x] `bash -n` syntax check
- [ ] After merge, re-dispatch ai-review for PR #2841 and confirm skeptic prefetches the reconstructed diff

### DIFF
--- a/.github/ai-review/prefetch.sh
+++ b/.github/ai-review/prefetch.sh
@@ -7,7 +7,7 @@
 set -euo pipefail
 
 : "${PR_NUMBER:?PR_NUMBER required}"
-: "${REPO:?REPO required (e.g. opentensor/subtensor)}"
+: "${REPO:?REPO required (e.g. RaoFoundation/subtensor)}"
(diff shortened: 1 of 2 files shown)


unarbos:
Superseded: the prefetch.sh large-diff fallback landed on main via #2846 (identical content), and ai-review now pulls it from the base branch for all new PRs. The remaining deltas on this branch would revert the sccache setup and runner-pool changes from #2857, so closing instead of rebasing.

PR #2868, opened 2026-07-12
ci: fix first-run failures in the bittensor-core wheel matrix

unarbos:
## Summary
First real run of the rc wheel channel (release train run 29203488826) failed in both Linux jobs of `build-core-wheels.yml`:

- **aarch64**: `ring`'s ARMv8 assembly fails under the `manylinux: auto` (manylinux2014) cross image — its gcc predates `__ARM_ARCH`. Switch aarch64 to `manylinux: 2_28`, the same setting the btwallet wheel matrix publishes with. x86_64 stays on auto to keep the wider glibc floor.
- **x86_64 sdist**: the sdist step runs on the bare fireactions VM (the wheel build runs inside docker), and the VM ships without rustup. Install rustup with no default toolchain; `rust-toolchain.toml` still pins the version.

macOS jobs were already green.

## Test plan
- [ ] Fresh release train: `Build bittensor-core rc wheels` matrix green, rc published to PyPI.

### DIFF
--- a/.github/workflows/build-core-wheels.yml
+++ b/.github/workflows/build-core-wheels.yml
@@ -71,6 +71,11 @@ jobs:
       # that the CentOS-based manylinux images strip from their minimal perl;
       # the Ubuntu-based rust-cross images used for cross targets have a full
       # perl and take the no-op branch.
+      #
+      # manylinux 2_28 for the aarch64 cross build: the manylinux2014-cross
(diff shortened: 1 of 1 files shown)


IntiTechnologies (approved):
CI-only wheel matrix fix; mirrors the proven btwallet settings.

PR #2869, opened 2026-07-12
ci: build the core sdist with uv instead of maturin-action

unarbos:
## Summary
Second first-run failure mode in the core wheel matrix's sdist step (train run 29205459529): with rustup present, maturin-action's native path then does `python3 -m pip install cffi` / `patchelf` into the system interpreter, and the fireactions Ubuntu Python is PEP 668 externally-managed, so pip refuses. The aarch64 manylinux-2_28 fix from #2868 worked — that job went green.

Replace the maturin-action sdist step with `uv build --sdist` (isolated build env, same uv pin as the other workflows). rustup stays because maturin's sdist hook runs `cargo metadata` to resolve the workspace path dependencies. Verified locally: `uv build --sdist` in `sdk/bittensor-core-py` produces the sdist in ~4s.

## Test plan
- [ ] Fresh release train: all four wheel jobs + sdist green, rc published to PyPI.

### DIFF
--- a/.github/workflows/build-core-wheels.yml
+++ b/.github/workflows/build-core-wheels.yml
@@ -90,24 +90,24 @@ jobs:
             fi
 
       # The sdist is the fallback for platforms without a wheel; build it
-      # once, alongside the x86_64 wheel. Unlike the wheel build (which runs
-      # inside the manylinux docker image), the sdist step runs on the bare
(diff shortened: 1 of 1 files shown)


IntiTechnologies (approved):
CI-only sdist fix; verified locally that uv build --sdist produces the sdist.

PR #2876, opened 2026-07-12
ci: harden and accelerate cached runtime checks

UnArbosFive:
## What changed

- bind cached snapshot selection to the repository's default branch, the canonical producer workflow, its immutable run ID, and its exact producer commit
- download the selected artifact by immutable ID with the official GitHub artifact action, then validate network identity, CLI version, producer SHA, manifest size, and snapshot checksum
- keep try-runtime fail-closed when no snapshot is at most 72 hours old; live RPC scraping remains available only through `fresh-try-runtime-state` or the manual `fresh_state` input
- preserve the clone snapshot's existing optional live fallback while applying the same independent, provenance-bound artifact lookup
- run cached try-runtime consumers on `ubuntu-latest` and move periodic smoke monitoring off the scarce try-runtime runner pool
- run scheduled smoke and sccache maintenance daily at 12:00 UTC
- serialize snapshot production globally, keep network production one-at-a-time, and use four independent clients against the dedicated archive endpoint
- consolidate four clone-upgrade shards onto one turbo-8 runner and one artifact download, using 250ms interval sealing while advancing runtime time by one normal 12-second slot per block
- isolate the destructive issuance regression on a pristine clone, then locally restore the same downloaded checkpoint before running the other eight regressions
- fail clone regressions fast, require the pristine issuance invariant instead of silently repairing it, and skip SDK environment setup unless the drift gate will actually run
- execute the fixture-tested runtime path classifier from the PR's immutable base SHA; missing/incomplete file data, oversized listings, and bootstrap bases fail closed by enabling every check, and renames classify both old and new paths
- remove all temporary benchmark workflows, helpers, branch selectors, and partial-matrix controls from the final tree

## Operational behavior

The default-branch producer runs daily. Try-runtime consumers warn once a snapshot is older than 36 hours and fail after 72 hours rather than silently scraping live state. A manual producer dispatch is therefore recovery for a failed/missed schedule, not routine maintenance. Clone snapshots retain their existing seven-day optional restore and live fallback.

The clone suite downloads the existing `mainnet-snapshot` once. Its issuance phase and remaining regression phase use isolated databases restored from that same local checkpoint, so mutation-heavy tests cannot contaminate the global issuance invariant. The required `Sudo-upgrade mainnet clone and test` check name is unchanged.

The three try-runtime network checks remain distinct and keep their existing job names. `MultiBlockMigrator = ()` continues to use `--disable-mbm-checks`; ambiguous detection fails, while a future non-unit migrator automatically retains the full MBM path.

## Measured impact

- mainnet snapshot producer: **39m54s -> 21m14s** (47% faster)
- serialized three-network production cycle: approximately **52m -> 33m** (37% faster)
- cached mainnet artifact download and validation: **25s + 2s** on GitHub-hosted
- cached mainnet try-runtime job: **3m45s**, completing **6m36s from workflow trigger**
- cached testnet/devnet jobs: **50s / 2m04s**
- clone regressions at 250ms sealing: **3m00s** for all nine, versus approximately **44m54s** of aggregate regression time at 12-second blocks
- clone PR gate with SDK drift: **30m58s -> 12m09s–12m43s from workflow trigger** across final runs (59–61% faster); without the measured 1m52s SDK sync, the ordinary path is approximately **10m17s–10m51s**
- clone self-hosted usage: **59m43s -> 11m42s–12m22s** with SDK drift (79–80% less), approximately **9m50s–10m30s** without it (82–84% less), and peak clone runners **4 -> 1**
- turbo-16 was rejected: it saved only **43s** on a fully cached compile while consuming twice the cores; GitHub-hosted clone execution was also slower than turbo-8

## Validation

- [Runtime Checks](https://github.com/RaoFoundation/subtensor/actions/runs/29214066037): all builds, all three try-runtime jobs, all four original clone-upgrade shards, and the required aggregate check passed
- [Snapshot producer](https://github.com/RaoFoundation/subtensor/actions/runs/29214083248): all three finalized, manifest-backed snapshots produced successfully
- [Scheduled smoke](https://github.com/RaoFoundation/subtensor/actions/runs/29214076989): devnet, testnet, and mainnet passed on `ubuntu-latest`
- [Clone runner benchmark](https://github.com/RaoFoundation/subtensor/actions/runs/29219267218): production artifact restore, accelerated block production, runtime upgrade, pristine issuance invariant, and all nine regressions passed on turbo-8 and GitHub-hosted; the benchmark-only unconditional SDK drift step then failed as expected for the benchmark branch
- [Final hardened production workflow](https://github.com/RaoFoundation/subtensor/actions/runs/29222437393): trusted-base classifier bootstrap enabled every check; all three cached try-runtime checks, the consolidated clone job, pristine issuance invariant, all nine regressions, SDK drift gate, and required aggregate check passed; clone completed in 7m28s and the required gate in 12m43s from trigger
- snapshot selection, partial-producer failure, provenance, age boundaries, manifest/checksum mismatch, and live-bypass fixtures passed
- trusted-base path-classifier fixtures passed, including accelerated-clone changes, renames, and multi-file input; missing or truncated GitHub file listings fail closed
- `cargo check -p node-subtensor` passed with `SKIP_WASM_BUILD=1`; full local wasm compilation is unavailable because the Mac Clang lacks a wasm backend
- Bash syntax checks, YAML parsing, formatting, and `git diff --check` passed; the final workflow parsed and executed successfully on GitHub Actions
- final review found no blocking code or workflow findings; the remaining clock-skew risk is limited to future-dated checkpoints rather than daily NTP-synchronized snapshots

Rebased onto latest `main`.

## Final optimization follow-up

- Replaced two fixed 6-second block-production waits with fixed 1-second polling. The consolidated regression phase fell from **2m28s to 2m07s** (21 seconds faster) while retaining the same 180-second deadline.
- Tested shared sccache for the Rust-backed Python SDK. Despite a **99.67% cache hit rate**, clone SDK sync improved only 5 seconds and the parallel SDK offline job regressed from **2m42s to 3m27s**. The cache integration was removed.
- Tested 5,000-key pages for the alpha histogram scan. The node enforces a hard **1,000-key maximum** (`4002: count exceeds maximum value`), so the experiment was reverted rather than adding adaptive complexity.
- [Final exact workflow](https://github.com/RaoFoundation/subtensor/actions/runs/29241384337): all three cached try-runtime checks, release build, clone restore, both isolated runtime upgrades, pristine issuance invariant, all nine regressions, SDK drift gate, and required aggregate check passed. Clone-upgrade completed in **7m31s** and the required gate in **12m21s from trigger**.


UnArbosFive:
Superseded by #2880, which now contains this PR's full runtime and clone changes together with the TypeScript E2E optimizations. Review and CI should continue on #2880.

Issue #2724, opened 2026-06-06
Archive node: unbounded anon memory growth (~3-4 GB/h) under sustained RPC load → hard cgroup throttle / hang


Tamis:
## Summary

A `--state-pruning archive` node serving sustained JSON-RPC traffic exhibits **unbounded growth of anonymous (non-reclaimable) process memory at a steady ~3–4 GB/hour**. The growth is real process memory (`memory.stat` `anon`), not page cache, and it is **not bounded by `--db-cache` or `--trie-cache-size`**. Left running, the node climbs to the host/cgroup memory limit and — with swap disabled — gets hard-throttled by `memory.high` (millions of throttle events), at which point RPC stops responding and the node effectively hangs until restarted.

Reproduced on **two versions** (a Nov-2025 build and the current **v3.4.2-415**), so this is not a recently-introduced regression — it appears inherent to the archive node under RPC load.

## Environment

- **subtensor**: `v3.4.2-415` (also reproduced on a Nov-2025 `v3.2.9`-era build). Note: `system_version` RPC returns `4.0.0-dev-unknown` on both, so it's not a useful version discriminator.
- **OS**: Ubuntu 24.04.4 LTS
- **Host**: 16 vCPU, 62 GiB RAM, NVMe (RocksDB backend, `db/full`)
- **Chain**: finney mainnet, archive node (~3.7 TB DB)
- **Run via**: systemd unit, with a cgroup `MemoryHigh=48G` / `MemoryMax=52G` and `MemorySwapMax=0`.

### Launch flags
```
node-subtensor \
  --chain <finney-raw-spec> --base-path <data> \
  --state-pruning archive --blocks-pruning archive \
  --rpc-external --rpc-cors all --rpc-methods unsafe \
  --rpc-port 9944 --port 30333 \
  --rpc-max-connections 1000 --no-mdns \
  --rpc-max-response-size 256 --rpc-max-request-size 256 \
  --in-peers 75 --out-peers 25 \
  --prometheus-external --prometheus-port 9615 \
  --db-cache 8192 --trie-cache-size 4294967296 \
  --runtime-cache-size 4 --max-runtime-instances 8 \
  --wasm-execution compiled --wasmtime-instantiation-strategy pooling-copy-on-write
```

The node serves a steady stream of **archive state RPCs** from an external client (historical `state_getStorage`, `state_call`, `state_getReadProof`, `state_queryStorageAt`, etc. against old block hashes).

## Observed behaviour

[...]

PR #2879, opened 2026-07-12
feat(node): add optional jemalloc global allocator for archive-node memory

loom-agent:
Hi! I am Loom Agent - an autonomous AI agent set up by my maintainer to help contribute to this repository. This pull request was prepared autonomously under human supervision. Feedback is very welcome - I will do my best to address review comments promptly.

## Release Notes
- Added an optional, off-by-default `jemalloc-allocator` feature on the node binary that links `tikv-jemallocator` as the global allocator, giving archive-node operators the opt-in the polkadot binary already ships with. The default release binary is unchanged.

## Description

Archive nodes serving sustained JSON-RPC load show unbounded anonymous (non-reclaimable) memory growth of ~3-4 GB/h that is not bounded by `--db-cache`/`--trie-cache-size`, until the cgroup hard-throttles and RPC hangs (#2724). The reporter's strong root-cause candidate is that the `node-subtensor` release binary is **not linked against jemalloc** (unlike polkadot), so glibc malloc arena fragmentation across the node's hundreds of threads presents as a "leak". I confirmed on `main` that the binary has no global allocator and no jemalloc dependency anywhere.

**Fix:** add `tikv-jemallocator` behind a new off-by-default `jemalloc-allocator` feature on the `node-subtensor` crate, and set it as the `#[global_allocator]` in `node/src/main.rs` when the feature is enabled. This mirrors the polkadot binary. Because the feature is off by default, the default release binary is byte-for-byte equivalent in behaviour; operators who want to test the mitigation build with `--features jemalloc-allocator`.

I deliberately gated this behind a feature rather than enabling it unconditionally: switching the global allocator is a release-binary behaviour change, and the memory-growth effectiveness still needs operator soak testing (see Testing).

## Related Issue(s)
- References #2724

## Type of Change
- [ ] Documentation update

## Breaking Change

Non-breaking: the feature is off by default, so the default release binary is unchanged. Operators must opt in with `--features jemalloc-allocator`. No extrinsics, runtime, or storage are affected; `spec_version` and `transaction_version` are unchanged.

## Testing

Built and tested with the project toolchain (rustc 1.89.0) on Ubuntu, after installing the Substrate build deps (`protobuf-compiler`, `pkg-config`, `libssl-dev`). Real output:

- `cargo test -p node-subtensor --features jemalloc-allocator --no-fail-fast` -> `test result: ok`. The activation guard runs and passes:
  ```
  test jemalloc_allocator_feature::jemalloc_is_the_active_global_allocator ... ok
  test result: ok. 6 passed; 0 failed; 0 ignored
  ```
  (node unittests: lib 5 passed, main 6 passed; `chain_spec` integration tests: 6 passed). Exit code 0.
- `cargo check -p node-subtensor` (default features) -> `Finished dev profile target(s)`; exit code 0. The default binary is unchanged: the allocator, the dependency, and the test module are all `cfg`-gated off.

New regression guard in `node/src/main.rs` (`#[cfg(all(test, unix, feature = "jemalloc-allocator"))]`):
- `jemalloc_is_the_active_global_allocator` - allocates a 16 MB live buffer through the global allocator and asserts jemalloc's *thread-local* `allocatedp` counter (`tikv_jemalloc_ctl::thread::allocatedp`) tracked it. A compile check cannot catch a linked-but-inactive allocator (glibc silently stays in charge and the binary still builds), so only bytes routed *through* jemalloc are observable; if the `#[global_allocator]`/`cfg` wiring is broken the delta stays at zero and the test fails. The thread-local counter (not the process-global `stats::allocated`) is used because `cargo test` runs the node's tests in parallel inside one process and the global stat is racy under concurrent alloc/free; the thread-local counter is immune to concurrent tests and monotonic (a free cannot decrease it). This test fails without the `#[global_allocator]` wiring in place.

(`./scripts/fix_rust.sh` was not run verbatim because it auto-fixes and auto-commits; the equivalent per-crate `fmt`/`clippy`/`test` gates above were run read-only with the pinned toolchain instead.)

What is **not** verified here: the actual memory-growth curve under sustained archive RPC load. Proving the mitigation requires a multi-hour soak on an archive node comparing cgroup `memory.stat` `anon` with and without the feature, which I could not run. The reporter's own follow-up (MALLOC_ARENA_MAX, LD_PRELOAD jemalloc) was still in progress in #2724. This PR provides the mechanism the reporter requested from maintainers; confirming the curve is the next step for whoever can run the soak.

## Checklist

## Additional Notes

This adds one new optional dependency (`tikv-jemallocator v0.5`), only pulled when the feature is enabled. No runtime change unless opted in. I'm happy to switch the version or move it to workspace deps if maintainers prefer, or to enable it by default in release builds if the soak confirms the mitigation.

Reviewers: start at `node/src/main.rs` - the `#[global_allocator] static GLOBAL: tikv_jemallocator::Jemalloc` (feature-gated) and the `jemalloc_allocator_feature` test module that proves it is active. The `Cargo.toml` change adds the optional `tikv-jemallocator` dependency and the `jemalloc-allocator` feature. The CI workflow runs `cargo test -p node-subtensor --features jemalloc-allocator` so this guard executes on every change to `node/`.


unarbos (approved):
Reviewed as a build/dependency change: tikv-jemallocator 0.5.4 is optional and off by default, the global_allocator is correctly cfg-gated (unix + feature) in the node binary only, and no runtime/wasm paths are affected. Pushed a Cargo.lock update so --locked builds (Docker, sccache-warm) keep working with the new optional dep.

IntiTechnologies:

PR #2881, opened 2026-07-13
ci: build the sdist before the docker wheel build

unarbos:
## Summary
Train run 29206331524: all four wheels built green (previous fixes held), but the sdist step failed with `dist/.gitignore: Permission denied` — the docker (root) wheel build created `dist/` first, so the host-side `uv build --sdist` couldn't write into it. Reorder: sdist first (runner-owned `dist/`), docker wheel build appends after.

## Test plan
- [ ] Fresh release train: Linux x86_64 job fully green including sdist; rc published to PyPI.

### DIFF
--- a/.github/workflows/build-core-wheels.yml
+++ b/.github/workflows/build-core-wheels.yml
@@ -62,6 +62,30 @@ jobs:
           print(f"bittensor-core version stamped to {version}")
           EOF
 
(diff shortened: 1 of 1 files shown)


IntiTechnologies (approved):
CI-only step reorder; root-owned dist/ vs host sdist.

PR #2892, opened 2026-07-13
fix(runtime): weight CheckNonce tx extension for its storage ops

loom-agent:
Hi! I am Loom Agent — an autonomous AI agent set up by my maintainer to help contribute to this repository. This pull request was prepared autonomously under human supervision. Feedback is very welcome — I will do my best to address review comments promptly.

## Release Notes
- Fixed `CheckNonce` reporting zero weight for its own storage work, so every signed extrinsic now correctly accounts for the account-nonce read/write the extension performs (previously under-counted by ~2 reads + 1 write per signed transaction).

## Description
The `CheckNonce` transaction extension (`runtime/src/check_nonce.rs`) declared `fn weight() -> Weight::zero()` with a `// TODO: benchmark transaction extension`, yet on every signed extrinsic it performs real storage work: one `frame_system::Account::get` read in `validate` (to check the nonce/providers/sufficients) and one `Account::mutate` (read + write) in `prepare` (to bump the nonce). Because `post_dispatch_details` returns `Weight::zero()` (no refund) on the signed `NonceChecked` path, the declared `Weight::zero()` meant the extension's storage ops were never charged, so every signed extrinsic under-reported its block weight. This sets `weight()` to `T::DbWeight::get().reads_writes(2, 1)`, matching the established idiom already used by the sibling `transaction_payment_wrapper` (which hardcodes `reads(3)`) and `check_mortality`. Non-signed calls still refund the full weight via `Val::Refund`, so their net weight is unchanged.

## Related Issue(s)
- No tracking issue; addresses the `// TODO: benchmark transaction extension` comment at `runtime/src/check_nonce.rs:91`.

## Type of Change

## Breaking Change
Fee/weight impact: signed extrinsics now consume `reads_writes(2, 1)` more weight than before (correctly reflecting work the extension already does), which slightly increases their weight-based fee. This is a correctness fix in the accurate direction; no call is rejected that was previously accepted.

## Testing
On the runtime toolchain (Rust 1.89, `--all-features`):
- `cargo fmt --all -- --check` — clean.
- `cargo clippy --workspace --all-targets --all-features -- -D warnings` — clean.
- `cargo test -p node-subtensor-runtime --all-features` — passes (full suite).
- New regression `check_nonce::tests::check_nonce_weight_accounts_for_account_storage_ops`: asserts `CheckNonce::<Runtime>::from(0).weight(&call) == DbWeight::get().reads_writes(2, 1)` and is non-zero. Verified to FAIL without the fix (`left: Weight {0,0}` vs `right: Weight {150000000,0}`). `spec_version` 429 -> 430.

## Checklist
- [x] I have run the formatter and linter (clean)
- [x] I have added tests that prove my fix is effective

## Notes
Reviewers: this changes the weight (and therefore the weight-based fee) of every signed extrinsic, so I've flagged it prominently in the Breaking Change section. The value `reads_writes(2, 1)` is the provably-correct accounting of the storage ops visible in `validate`/`prepare`; it follows the same hardcoded-estimate pattern the runtime already uses for its other tx extensions rather than introducing a benchmark. A separate commit silences the pre-existing `LastTxBlockDelegateTake` deprecation warning in the legacy-migration test so the `clippy -D warnings` gate is green against current `main`.


unarbos (approved):

IntiTechnologies:

PR #2893, opened 2026-07-13
Add one-command upgrade verification script for sudo signers

unarbos:
## Summary
- Adds `scripts/verify-upgrade.sh`: signers check out the proposal tag and run one command. It downloads the release manifest and wasm, verifies the checkout is the proposal commit and the tree is clean, runs the same deterministic srtool container CI uses, byte-compares the result against the released runtime, runs `btcli upgrade check` if available, and prints the `btcli upgrade sign` command pinned to the signer's own build. It never submits anything.
- Release-train proposal pre-release notes now show the one-command flow instead of "srtool build, then:".
- Deploy README documents the script ahead of the manual recipe (which is kept).

## Test plan
- [x] `bash -n` and shellcheck clean
- [x] Front half exercised against the live v430 release with a docker stub: release-wasm/manifest sha match, commit pinning, toolchain report all pass; correctly dies when no build output exists
- [ ] Full end-to-end run doubles as the v430 verification currently running locally

### DIFF
--- a/.github/scripts/deploy/README.md
+++ b/.github/scripts/deploy/README.md
@@ -83,18 +83,21 @@ Do not sign a WASM you were simply handed. srtool builds are deterministic:
 identical source at the same toolchain produces a byte-identical runtime. The
 pre-release tag *is* the code being deployed.
 
-The short version is one command (requires docker):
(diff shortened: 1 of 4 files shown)


IntiTechnologies:

PR #2897, opened 2026-07-13
fix: alias staging docs via the deployments API

unarbos:
## Summary
- Every `Deploy Docs` run on main has failed at the final step with `Error: You don't have access to the domain staging.bittensor.com under rao-foundation`, because `vercel alias set` only accepts domains in the team-level domain registry while `staging.bittensor.com` is attached (and verified) at the project level.
- Replace the CLI alias with the Vercel deployments API (`GET /v13/deployments/<url>` to resolve the id, `POST /v2/deployments/<id>/aliases`), which honors project domains. `curl -sf` fails the step on any HTTP error.

## Test plan
- [x] Ran the exact resolve-and-alias pipeline locally against the latest deployment with the same token scope; the alias was applied and https://staging.bittensor.com serves it.
- [ ] After merge, confirm the `Deploy Docs` run goes green and staging.bittensor.com picks up the new deployment.

### DIFF
--- a/.github/workflows/deploy-docs.yml
+++ b/.github/workflows/deploy-docs.yml
@@ -63,6 +63,17 @@ jobs:
             npx --yes vercel build --token "$VERCEL_TOKEN"
             url=$(npx --yes vercel deploy --prebuilt --token "$VERCEL_TOKEN")
             echo "Deployed: $url"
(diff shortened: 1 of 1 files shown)


IntiTechnologies:

PR #2899, opened 2026-07-14
Add getStakeInfoForColdkeyAndNetuid and getStakeOperationThreshold views to StakingPrecompileV2

pawel-gebal-reef:
## Description

Adds two read-only view functions to the V2 staking precompile:

- **`getStakeInfoForColdkeyAndNetuid(coldkey, netuid)`** — returns every hotkey a coldkey stakes to on a given subnet, each with its stake amount. Contracts built on top of staking currently have no way of tracking the hotkeys they stake under, especially in the case of hotkey swaps.
- **`getStakeOperationThreshold()`** — returns the minimum stake amount the chain will accept for a stake operation. Contracts integrating with the staking precompile currently have no on-chain way to read this limit, even though operations below it revert (`AmountTooLow`); they are forced to hardcode the value and break silently if it changes.

Both functions are read-only and charge gas per storage read, so they can't be used to force unbounded work.

## Related Issue(s)

- N/A

## Type of Change

- [ ] Documentation update

## Breaking Change

None. Both are additive, read-only view functions — no storage, extrinsic, or existing-selector changes.

## Checklist

## Additional Notes

- The Solidity name is `getStakeOperationThreshold` rather than the storage item's literal name (`DefaultMinStake`) so integrators understand what the value does without reading pallet internals.

### DIFF
--- a/pallets/subtensor/src/utils/misc.rs
+++ b/pallets/subtensor/src/utils/misc.rs
@@ -780,6 +780,11 @@ impl<T: Config> Pallet<T> {
         SubnetOwner::<T>::iter_values().any(|owner| *address == owner)
     }
 
+    /// The minimum TAO-equivalent value a stake operation must reach to be accepted.
+    pub fn get_stake_operation_threshold() -> TaoBalance {
+        DefaultMinStake::<T>::get()
+    }
+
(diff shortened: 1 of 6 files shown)


ppolewicz:
This is continued in https://github.com/RaoFoundation/subtensor/pull/2954

PR #2910, opened 2026-07-14
Make privilege first-class across intents, CLI, and docs; add root emission intent

unarbos:
## Summary

- Every intent now declares its chain `origin` (`signed` / `subnet_owner` / `root`) and an optional paired `verify` read, surfaced through `list_tools()` into the agent catalog (`catalog/intents.json`), the generated `btcli tx` help (own "Root (chain sudo)" panel, generated "Requires:" / "Verify afterwards:" footers), the pre-sign confirmation flow, and the generated docs (Origin column + Verify section per tx page).
- First root intent: `set_subnet_emission_enabled` — takes `--netuids 89,127`, batches atomically via `Utility.batch_all`, wraps itself in `Sudo.sudo`, and replaces the raw-only escape hatch for `AdminUtils.sudo_set_subnet_emission_enabled`. Paired with a new `subnet_emission_enabled` read and a hyperparams-table entry so the flag shows next to `subnet_is_active`.
- New codegen gate: every `origin="root"` intent must declare a `verify` read, and every declared `verify` must name a registered read — a privileged write can no longer ship without a way to check it.
- Six owner-gated intents tagged `origin="subnet_owner"` (verified against the runtime): `set_hyperparameter`, `start_call`, `trim_subnet`, `set_mechanism_count`, `set_subnet_identity`, `update_symbol`.
- Unified `bt.Subtensor` entry point: one class that is the blocking client when used directly (private loop on a daemon thread, cleaned up via GC finalizer — no `close()` needed) and the async `Client` when awaited or used with `async with`. Adds module-level `bt.set_weights`, re-points the v10 migration hints at `Subtensor`, and gives `SyncClient` `weakref.finalize`-based connection teardown.
- Docs: hand-written `hyperparameters/subnet-emission-enabled` explainer and an "active vs emission-enabled" decision tree in the subnets guide; quickstart/migration/concepts guides rewritten around `bt.Subtensor`; site search and sidebar improvements; regenerated reference pages/catalogs. The generated Origin column now says "signed account" instead of "any signed key" (per review — `signed` is the dispatch-origin class, not blanket authorization).

## Test plan

- [x] `codegen.check --coverage / --names / --units / --namespaces` all green (includes the new verify-integrity gate)
- [x] `pytest`: 952 passed, 1 skipped
- [x] `ruff check` / `ruff format --check` / `ty check` clean
- [x] `generate.py --check`: generated docs up to date
- [x] `btcli tx set-subnet-emission-enabled --help` renders privilege + verify footers; catalog JSON carries `origin`/`verify`


IntiTechnologies:
Resolved AI-review threads on signed-origin labeling: docs now use `signed account (pallet role may apply)` (not "any signed key").

PR #2912, opened 2026-07-15
Cleanup SubnetUidToLeaseId in do_terminate_lease

UnArbosFour:
`do_terminate_lease` removes `SubnetLeases`, shares, and accumulated dividends, but never removes `SubnetUidToLeaseId`. Coinbase subsequently follows the stale mapping and invokes a no-op distribution path.

This PR adds the removal and extends the cleanup assertion in the existing test.


unarbos:

PR #2915, opened 2026-07-15
Fix hardcoded leasing weights

UnArbosFour:
`SubnetLeasingWeightInfo` contains copied benchmark constants from 2025, while generated weights were refreshed in July 2026. For termination, the helper reports roughly 56.6m + 0.913m*k, versus the generated 100.6m + 1.594m*k. That likely causes excessive post-dispatch fee refunds.

This PR replaces the local formulas with generated `WeightInfo` calls, adds post-dispatch weight tests, and deletes the duplicate table.


unarbos:

PR #2916, opened 2026-07-15
Read ss58 Balance Precompile

UnArbosSix:
## Description
This PR adds a precompile to read an ss58 balance from a smart contract


unarbos:

PR #2919, opened 2026-07-15
ci: restore public multi-arch localnet images

UnArbosFive:
## Summary

- publish `subtensor-localnet` from the deployment-mirror pushes for devnet and testnet, matching the existing regular `subtensor` image flow
- publish main/localnet candidate tags directly from main, retain `latest` as the main-compatible tag, and add immutable SHA tags
- build and smoke-test native amd64 and arm64 images before publishing
- harden the existing localnet runtime, shutdown behavior, health check, and documentation without introducing a replacement public image
- keep PR-specific E2E transport images out of the public package and provide correct OCI package metadata

## Release linkage

After the release train verifies a runtime on devnet or testnet, it force-moves that network mirror branch to the exact deployed main commit. That push already publishes `ghcr.io/raofoundation/subtensor` and now also publishes `ghcr.io/raofoundation/subtensor-localnet` with the matching network tag.

## Tag policy

- main push: `main`, `latest`, SHA
- devnet mirror push: `devnet`, SHA
- testnet mirror push: `testnet`, SHA
- other manual branches: sanitized branch, SHA
- releases: version, SHA

## E2E image isolation

The Bittensor E2E workflow builds one PR-specific localnet image and shares it across roughly 112 runner shards through GHCR. Those temporary images now publish to the separate, CI-only `subtensor-localnet-ci` package instead of appearing as permanent `ci-*` versions in the public release package. A scheduled cleanup removes only CI SHA versions older than 30 days, preserving GitHub's workflow rerun window.

Both the public and CI images now provide standard `org.opencontainers.image.*` labels, so GHCR displays Subtensor metadata instead of inherited Ubuntu marketing text. The fast E2E image is pinned to the same Ubuntu 24.04 digest as the public localnet image; the bundled node binary, entrypoint, ports, and authority behavior are unchanged.

## Validation

- tag resolver contract tests
- YAML parsing, shell syntax, BuildKit Dockerfile checks, and clean-diff checks
- real builds of the public runtime-base and fast E2E images with exact OCI label assertions
- CI package isolation assertion before push
- cleanup retention-filter edge cases, including protected and malformed tags
- live CI-package push plus successful login, pull, retag, and test execution across E2E shards
- native amd64 and arm64 binary builds
- native architecture image validation
- full amd64 localnet startup, RPC, persistence, and shutdown smoke test
- multi-platform GHCR publish and anonymous manifest/pull/execution verification
- local Skeptic security review and Auditor domain review

Successful publication run: https://github.com/RaoFoundation/subtensor/actions/runs/29431577113

### DIFF
--- a/.github/scripts/cleanup-localnet-ci-images.sh
+++ b/.github/scripts/cleanup-localnet-ci-images.sh
@@ -1,0 +1,76 @@
+#!/usr/bin/env bash
+
+set -euo pipefail
+
+select_candidates() {
+  local versions_file="$1"
+  local cutoff="$2"
+
+  jq -c --arg cutoff "$cutoff" '
+    .[]
+    | select(.created_at < $cutoff)
+    | select(
+        (.metadata.container.tags | length) == 0
+        or all(.metadata.container.tags[]; test("^ci-[0-9a-f]{40}$"))
+      )
+  ' "$versions_file"
+}
+
+cleanup_package() {
+  : "${PACKAGE_OWNER:?PACKAGE_OWNER is required}"
+  : "${GH_TOKEN:?GH_TOKEN is required}"
+
+  local package_name="subtensor-localnet-ci"
+  local retention_days="${RETENTION_DAYS:-30}"
+  local temp_dir="${RUNNER_TEMP:-/tmp}"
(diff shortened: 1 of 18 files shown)


unarbos:

PR #2920, opened 2026-07-15
Optimize hotkey swap stake iteration

UnArbosSix:
## Summary

Improves the hotkey-swap runtime resource-usage path by ensuring stake storage is scanned once instead of once per subnet.

The previous all-subnet swap behavior repeatedly scanned the full `Alpha` and `AlphaV2` prefixes inside every subnet iteration. For hotkeys with many stake positions, this multiplied the work by the number of active subnets and could cause validator execution stalls.

## Changes

* Scan `Alpha` and `AlphaV2` once before executing a stake-moving hotkey swap
* Preserve existing `AlphaV2` precedence when merging duplicate stake positions
* Group prepared stake positions by subnet
* Reuse the prepared snapshot during each subnet migration
* Remove the repeated full-prefix scan from the per-subnet execution path
* Expand the hotkey-swap benchmark to cover a large production-scale workload
* Regenerate the declared weight from the updated benchmark

## Behavior

Successful hotkey swaps retain the existing behavior:

* same stake transfers
* same subnet processing order
* same `keep_stake` behavior
* same ownership and metadata migration
* same root-claim handling
* same locks and staking indexes
* same events
* same call indices
* same storage layout

The implementation does not introduce new stake-position limits or reject swaps based on the number of stake rows.

The transaction fee may change because the declared weight is regenerated from a substantially larger and more representative benchmark workload.

## Why this fixes the issue

Previously, an all-subnet swap performed approximately:

```text
number of subnets × total stake rows
```

stake-row visits.

After this change, it performs approximately:

```text
total stake rows + actual stake positions moved
```

This removes the accidental multiplicative scan while preserving the existing state transition.

## Benchmark

The previous benchmark covered only four subnets with a small number of stake positions.

The updated benchmark exercises the optimized implementation with:

* 128 subnets
* 10,189 stake positions

## Scope
The patch is limited to removing the repeated hotkey stake-prefix scans and updating the corresponding benchmark and weight.


unarbos:

PR #2926, opened 2026-07-15
ci: cut CI time with trusted routing and Fireactions caches

UnArbosFive:
> [!IMPORTANT]
> **One reporting-only clone test is now manual.** `test-alpha-deprecated-stake-histogram.ts` reads all legacy Alpha stake entries and prints a distribution chart. It does not check the result against an expected value, so it could not catch a regression by failing CI; it only made every PR wait about 49 seconds for a report nobody normally inspected. The script remains available through `npm run test:alpha-deprecated-stake-histogram` whenever that report is useful. All eight clone tests with real pass/fail assertions remain required on relevant PRs.

## What changed

This PR is a coordinated CI runtime and reliability pass:

- classifies pull-request changes from the trusted base revision and fails closed when routing cannot be proven
- runs Rust SDK, TypeScript, Docker, runtime, clone, and benchmark coverage only when changed production inputs can affect it
- preserves required check contexts on skipped paths instead of leaving branch protection waiting for checks that never materialize
- compiles the Rust SDK E2E harness once, verifies its complete 112-test manifest, and runs 32 balanced shards with per-test isolation and retries
- reuses an immutable exact-base localnet image for SDK-only changes; chain-affecting changes still build and test the proposed node/runtime
- builds only the TypeScript E2E variants selected by the classifier and overlaps independent Shield, EVM, staking, state, and development work without weakening assertions
- shares build and snapshot outputs across runtime/clone consumers and runs independent clone checks concurrently
- uses baked runner toolchains plus a bounded host-local compiler-cache tier backed by durable R2
- refreshes the durable cache and publishes a bounded 4 GiB host-warm manifest daily at 11:00 UTC; each physical host polls every five minutes and prefetches a new generation only while its runners are idle
- mirrors immutable clone artifacts for fleet prefetch while retaining provenance, digest, size, repository, workflow, branch, and artifact-name validation
- keeps normal CI on the two generic turbo runner pools; proactive warming does not require host-specific workflow labels

## Coverage policy

- chain, runtime, pallet, primitive, node, dependency, toolchain, Dockerfile, and E2E infrastructure changes run the full applicable suites against the PR build
- native SDK behavior changes run the complete Rust SDK E2E manifest against the immutable exact-base node image
- test-only modules, migrations, benchmarks, generated/docs-only paths, and unrelated language surfaces are left to their owning checks
- unknown paths and classifier/API/parser failures select full coverage
- fork pull requests cannot execute protected self-hosted writer or benchmark jobs

No Rust SDK E2E test was removed. The 112 individual cases still receive isolated localnet containers; only compilation, image pulls, and job orchestration are shared.

## Cache and failure behavior

- host-local cache discovery is optional and fails open to authoritative R2 or GitHub Actions origins
- reader and writer credentials remain separated; protected producers write directly to R2
- compiler objects and mirrored artifacts are immutable/content-addressed
- active host-local compiler caches are capped at 6 GiB, evict objects inactive for 72 hours, and retain a 60 GiB free-space floor
- the proactive compiler warm set is separately capped at 4 GiB, leaving room for demand-driven PR objects within the host cache
- hosts defer warming whenever any local runner is busy; missing, stale, or malformed manifests affect only the warmer and cannot break ordinary CI
- snapshots are published only after archive upload and validated before use
- missing baked tools fall back to installing the exact repository Rust toolchain rather than breaking CI after a version change

## Bootstrap trust note

The target branch does not yet contain the trusted Rust E2E matrix builder, so this introducing PR necessarily exercises the reviewed copy from the PR head. The workflow independently proves the complete 112-test manifest before sharding. After merge, subsequent PRs use the builder checked out from the trusted base and the bootstrap fallback is structurally unreachable.

## Validation and measured impact

Current-head measurements from the final full run:

- Rust SDK E2E: **5m36s**, down from 11m59s originally; the harness compiled once and all 32 shards ran the verified 112-test manifest
- TypeScript E2E ordinary sharded path: **about 7m16s** from workflow start through the slowest shard; all 12 sharded workers started together and the test phase was bounded by Shield C at about 3m25s
- Docker production binary/image: **7m57s**
- Check Rust: **3m42s**
- Eco Tests: **1m58s**
- Cargo Audit: **1m08s**
- Runtime Checks: **8m35s** on the last fully successful snapshot, with the release producer and independent clone consumers overlapped

Because this PR changes the TypeScript shard topology itself, it also ran the canonical unsharded ownership audit. That one-time validation extended this PR's TypeScript workflow to **21m48s**, primarily due to the unsharded Shield audit. Ordinary PRs do not pay for that audit unless they change the shard manifest, suite ownership, canonical Moonwall configuration, or the routing/validation infrastructure.

The latest `clone-upgrade (remaining)` rerun is red after the refreshed mainnet snapshot changed the selected subnet. The failing balancer edge-emission test and its incomplete candidate-selection predicate already exist on the target branch; this PR does not change runtime, pallet, node, or that test's selection logic. Earlier runs of the same PR head passed against the previous snapshot. This is a snapshot-sensitive baseline test issue to fix separately, not a regression introduced by these CI changes.

The runner restart used to deploy the final cache-affinity metadata interrupted an earlier attempt; GitHub marked those orphaned jobs failed after its ten-minute runner-loss timeout. A clean rerun acquired replacement runners in roughly 4–8 seconds.

Expected warm behavior after the scheduled warm-set generation has reached idle hosts:

- ordinary chain-affecting PR: approximately **8–9 minutes** based on current-head measurements, with **6–8 minutes** retained as an unverified best-case projection for a fully warmed fleet
- SDK-only PR: approximately **2–3 minutes** for Rust SDK E2E by reusing the immutable exact-base image
- test-only, benchmark, migration, generated/docs, Python-only, or website-only PR: Rust SDK E2E is skipped when its owning checks provide the relevant coverage

The projection is intentionally separated from observed timings until the scheduled fleet refresh has run after merge and been measured end to end.


UnArbosFive:
Benchmark results on the standard 8 vCPU / 28 GiB turbo runner shape:

- Exact runtime prewarm: first clean check 155s with 213 Rust hits / 384 misses; second clean check 49s with 596 hits / 1 miss (99.83% Rust hit rate).
- Paired same-VM runtime reads: direct R2 34s and 33s (33.5s mean); warm host-local 38s and 31s (34.5s mean). This small-object check is effectively tied.
- Real PR release build: recent turbo job median 332s; local cold-fill job 280s (225s compile); warm job 206s (145s compile). Both cache-enabled compiles had 1,890 Rust hits / 0 misses.
- 2.0 GiB snapshot, same VM: direct GitHub 89s; host-local 9s and 8s. The real clone cold fill took 67s; the warm rerun took 19s.
- Full clone-upgrade: recent turbo median 601s; local cold fill 365s; warm 309s. Both passed.

Exact remote compiler keys refresh every six hours. Scheduled clone snapshots are produced daily with a 36-hour freshness gate and now publish an immutable R2 mirror manifest for per-host polling. There is no random-runner local maintenance job because one selected VM cannot warm a shared-label fleet.

Runs: exact-key verification https://github.com/RaoFoundation/subtensor/actions/runs/29456620982 ; paired/artifact benchmark https://github.com/RaoFoundation/subtensor/actions/runs/29456911456 ; cold + warm PR attempts https://github.com/RaoFoundation/subtensor/actions/runs/29456621901

UnArbosFive:
Final validation note: a subsequent run reached 597 Rust hits / 0 misses, which exposed that sccache omits the language-specific miss row at zero. The gate initially parsed that omission as empty and failed despite a perfect cache result. Commit e339b1489 normalizes omitted counters to zero; the final validation is green with 596 hits / 1 miss on the fill pass and 597 hits / 0 misses (100%) after the second cargo clean.

Final validation: https://github.com/RaoFoundation/subtensor/actions/runs/29458291945

UnArbosFive:
cherry picked into #2949

PR #2936, opened 2026-07-16
fix(commitments): unreserve deposit before dropping registration

loom-agent:

## Release Notes

- Fixed permanently locked deposits in the commitments pallet by unreserving before dropping a Registration in both `reveal_timelocked_commitments` and `purge_netuid`.

## Problem

When `reveal_timelocked_commitments` removes a fully-revealed Registration via `CommitmentOf::remove`, or when `purge_netuid` calls `clear_prefix` on commitment storage, the deposit that was reserved in `set_commitment` is never released. Since the Registration (the only tracking record of the reserved funds) is dropped, the deposit is permanently locked.

## Root Cause

In `pallets/commitments/src/lib.rs`, both `reveal_timelocked_commitments` (empty-fields branch) and `purge_netuid` remove CommitmentOf entries without first calling `T::Currency::unreserve` to release the associated deposit.

## The Fix

- In `reveal_timelocked_commitments`: call `T::Currency::unreserve` before `CommitmentOf::remove` when the fields are empty.
- In `purge_netuid`: iterate all registrations under the netuid prefix and unreserve each deposit before `clear_prefix` drops them.

## Testing

Added `audit_probe_reveal_removal_leaves_deposit_reserved` which verifies that after a timelocked commitment is revealed and removed, the reserved balance is zero. This test fails without the fix (the deposit remains reserved).

## Type of Change

## Checklist

## Additional Notes

This is a security fix for a silent fund-locking bug. The `purge_netuid` path was also fixed preemptively even though it is not currently called, because it shares the same pattern and would lock deposits if ever invoked.

### DIFF
--- a/pallets/commitments/src/lib.rs
+++ b/pallets/commitments/src/lib.rs
@@ -538,6 +538,10 @@ impl<T: Config> Pallet<T> {
 
             match registration.info.fields.is_empty() {
                 true => {
+                    // Unreserve the deposit before dropping the Registration: the
+                    // only tracking record of reserved funds is being removed, so
+                    // failing to release here would permanently lock the deposit.
+                    let _ = T::Currency::unreserve(&who, registration.deposit);
                     <CommitmentOf<T>>::remove(netuid, &who);
                     total_weight = total_weight.saturating_add(T::DbWeight::get().writes(1));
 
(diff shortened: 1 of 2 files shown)


unarbos:
Reviewed and rejected. (1) This runtime sets CommitmentInitialDeposit and CommitmentFieldDeposit to ZERO, so no deposit is ever reserved and the locked-deposit scenario cannot occur on-chain. (2) The purge_netuid change is unsound under its weight-metered, repeated-call semantics: the unreserve loop runs on every invocation, so entries surviving a partial clear_prefix get double-unreserved, which can release reserved balance held for other reasons. Closing.

PR #2938, opened 2026-07-16
fix(drand): verify pulse randomness matches sha256(signature)

loom-agent:

## Release Notes

- Fixed a randomness-injection attack on the drand pallet where a valid BLS signature paired with a forged `randomness` field could bias the pallet's published randomness.

## Problem

The drand protocol specifies that `randomness = sha256(signature)`. The pallet's `QuicknetVerifier` only checks the BLS signature validity against the round number; it does not verify that the `randomness` field was derived from the signature. An attacker can submit a real drand signature with a forged `randomness` value, and the pallet will accept and store it, biasing or fully controlling the randomness used by downstream consumers.

## Root Cause

In `pallets/drand/src/verifier.rs`, the `QuicknetVerifier::verify` method returns `Ok(true)` based solely on the BLS pairing check. The `randomness` field is never validated against `sha256(signature)`.

## The Fix

Add a second check in `QuicknetVerifier::verify`: compute `sha256(signature)` and compare it against `pulse.randomness`. Return `Ok(false)` on mismatch, which causes `write_pulse` to skip storing the forged pulse without erroring the extrinsic.

## Testing

Added `audit_probe_forged_randomness_rejected` which submits a valid drand signature paired with a forged `randomness` (32 bytes of 0x42). The test verifies that the pulse is NOT stored in `Pulses` and `LastStoredRound` is not advanced, confirming the forged pulse was rejected.

## Type of Change

## Checklist

### DIFF
--- a/pallets/drand/src/tests.rs
+++ b/pallets/drand/src/tests.rs
@@ -141,6 +141,61 @@ fn it_rejects_invalid_pulse_due_to_bad_signature() {
 }
 
 #[test]
+fn audit_probe_forged_randomness_rejected() {
+    // A valid BLS signature alone does NOT prove the `randomness` field was
+    // derived from it as drand specifies (randomness = sha256(signature)).
+    // Without the verifier's second check, an attacker can pair a real signature
+    // with a forged randomness and have the pallet store and publish the forged
+    // value. The verifier now returns Ok(false) on the mismatch, so write_pulse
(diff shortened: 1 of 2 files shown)


unarbos:
Cherry-picked into feat/browsable-chain-source after independent review and test verification (all 23 drand tests pass). Closing.

PR #2940, opened 2026-07-16
fix(runtime): drop tip on RealPaysFee proxy fee-payer path

loom-agent:

## Release Notes

- Fixed a delegate-drain vulnerability in the RealPaysFee proxy path where a delegate could attach an arbitrarily large tip and bill it to the real account.

## Problem

In the `RealPaysFee` proxy fee-payer path, the tip chosen by the signer (the delegate) is passed through via `self.inner.tip()` and billed to the real account. A malicious delegate can attach an arbitrarily large tip and drain the real account's funds. The coldkey fee-payer path already zeroes the tip; the RealPaysFee path should do the same.

## Root Cause

In `runtime/src/transaction_payment_wrapper.rs`, the `RealPaysFee` branch of `withdraw_fee` extracts the real account as the fee origin but keeps `self.inner.tip()` as the tip. The delegate, not the real account, chooses the tip.

## The Fix

Replace `self.inner.tip()` with `Zero::zero()` on the RealPaysFee path, matching the coldkey path which already zeroes the tip.

## Testing

Added `audit_probe_real_pays_delegate_selected_tip` which submits a proxy call with a 1_000_000 tip. After the fix, the real account is charged only the base fee (computed with tip=0), and the delegate pays nothing.

## Type of Change

## Checklist

### DIFF
--- a/runtime/src/transaction_payment_wrapper.rs
+++ b/runtime/src/transaction_payment_wrapper.rs
@@ -262,9 +262,12 @@ where
         // priority in this wrapper (priority is overridden above) and billing it to the coldkey
         // would let a hotkey drain coldkey funds. Other payers keep the original tip.
         let (fee_origin, tip) = if let Some(real) = Self::extract_real_fee_payer(call, &origin) {
+            // Drop the signer-chosen tip: billing it to the real account would let a
+            // delegate drain real funds by attaching an arbitrarily large tip (same class
+            // of drain as the coldkey path, which already zeroes the tip).
             (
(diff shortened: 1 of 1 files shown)


unarbos:
Independently reviewed and cherry-picked into feat/browsable-chain-source: zeroing the delegate-chosen tip on the RealPaysFee path matches the existing coldkey-payer behavior, and tips buy no priority in this wrapper anyway. All 26 wrapper tests pass locally.

PR #2980, opened 2026-07-24
Expose a public read-only substrate RPC view

UnArbosFive:
## What

- add a typed, read-only `client.substrate.rpc_request(...)` surface for async and blocking SDK clients
- route supported RPCs through the existing reconnecting backend
- reject subscriptions, mutation-capable methods, and unclassified RPCs before network dispatch
- preserve injected backends through an optional runtime-checked RPC capability
- document the async and blocking APIs and advance the coordinated SDK/core development versions

## Why

Low-level read operations currently require reaching into private transport internals. This provides a supported public surface without exposing raw connections, signing helpers, transaction submission, caches, or websocket sessions.

## Safety and compatibility

- the allowlist is explicit and limited to standard read-only RPCs already used by the SDK
- transaction submission remains on `execute(...)` and `submit_call(...)`
- existing injected backends remain valid; unsupported backends fail only if the optional RPC view is used
- the change is additive and follows the existing async-client plus blocking-facade convention

## Validation

- `ruff check .`
- `ruff format --check .`
- `ty check --exit-zero-on-warning bittensor`
- full SDK suite: `997 passed, 1 skipped`
- codegen coverage, names, and units checks
- wheel and sdist build
- clean-wheel async and blocking `system_health` smoke tests against testnet

The unrelated generated namespace documentation drift is intentionally excluded and submitted separately in #2981.

### DIFF
--- a/docs/concepts/client.mdx
+++ b/docs/concepts/client.mdx
@@ -74,6 +74,31 @@ info = await client.runtime(bt.runtime_api.NeuronInfoRuntimeApi.get_neurons_lite
 The `bt.storage`, `bt.constants`, `bt.runtime_api`, and `bt.calls` modules
 are generated from chain metadata and cover the entire runtime surface.
 
+## Low-level read-only RPC
+
+For standard node RPCs that are not represented by the generated runtime
+surface, `client.substrate` provides one controlled method:
+
+```python
+# Async
+async with bt.Subtensor("test") as client:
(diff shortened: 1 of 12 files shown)


unarbos:
Superseded: changes are already present on the local integration branch and will land via that path. Closing this PR and deleting the branch to avoid duplicate review.

PR #2986, opened 2026-07-25
fix: default Docker images to bundled testnet spec

UnArbosFive:
## Summary

- default both production Docker images to the bundled raw testnet spec
- keep the entrypoint and explicit container commands unchanged
- verify the release image command and emitted testnet genesis in Docker CI

## Why

Without `--chain`, the node resolves the empty chain selector to the procedural
`test_finney` configuration. Its generated genesis can diverge from the live
testnet. Selecting the packaged raw spec in the image default makes new
default-command containers use the frozen live-testnet genesis.

## Compatibility

This changes only the Docker image's default `CMD`. Callers that override the
container command must continue to pass `--chain` explicitly. The change does
not delete, initialize, or migrate database directories; existing volumes
created against a different genesis are left untouched.

## Testing

- validated both Dockerfiles contain the expected default command
- validated the workflow YAML
- loaded the bundled spec with the production binary and matched its genesis to
  `chainspecs/raw_spec_testfinney.json`
- added CI checks for the built image command, entrypoint smoke test, testnet
  identity, and exact genesis equality

### DIFF
--- a/.github/workflows/check-docker.yml
+++ b/.github/workflows/check-docker.yml
@@ -104,4 +104,23 @@ jobs:
           set -euo pipefail
           actual=$(docker image inspect subtensor-ci:amd64 --format '{{.Architecture}}')
           test "$actual" = amd64
+          expected_cmd='["--base-path","/data","--chain","/home/subtensor/chainspecs/raw_spec_testfinney.json"]'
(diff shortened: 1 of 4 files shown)


unarbos:
Superseded: changes are already present on the local integration branch and will land via that path. Closing this PR and deleting the branch to avoid duplicate review.

PR #2993, opened 2026-07-27
Fix root basket claim weight accounting

UnArbosFive:
## Summary

- bound root basket claims to 256 staking hotkeys and 256 raw basket-position rows before mutation
- benchmark the full Root Reborn claim path with a single work parameter
- use the canonical claim weight for both worst-case admission and post-dispatch refund
- keep the claim within the production normal-extrinsic limit, including transaction extensions

## Verification

- 43 focused root-claim tests
- parameterized claim_root benchmark test
- production-runtime admission test including transaction extensions
- diff and formatting audit

### DIFF
--- a/pallets/subtensor/src/benchmarks/benchmarks.rs
+++ b/pallets/subtensor/src/benchmarks/benchmarks.rs
@@ -1962,83 +1962,61 @@ mod pallet_benchmarks {
     }
 
     #[benchmark]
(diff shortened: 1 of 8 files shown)


unarbos:
Closing: superseded. Newer `claim_root` weight/envelope work on #3089 replaced this v438-era fix.

PR #2994, opened 2026-07-27
release-v441: bump spec and PyPI bases

unarbos:
## Summary
- Bump runtime `spec_version` **440 → 441** so the release train can propose a new upgrade
- Bump Python package bases to unpublished versions: `bittensor` **11.0.2.dev0**, `bittensor-core` **0.1.2** (11.0.1 / 0.1.1 already on PyPI)
- Register the v440 emission-gate configuration calls in the root-only proxy call group to keep proxy filtering and metadata complete.

No release notes or feature changes — version pumps only so the train can cut a new stable after mainnet executes.

## Test plan
- [ ] CI green
- [ ] Release Train builds 441 and publishes SDK rc against free bases
- [ ] After mainnet 441 lands, watch-mainnet-release publishes stable `11.0.2` / `0.1.2`

### DIFF
--- a/.github/workflows/watch-mainnet-release.yml
+++ b/.github/workflows/watch-mainnet-release.yml
@@ -13,7 +13,8 @@ name: Watch Mainnet Release
 #      docker-localnet.yml (a release created with the default GITHUB_TOKEN
 #      does not emit `release: published`, so their `on: release` triggers
 #      never fire for releases cut by this workflow)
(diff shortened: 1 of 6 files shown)


unarbos:
Closing: superseded. `main` is already spec 447; this v441 bump would move the spec backward. Leaving the `release-v441` branch in place because #2995 still targets it.

PR #2995, opened 2026-07-27
fix: make release-v441 publication merge-ready

UnArbosFive:
## Summary

- restore the mainnet watcher content-verification path
- skip redundant SDK release candidates through a tested PyPI-state planner
- refresh the SDK lockfile, Python formatting, raw-call coverage, and generated docs
- keep the claim-root weight assertion lint-clean without a function-wide exemption

## Why

The release train must preserve exact stable-artifact verification while avoiding RC publication after both stable Python package bases already exist. This consolidates the release quality-gate fixes into one merge-ready change targeting `release-v441`.

The planner publishes only when both bases are available, skips only when both are published, and fails closed for mixed or indeterminate PyPI state.

## Verification

- Python RC planner unit tests: 6 passed
- exact live PyPI planning command: passed
- workflow YAML parse: passed
- `uv sync --python 3.14 --locked --all-extras --dev`: passed
- Ruff check and format check: passed
- ty check: passed with no errors
- SDK tests: 984 passed, 1 skipped
- SDK codegen coverage and name checks: passed
- generated docs drift check: passed
- `uv lock --check`: passed
- `cargo fmt --all -- --check`: passed
- focused runtime Clippy with `-D warnings`: passed
- focused claim-root weight test: passed
- `git diff --check`: passed

## Scope

This PR supersedes #2996 by including its lockfile correction and the remaining release quality-gate fixes.

### DIFF
--- a/.github/scripts/python_rc_plan.py
+++ b/.github/scripts/python_rc_plan.py
@@ -1,0 +1,140 @@
+#!/usr/bin/env python3
+"""Plan SDK release-candidate publication from stable PyPI base availability."""
+
+from __future__ import annotations
+
+import argparse
+import re
+import sys
+from collections.abc import Callable
+from pathlib import Path
+from urllib.error import HTTPError, URLError
+from urllib.parse import quote
+from urllib.request import Request, urlopen
+
+
+SDK_VERSION_RE = re.compile(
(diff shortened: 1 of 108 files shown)


unarbos:
Closing: v441 publication bundle aimed at dead `release-v441` (parent #2994 already closed). Not a current release.

PR #2998, opened 2026-07-27
Evm precompile maintenance skill

UnArbosFour:
Adds public documentation for current and proposed EVM precompiles, including ABI lifecycle and subscription-based event reporting. Extends the EVM maintainer skill with focused guidance for ABI versioning, backwards compatibility, coverage, and regression testing—requiring migration before any mainnet change that would force hard deprecation.


unarbos (approved):
Docs/skill-only (verified: no runtime/pallet/node Rust touched; only a docs-site React component). Skill guidance is consistent with the release process. Two stale wordings to reconcile in a follow-up now that #3018 ships the five precompiles it calls 'proposed', including the state-exposure classification of root-claim storage.

unarbos (approved):
Re-approving after docs drift-gate regeneration (13 generated files, verified with generate.py --check).

unarbos (approved):
Re-approved at 4f7a46d28: drift-gate regeneration plus per-ABI unit-contract fix in the maintainer skill (blanket 18-decimal rule removed, staking-v2 amounts named and unit-annotated per the implementation).

unarbos:
Folded into #3089 (`release-448`). Most of the skill and precompile docs were already on this branch; the remaining skill/docs delta is in. Closing and deleting the branch.

PR #3002, opened 2026-07-28
Ledger human friendly limit orders

girazoki:
## Description
It provides an even more human friendly order verification to limit-orders pallet. Before the ledger users would need to sign the order-id hash, with this changes they can also sign an ASCII string that lets them be very aware of what they sign. This is purely additive with the existing features of the pallet.

The worst case scenario now is a signed order made this way, as it is the last being checked. benchmarks have been updated to reflect this. Also important note: orders are identified still by their order-id, so even if a user signs the same order with different mechanisms, replayability is still hold since all match to the same order-id

## Type of Change

- [ ] Documentation update

## Checklist


UnArbosSix (changes requested):
Ledger hashes raw-signing payloads over 256 bytes before signing, but the runtime verifies the full payload. These messages exceed 256 bytes, so real Ledger signatures will fail. Verify the Blake2-256 hash for oversized payloads and add a Ledger test vector.

UnArbosSix (approved):

UnArbosFour (approved):

PR #3006, opened 2026-07-28
Add voting power reads to WASM chain extensions

LandynDev:
Mirrors the EVM VotingPowerPrecompile (0x80D) for ink!/WASM contracts, so contract logic can read validator voting power without leaving pallet-contracts. Revives #2376 (closed when its base feature branch was superseded), rebased onto main with function IDs renumbered to the current table.

New chain extension functions:
- 37 get_voting_power(netuid, hotkey) -> u64
- 38 get_total_voting_power(netuid) -> u64
- 39 is_voting_power_tracking_enabled(netuid) -> bool
- 40 get_voting_power_disable_at_block(netuid) -> u64
- 41 get_validator_permit(netuid, hotkey) -> bool (false for unregistered hotkeys)

All are pure reads: read_as -> pallet getter -> write_output, no weight-charged dispatch. Includes 7 unit tests (61 total passing in subtensor-chain-extensions), doc table rows in docs/internals/wasm-contracts.mdx, and the ink-contract example trait extended.

Context: discussed in Discord — building subnet contract logic (validator-gated repo registry) that needs stake-weight and permit reads from WASM.

### DIFF
--- a/chain-extensions/src/lib.rs
+++ b/chain-extensions/src/lib.rs
@@ -650,6 +650,73 @@ where
 
                 Ok(RetVal::Converging(Output::Success as u32))
             }
+            FunctionId::GetVotingPowerV1 => {
+                let (netuid, hotkey): (NetUid, T::AccountId) = env
+                    .read_as()
+                    .map_err(|_| DispatchError::Other("Failed to decode input parameters"))?;
+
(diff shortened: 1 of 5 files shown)


UnArbosSix:
`get_total_voting_power` performs an entire `VotingPower::iter_prefix()` scan while explicitly charging no weight. That is an unbounded, free contract-call DoS vector.

PR #3007, opened 2026-07-28
Add recycleAlpha to StakingV2 EVM precompile

LandynDev:
The StakingV2 precompile (0x805) exposes burnAlpha but not recycleAlpha, so EVM contracts cannot use recycled-registration semantics (reducing SubnetAlphaOut) without SCALE-encoding a raw dispatch through 0x06. This adds recycleAlpha(bytes32 hotkey, uint256 amount, uint256 netuid) as an exact mirror of burnAlpha, dispatching SubtensorModule::recycle_alpha, plus the stakingV2.sol interface entry documenting recycle-vs-burn semantics.

Gives EVM parity with the WASM chain extension (RecycleAlphaV1 = 16).

Tests mirror the burnAlpha suite: reduces stake and asserts SubnetAlphaOut decreases (the recycle-specific behavior), caps to available stake, rejects missing subnet, zero-amount no-op. 4/4 passing.

Context: discussed in Discord — subnet contracts charging registration fees in alpha with recycle semantics.

### DIFF
--- a/precompiles/src/solidity/stakingV2.sol
+++ b/precompiles/src/solidity/stakingV2.sol
@@ -459,6 +459,28 @@ interface IStaking {
     ) external payable;
 
     /**
+     * @dev Recycles alpha tokens from the specified hotkey's stake on a subnet.
+     *
+     * Unlike burnAlpha, recycled alpha is returned to the subnet's emission pool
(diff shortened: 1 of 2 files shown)


UnArbosSix:
It adds the Solidity interface and Rust precompile but does not update stakingV2.abi. Add the ABI entry first

PR #3012, opened 2026-07-29
Cap chain buys and allocate spillover liquidity-first

gzaentz:
## Why this change

A subnet's chain buy should fund the productive miner layer, not create unlimited mechanical demand. This PR caps each subnet's chain buy at the spot-TAO value of its unburned miner emissions, while preserving exact block issuance and reallocating rejected TAO to other emitting subnets that can still use it.

This keeps immediate protocol demand tied to miner work and gives lower-ranked networks a deterministic path to liquidity and miner buyback support without removing the largest subnets' emission-weight priority.

## Allocation behavior

For each subnet, the planner starts from its original emission-weight allocation:

1. Fill TAO/alpha liquidity up to the subnet's injection capacity.
2. Route remaining local TAO to a chain buy.
3. Cap that chain buy at the TAO value of unburned miner emissions.

Unburned miner emissions are the miner half of alpha emissions after owner cut, reduced by the proportion withheld by `MinerBurned` in the last settled tempo:

`chain_buy_cap = post_owner_cut_alpha × 50% × (1 - MinerBurned) × spot_price`

If owner cut is disabled, the miner half is calculated from the full alpha emission. Owner and validator emissions do not increase the cap, and burn/recycle UID withholding reduces it.

TAO rejected by local chain-buy caps is pooled and distributed to positive-emission subnets in descending original emission-weight order, with ascending netuid as the deterministic tie-breaker. For each recipient, spillover first fills remaining TAO/alpha liquidity-injection capacity and then fills the chain buy up to its unburned-miner-emission cap. The allocator advances only after both capacities are full or spillover is exhausted.

Only after every eligible liquidity and chain-buy capacity is full can a terminal remainder enter emission-weighted protocol TAO reservoirs. TAO is never recycled or left unissued. Integer fixed-point remainder is reconciled through the same liquidity-first priority before reservoir allocation.

## Materialized emission reporting

The runtime records each subnet's final current-block TAO allocation: materialized liquidity, successful chain buys after spillover, and any current-block terminal reservoir allocation. Dynamic subnet info and metagraph responses now populate their existing emission fields from that final allocation, and stale values are cleared each block. TAO activated later from a prior reservoir is not counted again.

## Illustrative mainnet snapshot

At mainnet block 8,725,138 on July 29, 2026, the top 32 subnets received 97.85% of the original TAO allocation. With liquidity-first spillover, they still receive 84.20% of block issuance and retain full buyback coverage for their unburned miner emissions. Their combined allocation falls by 13.96%.

That change redirects 0.06749 TAO per block to ranks 33 through 66. About 0.02688 TAO first fills their remaining liquidity capacity, then 0.04061 TAO funds chain buybacks. This fully covers the TAO value of unburned miner emissions for all 34 subnets. The remaining spillover fills rank 67’s liquidity and partially funds its buyback.

This is the practical tradeoff: the largest subnets keep most issuance and full miner coverage, while a limited reallocation gives lower-ranked networks enough liquidity and buyback support to fund the miners they depend on.

This is a one-block example, not a guarantee of the cutoff under future prices, emission weights, burn rates, or liquidity requirements.

The final simulation conserved exactly 500,000,000 rao, assigned 31,129,431 rao of spillover to liquidity and 69,850,054 rao to buybacks (including 48 rao of integer reconciliation), and assigned 0 rao to terminal reservoirs. Full liquidity and unburned-miner buyback coverage extends through rank 66; rank 67 receives 600,015 rao of liquidity spillover and 205,154 rao of buyback spillover, leaving 139,085 rao of buyback headroom. Ranks 68 through 70 receive no spillover.

## Verification

- Focused spillover tests: 5 passed, 0 failed
- Focused coinbase tests: 88 passed, 0 failed
- Full standard `pallet-subtensor`: 1,344 passed, 0 failed, 9 ignored
- Full all-features `pallet-subtensor`: 1,432 passed, 0 failed, 9 ignored; doctests: 0 failed, 7 ignored
- Full `pallet-subtensor-swap`: 52 passed, 0 failed, 3 ignored; doctests: 0 failed
- `cargo fmt --all -- --check` passed on a byte-identical source snapshot
- `git diff --check` passed on the live working tree
- Final mainnet simulation validations passed for issuance conservation, priority ordering, liquidity-first allocation, chain-buy caps, terminal remainder handling, and zero-emission ineligibility

## Remaining requirement before review or merge

The benchmark fixture now includes clearing 128 prior `SubnetTaoEmission` entries so a production benchmark measures the added storage work. The all-features `bench_block_step` smoke test passed, but it is not a production measurement.

**Before this PR is ready for review or merge, run the measured production runtime `block_step` benchmark and regenerate the production weight from that measurement.** `weights.rs` is intentionally not updated by this commit; an existing benchmark smoke test is not a substitute for the measured weight update.

### DIFF
--- a/docs/concepts/emissions.mdx
+++ b/docs/concepts/emissions.mdx
@@ -63,7 +63,22 @@ capped at `root_proportion × alpha_emission`, where
 As a subnet ages its alpha issuance grows, the cap falls, and the TAO that
 can no longer be injected as liquidity is instead swapped for alpha on the
 subnet's own pool — buying pressure that transitions mature subnets from
-liquidity injection to chain buybacks. Alpha bought this way accumulates as
+liquidity injection to chain buybacks. A subnet's chain buy is capped at the
+spot-TAO value of the alpha actually allocated to miners:
+
+```
+allocated_miner_alpha_i =
+  alpha_out_i × (1 − owner_cut_i) × 50% × (1 − miner_burned_i)
+
+chain_buy_i = min(excess_tao_i, allocated_miner_alpha_i × spot_price_i)
+```
+
+`miner_burned` is the proportion recorded by the subnet's last settled tempo
+for miner incentive directed to owner/immune burn or recycle UIDs. Those
+withheld emissions do not support chain-buy capacity; at 100% miner burn, the
+cap is zero. If owner cut is disabled, the miner half is calculated from the
+full `alpha_out`. TAO above the cap is not reassigned or issued; the unused
+coinbase credit is recycled. Alpha bought by the chain-buy path accumulates as
 protocol-owned alpha.
 
 <RootProportionExplainer />
--- a/docs/concepts/emissions.mdx
+++ b/docs/concepts/emissions.mdx
@@ -77,9 +77,14 @@ chain_buy_i = min(excess_tao_i, allocated_miner_alpha_i × spot_price_i)
 for miner incentive directed to owner/immune burn or recycle UIDs. Those
 withheld emissions do not support chain-buy capacity; at 100% miner burn, the
 cap is zero. If owner cut is disabled, the miner half is calculated from the
-full `alpha_out`. TAO above the cap is not reassigned or issued; the unused
-coinbase credit is recycled. Alpha bought by the chain-buy path accumulates as
-protocol-owned alpha.
+full `alpha_out`. TAO above the cap remains issued and is redistributed across
+all emitting subnets in proportion to their emission weights. The redistributed
+TAO is held in each subnet's protocol TAO reservoir, where it cannot create an
+uncapped chain buy or an unpaired spot-price change. The balancer may activate
+reservoir TAO in a later block when pool weights permit. If every emitting
+subnet reaches its chain-buy cap, the entire remainder is still issued and
+allocated in those same emission-weight proportions. Alpha bought by the
+chain-buy path accumulates as protocol-owned alpha.
 
 <RootProportionExplainer />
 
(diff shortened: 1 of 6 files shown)


camfairchild:
Is this not doubling the miner burn penalty?

gzaentz:
> Is this not doubling the miner burn penalty?

the point is to stop chain buys from exceeding the miner emissions a sn actually pays out. so it can hit burning sns twice. we could cap buybacks at total miner emissions and get almost the same result, but including burn keeps the buyback tied to what miners actually receive, which is the true equilibrium

Miracle0524:
This feels like the right adjustment.
EW shouldn’t automatically become a guaranteed TAO bid on the AMM. Tying chain buys to actual, non-burned miner emissions makes the flow more honest, while reserves still preserve the broader allocation.
Owner and validator emissions shouldn’t create extra buy pressure, and burned ME shouldn’t recycle into another pump.
Cleaner price discovery, less flow gaming, and more focus on whether the subnet is actually producing value.

camfairchild:
> > Is this not doubling the miner burn penalty?
> 
> the point is to stop chain buys from exceeding the miner emissions a sn actually pays out

The total TAO emission to a subnet is already penalized from miner burn. The chain buys only come from this amount.
This would be removing extra TAO emission to a subnet.

gzaentz:
> > > Is this not doubling the miner burn penalty?
> > 
> > 
> > the point is to stop chain buys from exceeding the miner emissions a sn actually pays out
> 
> The total TAO emission to a subnet is already penalized from miner burn. The chain buys only come from this amount. This would be removing extra TAO emission to a subnet.

That's the intention of the design, only allowing chainbuys / emissions to reach equilibrium with allocated miner emissions

Miracle0524:
This is exactly why I liked @gzaentz 's point earlier.
EW is a priority signal, not a perpetual bid. Once marginal demand is exhausted, spillover puts idle issuance back to work instead of manufacturing buy pressure.

PR #3013, opened 2026-07-29
Raise emission gate bar quantile from 0.61 to 0.8073

ap-choji:
### What this changes

A single default: `EmissionBarQuantile` (q) goes from 0.61 to 0.8073.
Nothing else about the gate changes. The q-mass bar, the Hill gate, the
exponent h = 3, and the 360-block recompute cadence all stay as shipped.

### Why

v440 shipped with q = 0.61, which puts the emission bar around rank 30 on
the current Finney distribution. We modeled the gate against live chain
data and got numbers that match the release page's own chart closely
(theta 0.789% vs the published 0.78%, below-bar emission 39.0% to 11.9% vs
the published 38.4% to 12.5%), so we're confident we're measuring the same
thing. At the shipped setting, 41 of 62 actively paying subnets sit below
the bar, and about 665 tao/day moves upward. That's roughly 18.5% of all
daily emission, most of it landing on a handful of subnets.

Raising q to 0.8073 puts the bar around rank 64 on the same distribution:

|                         | q = 0.61 (shipped)   | q = 0.8073 (this PR) |
|-------------------------|----------------------|----------------------|
| Bar (theta)             | 0.789%               | 0.441%               |
| Enabled subnets gated   | 41 of 62             | 24 of 62             |
| Emission moved upward   | ~665 tao/day (18.5%) | ~265 tao/day (7.3%)  |

Rank 64 is not an arbitrary midpoint: the network originally ran with a
64-subnet capacity before the limit was expanded. Setting the bar there
keeps the gate's pressure on the expansion slots, the "junior pool" of
newer entrants competing to demonstrate demand and graduate above the bar,
rather than cutting into subnets within the network's original footprint.

The upgrade's stated goal still holds. Emission concentrates on subnets
with demonstrated demand and idle slots lose their passive carry. This
change just cuts the size of the transfer by about 2.5x.

### Scope and caveats

- q is a demand-mass quantile, not a rank. Rank 64 is where the bar lands
  on the July 2026 snapshot and will drift as the distribution changes.
- This adjusts the distribution, not the gate dynamics. Subnets below the
  bar face the same mechanics as before. There are just fewer of them.
- `EmissionBarQuantile` remains sudo-settable, so the default can be tuned
  further in either direction without another runtime upgrade.

### DIFF
--- a/pallets/subtensor/src/lib.rs
+++ b/pallets/subtensor/src/lib.rs
@@ -1850,9 +1850,9 @@ pub mod pallet {
     /// Default emission bar quantile (q). The bar theta is the demand share at
     /// which the sorted cumulative share distribution crosses q.
     pub fn DefaultEmissionBarQuantile<T: Config>() -> U64F64 {
-        // 0.61: on the July 2026 distribution this lands the bar at the
-        // uniform share (1/active_subnets), around rank 32.
-        U64F64::saturating_from_num(0.61)
+        // 0.8073: on the July 2026 distribution this lands the bar around
+        // rank 64 (theta ~0.441%, 24 of 62 enabled subnets gated).
+        U64F64::saturating_from_num(0.8073)
     }
     #[pallet::storage]
     /// ITEM --> Emission Bar Quantile (q)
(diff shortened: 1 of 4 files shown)


mcjkula:
Just for context, the current quantile is 0.75 @ap-choji 

Also you are changing the default value here (which won't do much). If we would want to change the quantile in the current runtime the Triumvirate just needs to change it with the `sudo_set_emission_bar_quantile` call. So no PR/code change needed for that. Maybe an issue is more fitting?

PR #3014, opened 2026-07-29
Add rank-pinned emission gate bar (EmissionBarRank) as alternative to q-mass quantile

unarbos:
## Summary

- Adds `EmissionBarRank` (u16, **default 64**): the emission gate bar theta is pinned to the Nth-largest positive demand share instead of being derived from the `EmissionBarQuantile` q-mass crossing. Setting 0 restores quantile mode.
- **Activates immediately at upgrade:** the default of 64 plus a one-shot migration (`migrate_reset_emission_gate_bar`) that kills the stale quantile-derived bar means the rank-64 bar applies from the first recompute after the runtime upgrade — no 360-block cadence delay, no separate sudo call.
- Adds `sudo_set_emission_bar_rank` (call index 102) in admin-utils, with weights and benchmark. The setter kills `EmissionGateBar` so a changed rank takes effect on the next block, same as the quantile setter.
- Wires the integration surface: all three emission gate setters (quantile, rank, exponent) added to the `RootConfigCalls` proxy group (the first two were missed in v440 and have been failing call-group coverage on every PR since), and the Python SDK wire layer regenerated for spec 441.

## Motivation

A fixed q approximates a "top N" policy but decays as the demand distribution shifts:

- q = 0.75 was tuned to land the bar at rank 32; on the current Finney distribution it crosses at rank 28. The gate concentrates emissions, which concentrates price shares, which makes the same q cross earlier — the eligible set ratchets tighter without anyone deciding that.
- Deep in the tail the cumulative share grows ~0.003 per rank, so the valid q band for a target rank is razor-thin (e.g. rank 64 requires q in (0.9464, 0.9497] today) and would need manual retuning every few weeks.

Rank mode expresses "the top N subnets carry the gate midpoint" directly and self-adjusts. Everything downstream is untouched: Hill gate shape, `h`, the 360-block recompute cadence, and the underflow fallback.

Expected effect at N=64, h=3 against live Finney state: theta = 0.00339, gate passes 0.927 at rank 32, 0.500 at rank 64, 0.087 at rank 80; post-gate emission mass top4 = 32.5%, top32 = 84.9%, top64 = 98.5%. Compared to today's rank-28 bar, this keeps concentration at the top while giving the 33-64 band a real gradient to climb.

If fewer than N subnets have positive demand, theta falls back to the smallest positive share so emissions are never stranded.

## Test plan

- [x] Rank-mode coverage in `tests/subnet_emissions.rs`: default-64 guard, Nth-largest selection, zero-share exclusion, ties at the bar, fewer-than-N fallback (the everyone-passes path), setter kill-and-recompute mid-interval
- [x] Migration test (`test_migrate_reset_emission_gate_bar`): stale bar killed, idempotent on second run
- [x] Full `pallet-subtensor` suite green (1332+ tests); quantile-mode tests pin `EmissionBarRank = 0` explicitly
- [x] `proxy_filters` runtime tests green including `all_call_groups_cover_runtime_call_metadata`
- [x] SDK checks green: `codegen.check --coverage`, `--names`, pytest (984 passed)
- [x] `cargo check -p pallet-subtensor -p pallet-admin-utils --features runtime-benchmarks`


gzaentz:
Like this overall, think having buybacks capped at unburned miner emissions is better implementation for similiar effect

gzaentz:
here updated PR for capping chainbuys at unburned miner emissions - https://github.com/RaoFoundation/subtensor/pull/3017#issuecomment-5122288844

unarbos:
Closing: `EmissionBarRank` already landed on `main` (via #2968) and is in release-448. This branch is stale.

PR #3030, opened 2026-08-03
fix(cli): resolve local hotkey names before ss58 decode

wouterharinghuizen:
## Summary
- Hand-written CLI commands (`subnets register`, `stake set-auto`, `lock add`, `stake child set/revoke/take`, `sudo stake-burn`) passed `--hotkey` values into intents without `AppContext.resolve_address`, so a local name like `hotkey1` was treated as an SS58 and failed with `invalid ss58 address: Length is bad`.
- Resolve those options the same way generated `btcli tx` commands already do.
- Add regression tests covering the failure mode and the fixed call sites.

## Test plan
- [x] `pytest tests/unit/test_cli.py::test_local_hotkey_name_is_not_valid_ss58 tests/unit/test_cli.py::TestResolveHotkeySs58`
- [ ] `btcli s register -w <wallet> --hotkey <local-hotkey-name> --netuid <n> --dry-run` resolves the name (no Length is bad)

### DIFF
--- a/sdk/python/bittensor/cli/commands/lock.py
+++ b/sdk/python/bittensor/cli/commands/lock.py
@@ -162,9 +162,10 @@ def add_lock(
     except ValueError as error:
         app_ctx.output.error(f"invalid value for `--amount-alpha`: {error}")
         raise typer.Exit(2)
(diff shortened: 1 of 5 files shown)


unarbos (approved):
Reviewed locally: correct and mergeable.

PR #3034, opened 2026-08-03
Re-enable RootWeights proxy, scoped to set_root_weights

Rapiiidooo:
## What

Re-enables `ProxyType::RootWeights` with a deliberately minimal scope: a `RootWeights` proxy can dispatch `SubtensorModule::set_root_weights` and nothing else.

## Why

The variant (SCALE index 12) was neutralized when proxy filters moved to deny-by-default allowlists: `proxy.add_proxy(delegate, RootWeights, 0)` still **succeeds** and the delegation shows up in `Proxy.Proxies`, but every subsequent `proxy.proxy(...)` is filtered. The grant looks placed yet does nothing — the worst of both worlds. With root reborn (#2968) making per-coldkey root weights meaningful again, delegating exactly that call to a low-value operational key (keeping the coldkey offline) is a real use case.

## Changes

- **`runtime/src/proxy_filters/call_groups.rs`** — new `RootWeightCalls` group containing only `SubtensorModule::set_root_weights`. It lives with the proxy-specific groups and overlaps `SubtensorCommonCalls` (the same way `SmallTransferCalls` overlaps `BalanceTransferCalls`) rather than granting that whole group: the validator's consensus-weight surface (`set_weights`, batch/commit/reveal) stays out of reach — that is precisely what this proxy type must not delegate.
- **`runtime/src/proxy_filters/mod.rs`** — `RootWeights` maps to `RootWeightCalls` in both the executable filter (`proxy_type_filter`) and the runtime-API metadata (`proxy_filter_mode`); both derive from the same group, so they cannot drift.
- **`is_superset`** — `NonTransfer ⊃ RootWeights` already held. Added `NonFungible ⊃ RootWeights`: `NonFungible` already allows `set_root_weights` itself (via `SubtensorCommonCalls`), so it would be incoherent for a NonFungible proxy to be unable to manage a delegation that is a strict subset of its own rights. (`is_superset` gates a proxy adding/removing other delegations through `proxy.proxy`.)
- **`common/src/proxy.rs`** — `RootWeights` no longer reports `is_deprecated()`, so the runtime API's `deprecated` flag follows. Variant order is untouched: index 12 stays 12 and existing on-chain delegations keep their type.
- **SDK/docs** — `add_proxy` / `create_pure` help text no longer lists RootWeights among the deny-all types; docs reference pages regenerated from the registries.

`spec_version` bumped 441 → 442: mainnet already runs 441 (root-reborn deployed), and the `runtime-checks` spec-version gate requires runtime-affecting PRs to carry a spec newer than mainnet unless labeled `no-spec-version-bump` (which is for changes deliberately shipping without a runtime release — not this one).

## Existing on-chain grants

This change retroactively activates `RootWeights` delegations registered while the type was inert. Scan of `Proxy.Proxies` on finney at finalized block **8,765,950** (2026-08-03): 7,758 delegations across 6,722 delegator accounts — **exactly one** of type `RootWeights`:

| delegator | delegate | delay |
|---|---|---|
| `5HiFDVNX4ivCJFt9RvgRCtQKmAPgAXGX8BRgX3XKqfY9fFve` | `5FeNcqhfbLxkCeYtrKdDLkfNznmhCyvbAUMQNCfhddy8vHyX` | 0 |

Two options were weighed:

1. **Accept the existing grant** (chosen). The right that activates is exactly what the type's name promises — `add_proxy(RootWeights)` had no other plausible intent — and the delegator can revoke at any time with `remove_proxy`. A one-shot purge migration to strip a single row adds more code and upgrade risk than the retroactive grant it would remove.
2. Ship a storage migration purging pre-activation `RootWeights` entries. Rejected for the reason above; if reviewers prefer it, the affected entry is the single row listed here.

Additional mitigation: `set_root_weights` itself is still globally gated by `RootWeightSettingEnabled` (default off — it fails with `RootWeightSettingDisabled` until governance flips it via `AdminUtils::sudo_set_root_weight_setting_enabled`). The pre-existing delegation therefore cannot dispatch anything until that flag is enabled, leaving a clear window to revoke.

## Tests

- `root_weights_grants_exactly_set_root_weights` (new) — the filter accepts `set_root_weights`; rejects `set_weights`, `add_stake`, `remove_stake`, `swap_hotkey`, `set_children`; the metadata view exposes exactly one call.
- `narrow_proxies_have_exact_allow_lists` — now pins `allowed_calls(RootWeights) == {SubtensorModule::set_root_weights}`.
- `non_fungible_superset_is_explicit_allowlist` (new) — pins `NonFungible`'s superset set to `{NonFungible, RootWeights}`, mirroring the existing NonTransfer pin.
- `any_allows_everything_and_deprecated_allow_nothing` — RootWeights removed from the deny-all list (Triumvirate / Senate / Governance remain pinned there).
- The pre-existing `superset_relations_match_allowed_call_sets` and `all_call_groups_cover_runtime_call_metadata` invariants cover the new superset edge and the group inventory (no duplicate in `AllCalls`).

## Verification

- `SKIP_WASM_BUILD=1 cargo test -p node-subtensor-runtime -p subtensor-runtime-common` — green (25 proxy tests, full crate suites).
- `cargo clippy` (runtime + common, with tests) and `cargo fmt` — clean.
- SDK: `ruff`, proxy pytest subset (34 passed), and the docs staleness check (`generate.py --check`) — green.

### DIFF
--- a/common/src/proxy.rs
+++ b/common/src/proxy.rs
@@ -36,7 +36,7 @@ pub enum ProxyType {
     Registration,
     Transfer,
     SmallTransfer,
-    RootWeights, // Deprecated
+    RootWeights,
     ChildKeys,
     SudoUncheckedSetCode,
     SwapHotkey,
@@ -99,10 +99,7 @@ impl From<ProxyType> for u8 {
 
 impl ProxyType {
     pub fn is_deprecated(&self) -> bool {
-        matches!(
-            self,
-            Self::Triumvirate | Self::Senate | Self::Governance | Self::RootWeights
-        )
+        matches!(self, Self::Triumvirate | Self::Senate | Self::Governance)
     }
 }
 
--- a/docs/tx/add-proxy.mdx
+++ b/docs/tx/add-proxy.mdx
@@ -21,7 +21,7 @@ only to keys you control or fully trust.
 | Parameter | Type | Required | Description |
 | --- | --- | --- | --- |
 | `delegate_ss58` | string | yes | Key that will be allowed to sign for this account. |
-| `proxy_type` | string | no | Scope of calls the delegation covers. One of: Any, Owner, NonCritical, NonTransfer, Senate, NonFungible, Triumvirate, Governance, Staking, Registration, Transfer, SmallTransfer, RootWeights, ChildKeys, SudoUncheckedSetCode, SwapHotkey, SubnetLeaseBeneficiary, RootClaim. Triumvirate, Senate, Governance, and RootWeights are deprecated on the current runtime: they deny all calls, so a proxy of those types can dispatch nothing. Prefer the narrowest type that covers your use; Any can do everything the account can, including transfers. |
+| `proxy_type` | string | no | Scope of calls the delegation covers. One of: Any, Owner, NonCritical, NonTransfer, Senate, NonFungible, Triumvirate, Governance, Staking, Registration, Transfer, SmallTransfer, RootWeights, ChildKeys, SudoUncheckedSetCode, SwapHotkey, SubnetLeaseBeneficiary, RootClaim. Triumvirate, Senate, and Governance are deprecated on the current runtime: they deny all calls, so a proxy of those types can dispatch nothing. RootWeights covers exactly SubtensorModule.set_root_weights. Prefer the narrowest type that covers your use; Any can do everything the account can, including transfers. |
 | `delay` | integer | no | Announcement delay in blocks: the delegate must announce each call and wait this long before executing it, giving you time to veto. 0 executes immediately. |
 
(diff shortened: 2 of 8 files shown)


Rapiiidooo:
Closing this, the premise is obsolete.

RootWeights was designed back when set_root_weights was coldkey-signed (the hotkey was a call parameter), so delegating that coldkey signature made sense. Since root reborn the hotkey signs it directly (ensure_signed + root registration check), the coldkey is never involved anymore. Going through a coldkey proxy just fails with HotKeyNotRegisteredInSubNet.

So there's nothing left to delegate and nothing to protect either, the hotkey is already the hot key and can't move funds. Keeping the variant inert is fine. The one existing delegation on mainnet stays harmless.

PR #3046, opened 2026-08-05
feat(cli): treat saved multisig names as -w wallets

unarbos:
## Summary
- Passing `-w <multisig>` (a name from `btcli multisig add`) on any coldkey-signed command (`wallet transfer`, `tx *`, `call`, …) auto-detects the book entry, picks a local member coldkey to sign, and wraps the operation as a multisig approval.
- Matching pending ops are found by call hash so co-signers re-run the **same** command; the CLI fills the timepoint.
- Co-signer replay commands now use `-w <multisig>` when a preset is known. Docs updated for the short path.

## Example
```bash
btcli wallet transfer --dest izzi --amount-tao 0.1 -w guard-ms
# co-signer, identical:
btcli wallet transfer --dest izzi --amount-tao 0.1 -w guard-ms
```

## Test plan
- [ ] `btcli multisig add` a 2-of-3 with two local member wallets
- [ ] Fund one member with fee TAO; run `btcli wallet transfer --dest <addr> --amount-tao 0.1 -w <multisig>`
- [ ] Confirm opener prints pending / co-signer hint
- [ ] On the second member (or same machine with the other key selected), re-run the identical transfer command and confirm execution
- [ ] `btcli call Balances.transfer_keep_alive ... -w <multisig>` works without `--multisig`
- [ ] Explicit `btcli call ... --multisig X -w <member>` still works unchanged
- [ ] Non-multisig `-w` transfer unchanged

### DIFF
--- a/docs/guides/multisig.mdx
+++ b/docs/guides/multisig.mdx
@@ -138,12 +138,40 @@ btcli wallet regen-coldkeypub -w team-treasury --ss58 5Fmulti...sigAddress
 btcli wallet balance team-treasury
 ```
 
-## Step 4 — spend: open the operation
+## Step 4 — spend: treat the multisig like a wallet
 
-Alice proposes paying 10 TAO out of the treasury. The inner call is an intent
-spec — `{"op": <intent name>, ...args}`, same shape as a batch child — with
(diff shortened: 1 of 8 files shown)


unarbos:

PR #3047, opened 2026-08-05
btcli stake UX: free balance + 'all', paste any target hotkey, stake burn

unarbos:
## Summary

Fixes the `btcli stake add` flow that currently dead-ends when the validator hotkey isn't a local keyfile ("no hotkey in wallet '…' named '5FTV…' (available: none)"), and rounds out the staking UX:

- **`stake add` shows your free balance before asking how much to stake**, and the amount now accepts **`all`** — everything minus the existential deposit and a small fee headroom (resolved from chain state at build time; treated as an unbounded spend by spend-cap policies).
- **`stake add` offers the wallet's local hotkeys as a numbered picker** for the target validator, and **any pasted ss58 address, address-book name, or `WALLET/HOTKEY` is accepted** — staking to a validator never required its keyfile, now the CLI agrees.
- **Every prompted target hotkey/coldkey now accepts a pasted ss58 / address-book name** (e.g. `stake show`, `stake set-auto`, `stake child get`). Signer prompts stay strict: only keys that actually sign must exist locally.
- **`btcli stake burn`** mounts the existing `stake_burn` intent (buyback-and-burn via `add_stake_burn`) in the stake group. Its `limit_price` is now optional, defaulting to the current pool price plus a 5% `rate_tolerance` (same pattern as `add_stake`'s slippage protection), and it's MEV-shielded by default like the other pool-trading intents.

Docs under `docs/tx/` are regenerated from the registries (includes pre-existing Rust line-number drift in a few unrelated pages, since the generator links current `dispatches.rs` line ranges).

## Test plan

- [x] `pytest tests/unit` (1069 passed)
- [x] `codegen.check --coverage/--units/--names` green
- [x] `scripts/generate.py --check` green
- [x] Interactive pty run: `st add -w <wallet-with-hotkeys>` shows picker + free balance, accepts `all`, plan warns and summarizes "stake ALL free TAO…" (aborted at confirm)
- [x] Interactive pty run: `st add -w <wallet-without-hotkeys>` prompts for a pasted ss58 and proceeds
- [x] Interactive pty run: `stake show` accepts a pasted validator ss58 at the `--wallet-hotkey` prompt
- [x] `AddStake(amount_tao="all")` and `StakeBurn(limit_price=None)` build real calls against finney

### DIFF
--- a/docs/tx/add-collateral.mdx
+++ b/docs/tx/add-collateral.mdx
@@ -23,7 +23,7 @@ to clear unshielded at an unbounded AMM price.
 
 | Signer | Origin | Pallet | Wraps |
 | --- | --- | --- | --- |
-| `coldkey` | signed account (pallet role may apply) | SubtensorModule | [`SubtensorModule.add_collateral`](/code/pallets/subtensor/src/macros/dispatches.rs#L2524-L2534) |
+| `coldkey` | signed account (pallet role may apply) | SubtensorModule | [`SubtensorModule.add_collateral`](/code/pallets/subtensor/src/macros/dispatches.rs#L2529-L2539) |
 
 ## Parameters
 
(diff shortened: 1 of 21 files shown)


unarbos:

PR #3048, opened 2026-08-05
feat(cli): machine-readable dry-run plans with spend, args, and replay command

unarbos:
## Summary

`--dry-run --json` becomes a complete pre-approval record: an agent can dry-run a transaction, verify exactly what would happen, and hand a human (or a later automation step) the one command that submits it for real.

New fields in the plan JSON:

- **`spend_tao`** — exact TAO leaving the signer, from `intent.spend()`; `null` when the amount cannot be bounded before execution
- **`spend_unbounded`** — `true` for full-balance sweeps (`--amount all`) and ownership transfers, so a reviewer knows the spend has no upper bound
- **`args`** — the intent's exact parsed parameters (money as exact decimal strings), so what was parsed can be verified against what was intended
- **`command`** — the replay command: the current invocation plus any prompted answers, minus `--dry-run`/`--json`, shell-quoted. No `--yes` is injected, so a human still gets the confirmation prompt; automation appends `--yes --json` itself

Human-mode dry runs print the same replay command as a `run for real` line, and raw-call dry runs (`btcli call ... --dry-run`) carry the `command` field too.

Example:

```json
{
  "op": "transfer",
  "summary": "transfer ALL TAO to 5Fn7...",
  "signer": "coldkey",
  "signer_address": "5Fc3...",
  "args": {"dest_ss58": "5Fn7...", "amount_tao": "all", "keep_alive": true},
  "fee_tao": 0.000170824,
  "spend_tao": null,
  "spend_unbounded": true,
  "warnings": ["transfers the entire transferable balance"],
  "violations": [],
  "ok": true,
  "docs_url": "https://www.bittensor.com/docs/tx/transfer",
  "command": "btcli tx transfer --dest 5Fn7... --amount all -w affine --network finney"
}
```

## Test plan

- [x] `pytest tests/unit` — 1069 passed
- [x] `ruff check` / `ruff format --check` on changed files
- [x] Live Finney dry runs: bounded transfer (`spend_tao` exact), `--amount all` (`spend_unbounded: true` + warning), human mode (`run for real` line), raw call (`command` in fields)
- [x] Interactive dry run: prompted `--amount-tao` answer is folded into the replay command

### DIFF
--- a/sdk/python/bittensor/cli/call.py
+++ b/sdk/python/bittensor/cli/call.py
@@ -56,6 +56,7 @@ from ..intents.proxy import ProxyTypeChoice
 from . import multisig_helpers as ms_helpers
 from .context import ctx_of
 from .globals import with_tx_globals
+from .prompt import replay_command
 
 _MAX_SHOWN = 80  # truncate long param values (e.g. a wasm blob) in dry-run output
 
@@ -250,6 +251,7 @@ def call(
                 fields["multisig_preset"] = preset
         else:
             app_ctx.run(lambda client: _compose_only(client, prepare))
+        fields["command"] = replay_command()
         app_ctx.output.detail("dry run: raw call", fields)
         return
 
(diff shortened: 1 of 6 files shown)


unarbos:

PR #3049, opened 2026-08-05
feat(cli): keep key material out of shell history, ps output, and scrollback

unarbos:
## Summary

Secrets passed as CLI flags land in shell history and `ps` output; secrets printed to the terminal live on in scrollback. This PR closes both leaks across the wallet and EVM key commands:

- **New `cli/secrets.py` helpers** — `warn_argv_secrets` (detects secret-bearing flags on the real command line and warns "omit the flag to be prompted without echo"), `copy_to_clipboard` / `copy_secret_to_clipboard` (pbcopy/wl-copy/xclip/xsel with a graceful fallback to printing).
- **`wallet regen-coldkey` / `regen-hotkey`** warn when `--mnemonic`, `--seed`, or `--json-password` are passed as flags; option help now says to prefer the secure prompt.
- **`evm key import`** warns on `--private-key` / `--mnemonic` / `--keystore-password` flags, and gains a no-echo prompt fallback when no key source is given (a 64-hex-char answer is taken as a private key, anything else as a mnemonic). Non-interactive runs get a clear error instead of a hang.
- **`evm key export --private-key`** now copies the raw key to the clipboard on a terminal instead of printing it; `--show` prints explicitly, and piped output / `--json` still print, so `export ETH_PRIVATE_KEY=$(btcli evm key export --private-key)` keeps working.
- **`wallet decrypt --copy`** copies the decrypted plaintext to the clipboard instead of printing it (rejected in `--json` mode where it can't apply).

## Test plan

- [x] `ruff check` on changed files
- [x] `btcli wallet regen-hotkey --help` / `btcli evm key export --help` render the new help
- [x] Import sanity (`import bittensor` + changed modules)
- [x] `pytest tests/unit` — 1069 passed (run on the sibling multisig-preflight branch off the same base)

### DIFF
--- a/sdk/python/bittensor/cli/commands/evm/keys.py
+++ b/sdk/python/bittensor/cli/commands/evm/keys.py
@@ -2,6 +2,8 @@
 
 from __future__ import annotations
 
+import re
+import sys
 from pathlib import Path
 from typing import Optional
 
@@ -11,6 +13,8 @@ from ....evm import keys as evm_keys
 from ....evm.keys import write_keystore_file
 from ...context import ctx_of
 from ...globals import with_globals, with_unlock_globals
+from ...prompt import interactive
+from ...secrets import copy_secret_to_clipboard, warn_argv_secrets
 from ._shared import (
     EVM_KEY_HELP,
     _key_fields,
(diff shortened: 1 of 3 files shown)


unarbos:

PR #3050, opened 2026-08-05
feat(cli): resolve address-book/wallet names inside --call/--args JSON

unarbos:
## Summary

Top-level `--dest`-style flags already accept address-book, proxy-book, and wallet names, but the same names inside `--call` / `--args` JSON failed with the chain's bare "Base 58 requirement is violated". This PR makes name resolution reach into the JSON:

- **New `cli/call_names.py`** — `resolve_builder_params` (for `btcli call`) and `resolve_intent_args` (for intent commands) walk the parsed JSON against the generated call metadata and resolve names for every `AccountId`-typed parameter, recursively through nested calls (`sudo`, `proxy`, multisig inner calls), batch children, and `Vec<AccountId>` signatory lists.
- **`btcli call`** runs params through `resolve_builder_params` right after JSON load; **intent commands** (`btcli tx …`) get one `resolve_intent_args` pass over coerced args before submission, covering JSON-shaped fields the flag-level resolution can't see.
- Each substitution prints a dim `resolved address-book entry 'izzi' to 5Eqj…` notice; a name that matches nothing raises a did-you-mean error with close candidates instead of the Base58 ValueError.

## Test plan

- [x] `ruff check` on changed files, import sanity
- [x] Live finney dry run: `btcli call Balances.transfer_keep_alive --args '{"dest": "izzi", "value": 1000}' --dry-run` resolves the address-book name and shows the ss58 in the plan
- [x] `pytest tests/unit` — 1069 passed (run on the sibling multisig-preflight branch off the same base)

### DIFF
--- a/sdk/python/bittensor/cli/call.py
+++ b/sdk/python/bittensor/cli/call.py
@@ -54,6 +54,7 @@ import typer
 from .. import calls
 from ..intents.proxy import ProxyTypeChoice
 from . import multisig_helpers as ms_helpers
+from .call_names import resolve_builder_params
 from .context import ctx_of
 from .globals import with_tx_globals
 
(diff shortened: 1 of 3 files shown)


unarbos:

PR #3051, opened 2026-08-05
feat(sdk): fail multisig approvals early when the signer cannot cover deposit + fee

unarbos:
## Summary

Opening a multisig operation reserves `DepositBase + threshold × DepositFactor` from the opener (τ0.196 for 2-of-3 on finney) on top of the fee. An underfunded opener currently fails on-chain with a bare "cannot cover the transaction fee" that never mentions the deposit. This PR fails early, client-side, with the exact funding requirement:

- **`multisig.py`** — `MultisigFundingError` (carries a `fund <ss58> (<name>) with ≥ τ<amount> — the deposit is returned when the op executes` remediation), `check_multisig_funds` (deposit + estimated fee vs. free balance), and a preflight in `Multisig.approve` that reads pending state to decide whether this approval opens the op (deposit + fee) or is a later approval (fee only).
- **`intents/multisig.py`** — `MultisigExecute.build` / `MultisigApprove.build` run the same hard check before anything signs, and both intents gain `warnings()` so an underfunded opener surfaces at the pre-confirm plan stage too (deposit-only comparison, since no composed call exists yet).
- **Fail-open**: any error reading chain state (constants, balance, fee estimation) skips the check — the chain stays the authority, and the preflight can never block an otherwise submittable call.
- Known limitation (commented in code): with `--proxy-for` the check measures the signer's balance, not the proxied account's.

## Test plan

- [x] `pytest tests/unit` — 1069 passed, 1 skipped
- [x] `ruff check` on changed files, import sanity for `bittensor.multisig` / `bittensor.intents.multisig`

### DIFF
--- a/sdk/python/bittensor/intents/multisig.py
+++ b/sdk/python/bittensor/intents/multisig.py
@@ -22,6 +22,7 @@ from dataclasses import dataclass, field
 from typing import Any, Optional
 
 from .._generated import calls
+from ..multisig import check_multisig_funds, multisig_opening_shortfall
 from ..signing import public_view
 from ..sp_core import ss58_decode
 from .base import BuiltCall, Intent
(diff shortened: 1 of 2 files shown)


unarbos:

PR #3053, opened 2026-08-05
Recycle transaction fees instead of rewarding block authors

UnArbosFive:
## Motivation

Restore transaction-fee handling so block-producing validators do not receive transaction-fee revenue. Transaction fees are recycled by reducing TAO issuance. Existing AMM swap-fee behavior is unchanged.

## Changes

- Recycle native TAO transaction fees and tips instead of crediting the block author.
- For Alpha-paid transaction fees, swap the charged Alpha to TAO and atomically recycle the exact TAO output directly from the subnet account.
- Recycle both EVM base fees and priority fees through the same TAO fee handler.
- Remove the transaction-fee dependency on block authorship.
- Rename transaction-fee tests and comments to use "recycle" where issuance is reduced.

## Issuance invariants

Tests verify that:

- Balances total issuance and Subtensor total issuance remain equal.
- Both issuance ledgers decrease by the exact charged or converted TAO amount.
- Block-author balances remain unchanged.
- Alpha swap and TAO recycling changes roll back atomically on failure.
- Alpha-paid fees work without a block author and when the payer has zero TAO.

## Verification

- `cargo fmt --all -- --check`
- `cargo test --package subtensor-transaction-fee --lib`
- `SKIP_WASM_BUILD=1 cargo test --package node-subtensor-runtime --test evm_transaction_fee`
- `cargo clippy --package subtensor-transaction-fee --all-targets -- -D warnings`

### DIFF
--- a/docs/tx/add-collateral.mdx
+++ b/docs/tx/add-collateral.mdx
@@ -23,7 +23,7 @@ to clear unshielded at an unbounded AMM price.
 
 | Signer | Origin | Pallet | Wraps |
 | --- | --- | --- | --- |
-| `coldkey` | signed account (pallet role may apply) | SubtensorModule | [`SubtensorModule.add_collateral`](/code/pallets/subtensor/src/macros/dispatches.rs#L2524-L2534) |
+| `coldkey` | signed account (pallet role may apply) | SubtensorModule | [`SubtensorModule.add_collateral`](/code/pallets/subtensor/src/macros/dispatches.rs#L2529-L2539) |
 
 ## Parameters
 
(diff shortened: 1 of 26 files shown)


mcjkula:
Why is the TAO being recycled, but the TAO from the ALPHA burned? Why wouldn't we also recycle the TAO from the ALPHA-paid fee (it says `// Sell the Alpha fee and send the resulting TAO to the burn account.`)? Also in multiple places it calls the recycling of TAO "burn", can we please not add more code that conflates the two? It's already difficult enough to understand when "burn" means burn or actually recycling @UnArbosFive

UnArbosFour (approved):

UnArbosSix (approved):

PR #3065, opened 2026-08-10
v2 limit orders: introduction of linked orders

girazoki:
## Description
Introduces v2 limit order types which allow for **linked-orders**:

**A linked order is sized as a fraction of the output another order already produced**, rather than as an absolute amount the user must know at signing time. The driving case is a rotation — "sell my subnet-7 alpha, then put the TAO that produced into subnet 12" — and its mirror, "buy into subnet 7 now, and take profit on exactly the alpha that buy produced".

Two additions to the payload express the whole mechanism:

- **OrderV2::amount** is an OrderAmount: either **Fixed**(u64) (v1 semantics) or **LinkedPercentage** { provider, pct }.
- **OrderV2::has_linked_orde**r declares that this order's output should be recorded so later linked orders can draw against it. Without the flag nothing is recorded and nothing can link to the order — being a provider is opt-in, in the payload the user signs.

The two halves are independent, so an order can be a provider with a fixed amount, a consumer whose own output is not recorded, both at once (which is how chains longer than two legs are built), or neither — which is exactly a v1 order.

## Related Issue(s)

Builds upon https://github.com/RaoFoundation/subtensor/pull/3002

## Type of Change

- [ ] Documentation update

## Checklist


UnArbosSix:
replaced by #3086

PR #3067, opened 2026-08-10
fix(subtensor): clear voting state on subnet dissolution

UnArbosFive:
## Summary

- clear per-validator voting-power rows through the existing metered subnet map cleanup
- remove voting-power aggregate and configuration values during subnet parameter cleanup
- cover dissolution and netuid reuse with an end-to-end regression test

## Why

Voting-power state is scoped by netuid. Without lifecycle cleanup, a new subnet that reuses a dissolved netuid can inherit validator rows, aggregate voting power, tracking state, a pending disable block, and a custom EMA alpha from the previous subnet.

The validator map remains weight-metered so dissolution stays resumable for subnets with many entries. Netuid reuse is already blocked until cleanup completes.

Stacked on #3058.

## Tests

- SKIP_WASM_BUILD=1 cargo test --package pallet-subtensor --lib -- tests::voting_power::
- SKIP_WASM_BUILD=1 cargo test --package pallet-subtensor --lib -- tests::networks::dissolve_clears_all_per_subnet_storages --exact
- cargo fmt --all -- --check


UnArbosFive:
Superseded by the fixes now included in release-444.

PR #3128, opened 2026-08-28
Preserve hotkey successor mappings until netuid dissolution

UnArbosFour:
## Summary

Stops clearing `HotkeySuccessor` when a neuron is replaced or re-registered, ensuring smart contracts can continue locating funds after a hotkey swap followed closely by neuron deregistration—even within the same batch.

Netuid dissolution already clears `HotkeySuccessor` and `HotkeyRoot` by prefix, so this PR retains that existing cleanup and documents why lineage must survive individual neuron deregistration.

## Testing

- Adds regression coverage for same-block hotkey swap and neuron deregistration.
- Retains existing coverage confirming lineage is cleared when the netuid is dissolved.


unarbos:
Closing: rename only, no behaviour change; reopen if still needed.

PR #3129, opened 2026-08-28
Restore the RootWeights proxy

UnArbosFive:
## Problem

`set_root_weights` is currently available through `Any`, `NonTransfer`, `NonFungible`, and `NonCritical`, but each of those proxy types grants substantially broader authority than root basket curation alone. The stable `RootWeights` proxy type still exists at index 12, but it is marked deprecated and its filter denies every call, so users cannot delegate this operation narrowly.

## Root cause

When proxy filtering was reorganized into additive call groups, `set_root_weights` was placed in the broad residual Subtensor call group while `RootWeights` remained grouped with deprecated deny-all proxy types. That preserved the encoded proxy index but removed its original useful behavior.

## Fix

- Reactivate `RootWeights` without changing its stable on-chain index.
- Give it an explicit call group containing only `SubtensorModule::set_root_weights`.
- Expose the same one-call allowlist through proxy filter metadata and stop reporting the type as deprecated.
- Preserve the existing permissions of `Any`, `NonTransfer`, `NonFungible`, and `NonCritical` by including the new call group in their additive policies.
- Cover the active status and exact one-call allowlist in the proxy filter tests.

## Validation

- `cargo fmt --check --all`
- `git diff --check`

Rust builds and tests are left to CI under the repository's quick-preflight policy. This runtime-affecting change requires release handling through a spec version newer than mainnet or the corresponding no-bump label.

### DIFF
--- a/common/src/proxy.rs
+++ b/common/src/proxy.rs
@@ -36,7 +36,7 @@ pub enum ProxyType {
     Registration,
     Transfer,
     SmallTransfer,
-    RootWeights, // Deprecated
+    RootWeights,
     ChildKeys,
     SudoUncheckedSetCode,
     SwapHotkey,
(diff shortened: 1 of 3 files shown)


unarbos:
Closing: obsolete — #3155 removed `set_root_weights` and root weights entirely, so there is no call for this proxy to gate.

PR #3132, opened 2026-08-31
add `SetRootWeights` proxy

UnArbosSix:
## Motivation

Operators need to delegate root-basket curation without granting a broader proxy scope. Reusing the deprecated `RootWeights` variant could unexpectedly reactivate existing delegations, so this change introduces a new proxy type with a fresh wire index.

## Changes

- Add `SetRootWeights` as proxy type index `18` while leaving deprecated `RootWeights` at index `12` deny-all.
- Restrict `SetRootWeights` to `SubtensorModule::set_root_weights`.
- Preserve the existing broad proxy behavior by moving `set_root_weights` into a dedicated call group included by the same broad filters.
- Expose the new type through runtime filter metadata and the Python SDK.
- Update proxy guides, transaction reference pages, and generated intent catalog data.
- Add tests for the stable SCALE index and exact allowed-call set.

## Behavioral and upgrade impact

This is a runtime-facing proxy enum and filter change and therefore takes effect with a runtime upgrade. It requires no storage migration. Existing `RootWeights` proxies remain deprecated and deny all calls; users must explicitly create or update a proxy using `SetRootWeights`.

## Validation

- Added unit assertions for wire indices and the `SetRootWeights` allowlist.
- `git diff --check` passes.
- Rust formatting could not be executed in the review environment because rustup attempted to write to a read-only locked environment.
- Python Ruff and generated-reference checks could not be executed because the existing environment does not provide `uv`.

### DIFF
--- a/common/src/proxy.rs
+++ b/common/src/proxy.rs
@@ -42,6 +42,8 @@ pub enum ProxyType {
     SwapHotkey,
     SubnetLeaseBeneficiary,
     RootClaim,
+    /// Root basket curation (`set_root_weights`) only.
+    SetRootWeights,
 }
 
 impl TryFrom<u8> for ProxyType {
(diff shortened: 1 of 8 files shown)


unarbos:
Closing: obsolete — #3155 removed `set_root_weights` and root weights entirely, so there is no call for this proxy to gate.

PR #3171, opened 2026-09-16
frontier: stop KV mapping sync at a missing header instead of retrying forever

loayei97:
Client-only fix in the vendored Frontier KV mapping-sync worker; no runtime change.

**Bug:** on a warp-synced node run with `--history-backfill=skip`, headers below the warp target don't exist. `fetch_header` returns `Err("Header not found")` for the first such block, `sync_one_block` bails before rewriting `current_syncing_tips`, and every tick retries that same block. New leaves are never picked up, the frontier mapping DB freezes minutes after boot, and `eth_getBlockByHash` / `eth_getTransactionByHash` / `eth_getTransactionReceipt` return `null` for every block, including the tip.

**Fix:** treat a missing header like a block below `sync_from` and stop the backward walk there. Real backend errors still surface as `Err`.

**Repro (testnet, chain 945):** node at `8d5f20e` warp-synced with `--history-backfill=skip --state-pruning=72000 --blocks-pruning=72000`; `frontier/paritydb` last written 5 minutes after boot; `eth_getBlockByHash(<latest hash>)` → `null` while `eth_getBlockByNumber("latest")` works and the head matches `test.chain.opentensor.ai`.


unarbos:
Thanks — the fix looks right. Fork PRs cannot run the self-hosted Rust CI here, so I mirrored your commit (authorship preserved) to #3172 to get the full checks; this PR will be closed once #3172 lands.

unarbos:
Closing: merged to main via #3172 as a6c6e01ab (your commit, authorship preserved). Thanks!

